PR #61, opened 2022-11-08
46 juergen documentation cable pop

juergenknauer:
# CABLE

## Description

Fixes #(issue)

## Type of change

- [ ] Bug fix

## Checklist

### DIFF
--- a/src_pop/core/biogeochem/POP.F90
+++ b/src_pop/core/biogeochem/POP.F90
@@ -12,7 +12,9 @@
 ! This SUBROUTINE calls most other subroutines and functions in POP.
 ! POP receives StemNPP from CASA and returns woody turnover rate and vegetation height to the
 ! main model. External inputs is disturbance regime (total or total and partial) and disturbance interval.
+! 
 ! Main variables and their units are:
+!
 ! Inputs:
 ! Stem NPP (kg C m-2 yr-1)
 ! disturbance interval (yr)
@@ -2077,9 +2079,6 @@ CONTAINS
 
   !------------------------------------------------------------------------------------
 
-  !! Patch Disturbance
-  !! This subroutine kills all biomass in a patch when prescribed disturbance
-  !! interval is reached.
   SUBROUTINE Patch_disturb(pop,idisturb,precip)
     !*
     ! Simulates a disturbance event by killing all biomass in a patch.
--- a/src_pop/core/biogeochem/POP.F90
+++ b/src_pop/core/biogeochem/POP.F90
@@ -8,10 +8,18 @@
 !--------------------------------------------------------
 !*
 ! When referring to this code in publications, please cite:
+!
 ! Haverd, V., Smith, B., Cook, G., Briggs, P.R., Nieradzik, L., Roxburgh, S.R., Liedloff, A.,
 ! Meyer, C.P. and Canadell, J.G., 2013.
 ! A stand-alone tree demography and landscape structure module for Earth system models.
 ! Geophysical Research Letters, 40: 1-6.
+!
+! A more detailed description of the code and the equations
+! can be found in (Appendix):
+! Haverd, V., Smith, B., Nieradzik, L.P., Briggs, P.R., 2014.
+! A stand-alone tree demography and landscape structure module for Earth system models:
+! integration with inventory data from temperate and boreal forests.
+! Biogeosciences, 11, 4039-4055. doi:10.5194/bg-11-4039-2014
 
 
 !DISCLAIMER, COPYRIGHT AND LICENCE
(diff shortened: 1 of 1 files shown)


juergenknauer:
High-level documentation of the POP module. Some subroutines could not be documented due to insufficient understanding of the code and lack of knowledge wrt the purpose of the subroutines. Code contains many (apparently redundant, but to be checked) similar subroutines that do the same calculations on a different spatial organisational level (single patch vs. all patches).  Presence of optional input parameters with unknown function.

src_pop/core/biogeochem/POP.F90
```diff
@@ -6,11 +6,20 @@
 
 !CITATION
 !--------------------------------------------------------
-!When referring to this code in publications, please cite:
```
har917, line 9: I would start with a lead-in comment such as 

!> This MODULE contains the code for the POP model.  The principal components of the MODULE are the 
! initialisation SUBROUTINE [[InitPOP2D_Poission]], the time stepping SUBROUTINE [[POPStep]] and routines for 
! allometry, recruitment and disturbance.

then the existing inline documentation is ok

src_pop/core/biogeochem/POP.F90
```diff
@@ -2007,14 +2048,23 @@ END SUBROUTINE Patch_partial_disturb2
 
   !*******************************************************************************
 
-
+  !! # Patch Disturbance
```
har917, line 2051: Do you want this comment in the in line documentation?

har917 (changes requested):
Jurgen,

A couple of suggestions in line (one notable, the other typographic) - plus
1.  There are some existing comments that start with !**************  These are being picked up by FORD as inline comments (at the MODULE level) and I think they don't need to be
2.  Have a think about whether (or not) it would be better to add in units (for the main inputs/outputs) for each SUBROUTINE.

As an aside we will have to break this MODULE up at some point.

src_pop/core/biogeochem/POP.F90
juergenknauer: I have replaced all "!****" comments with "!----"

src_pop/core/biogeochem/POP.F90
juergenknauer: Thank you for the Review, Ian. I have followed your suggestions. See committed code.

src_pop/core/biogeochem/POP.F90
```diff
@@ -2005,16 +2075,25 @@ SUBROUTINE Patch_partial_disturb2(pop,idisturb)
   END SUBROUTINE Patch_partial_disturb2
 
 
-  !*******************************************************************************
-
+  !------------------------------------------------------------------------------------
 
+  !! Patch Disturbance
```
har917, line 2080: did you really want these !! lines here? They appears as a rogue comment at the top pf the POP MODULE page - not in the Patch_disturb box

src_pop/core/biogeochem/POP.F90
```diff
@@ -4,13 +4,51 @@
...
+!*
+! This MODULE contains code for the woody demography model POP. The principal components
+! of this MODULE are the initialisation SUBROUTINE [[InitPOP2D_Poisson]], time stepping
+! SUBROUTINE [[POPStep]] and routines for allometry, recruitment, mortality and disturbance.
+! The main SUBROUTINE is [[POPStep]] which is called from SUBROUTINE POPdriver in casa_cable.F90.
+! This SUBROUTINE calls most other subroutines and functions in POP.
+! POP receives StemNPP from CASA and returns woody turnover rate and vegetation height to the
+! main model. External inputs is disturbance regime (total or total and partial) and disturbance interval.
+! Main variables and their units are:
+! Inputs:
```
har917, line 16: I'd include a blank line here to get this onto a new line

har917 (changes requested):
Hi Jurgen,

I'm generally happy with this - one comment about a rogue comment (before patch_disturb) and one suggestion.  

More generally though the math typesetting needs looking at - it needs to be \( ... \) not $...$.

Perhaps also apply to the top level description as well (rather than as e.g. kg C m-2 yr-1)

src_pop/core/biogeochem/POP.F90
juergenknauer: I have added a few more blank lines. Looks like there is a tradeoff between readability of the code and readability of the rendered documentation.

src_pop/core/biogeochem/POP.F90
juergenknauer: These comments were indeed redundant and at the wrong place. I have removed them.

har917:
@ccarouge I was going to approve (even merge this doc PR in) but it looks like the process has changed and we're not deploying draft docs anymore.  I wanted to check those before doing so - any advice?

ccarouge:
@har917 This is a very old PR. I know Juergen started this way and then went to do it differently (directly in the CABLE-POP_TRENDY branch I think). So I have on my todo list for a long time to review if this is still needed and same with the `src_pop` directory and remove everything that's outdated. But I haven't yet gone through it.

juergenknauer:
I think this one refers to the documentation of CABLE-POP only and not to code changes. I remember correctly, all that happened on a separate repository that's there specifically for documentation. 

It would be good if that one could be merged as it is almost complete and was just pending minor editorial suggestions.

ccarouge:
See #597 .

PR #67, opened 2022-11-10
AddFlowchart

ccarouge:
# CABLE

## Description

I am trying out the flowchart capabilities of the documentation.

It looks ok but the "code" to produce it would become quite tedious for large flowcharts. The interesting bit is the flowchart boxes can be clickable to send people to different parts of the documentation.

## Type of change

## Checklist

- [X] The new content is accessible and located in the appropriate section.
- [X] I have checked that links are valid and point to the intended content.

Note: the links won't work in the preview as they refer to the address of the published website but I have tested them locally.

### DIFF
--- a/documentation/docs/user_guide/tutorial/compilation.md
+++ b/documentation/docs/user_guide/tutorial/compilation.md
@@ -1,26 +1,25 @@
-# Install and compile
-To install cable you need svn.
+# Installation instructions
 
-## To install CABLE
+To install CABLE you need to have the following already installed on your system:
 
-CABLE can be downloaded from the [subversion repository](https://trac.nci.org.au/svn/cable) hosted at [NCI](https://nci.org.au) once a user has requested admission to the CABLE software group at NCI, as described [here](https://trac.nci.org.au/trac/cable/wiki/CABLE_Registration). We recommend that within the repository you make a copy of the [trunk](https://trac.nci.org.au/svn/cable/trunk) to your own user branch, and then checkout CABLE from there. 
+ - SVN
+ - a Fortran compiler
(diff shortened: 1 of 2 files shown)


ccarouge:
@JhanSrbinovsky @rkutteh After reviewing the installation doc page, I decided to try out the [Diagram](https://mermaid-js.github.io/mermaid/#/flowchart) ability of the documentation. I don't know if it's useful yet so I'd appreciate any thoughts.

The advantages:
 
- easier to keep up to date than a separate image file
- clickable nodes to refer users to specific places in the documentation

The disadvantages:

 - setting up the diagram in the Markdown file is not that simple, need to learn yet a different syntax

It is possible to define classes for the nodes and links centrally so people don't have to do that for each diagram but for the moment, I've put that within the diagram.

ccarouge:
I am wondering if this is a useful addition to have or not.

rkutteh (approved):

rkutteh:
I think the flowchart looks great. To me the more diagrams like this we have the more user friendly the documentation will be. Given that at the present clicking on the boxes in the diagram leads nowhere, I hope when it is deployed there is a way to open a new tab when a user clicks on a box to get more info, so as to keep the original page with the diagram available to look at and not have to navigate there again.

ccarouge:
The links from the graphs should work. I'm updating them to point to the correct path.

For opening in a new tab, one would have to use their browser/computer way to do that. For me, I know I can do Cmd+click to open in a new tab. I'm guessing on a Windows machine, it might work with Ctrl+click but that's likely dependent on each browser/OS combination.

PR #68, opened 2022-11-10
Alternate documentation organisation for cbl_hgtAbove_snow.F90

ccarouge:
# CABLE

## Description

I like having a summary in English of what to expect a subroutine will do at the start of the subroutine. It helps a lot in understanding each line but also the order of operations. As well as understanding if the order of operations is appropriate or could be modified. And putting it at the top usually helps the writer to abstract the explanation rather than following too much the code itself.

Also, the explanation should not necessarily be a repeat of the equations in the code but an explanation of these. For example, in your text, I had to wait until the end to understand that `1.2 * SnowDepth/SnowDensity` is actually simply calculating the snow height in m. This should come first because it isn't clear from the code! Nowhere in the code, we can see: we want to do Canopy Height - Snow Height. But we don't have the snow height directly so it's calculated from ... 

You were using different notations in the equations compared to the variable names. Are these notations some kind of standard? I have replaced by the variable names. Do you prefer the first way it was done? With an alternate notation, people might not know the standard and would need an explanation of what each variable is.

## Type of change

## Checklist

- [X] The new content is accessible and located in the appropriate section.


ccarouge:
Handled by another pull request. Not needed.

PR #74, opened 2022-11-14
Documentation for cbl_hgtAbove_snow and cbl_lai_eff

### LINKED ISSUE #71: documentation for cbl_hgtAbove_snow.F90

ccarouge:
# CABLE

## Description

This pull request finishes the documentation for the roughness calculations in CABLE. 

Fixes #71 #72 

## Type of change

## Checklist

### DIFF
--- a/src/science/roughness/cbl_hgtAbove_snow.F90
+++ b/src/science/roughness/cbl_hgtAbove_snow.F90
@@ -12,15 +12,21 @@ subroutine HgtAboveSnow( HeightAboveSnow, mp, z0surf_min, HGT_pft, &
   !* Subroutine computes height of canopy above ground/snow surface when there
   !  is snow present
   !
-  !  inputs:  <ul> <li> mp - number of land points</li>
-  !           <li> z0surf_min - minimum roughness length of surface (in m) </li>
-  !           <li>HGT_pft(mp) - height of canopy with no snow (in m)</li>
-  !           <li>SnowDepth(mp) - depth of snow (in mm/m2 liquid water)</li>
-  !           <li>SnowDensity(mp) - density of snow (in kg/m3)</li> </ul>
+  !  inputs:
   !
-  !  outputs: <ul> <li> HeightAboveSnow (m) - effective height of canopy (in m) </li></ul>
-  !           output variable formerly known as rough%hruff
+  !  * mp - number of land points
+  !  * z0surf_min - minimum roughness length of surface (in m)
+  !  * HGT_pft(mp) - height of canopy with no snow (in m)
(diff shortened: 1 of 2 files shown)


ccarouge:
@har917 This is a new pull request with the documentation for the canopy height and LAI due to snow. I have added some minor modifications from me. Could you review?

I have one remaining question: in the equations you introduce new notations (hc,snow, dsnow, etc.). Why not use the variable names from the subroutines? Otherwise, I would say we need to add the mapping between the notations in the equations and the variable names in the description of the variables.

When an equation comes from a paper, I understand the need to keep the notation but we don't have any reference here so I'm not sure a notation change is needed.

src/science/roughness/cbl_hgtAbove_snow.F90
```diff
@@ -1,47 +1,71 @@
...
+    !! Output. Effective height of canopy, known as `rough%hruff` elsewhere. (m)
   
   !re-decl input args  
-  real :: z0surf_min
-  real :: HGT_pft(mp) 
-  real :: SnowDepth(mp) !snow depth (liquid water ) 
-  real :: SnowDensity(mp) 
+  real :: z0surf_min !! minimum roughness length of the surface (m)
+  real :: HGT_pft(mp) !! height of the canopy without snow (m)
+  real :: SnowDepth(mp) !! amount of snow (mm m(^{-2}\) liquid water)
```
har917, line 30: math typo - should be
!! amount of snow (mm m\(^{-2}\) liquid water)

har917 (changes requested):
One typo to address - otherwise all good.

I've left the issue of matching equations to variables to one side.  I see your point - but with the variable names as they are at this time the result would be look awful.  Another issue here is that some of the variable names are misleading e.g. SnowDepth is not a depth of snow but an amount of water.  I'd prefer not to go about fixing the variable naming convention at the same time as writing the docs.

src/science/roughness/cbl_hgtAbove_snow.F90
har917: aargh - that should be
!! amount of snow (mm m\(^{-2}\) liquid water)

har917: okay the comment is not picking up the Markup.  It needs a math start after 'mm m' and a math end after the '2}'

ccarouge: Just some help with the Markdown, you need to quote the whole thing with \` or put in a code block to say it is pre-formatted:
`(mm m\(^{-2}\)`

Otherwise `\` is the escape character that allows printing special characters. E.g:

    \# It's not a title if I start with `\`

renders as
\# It's not a title if I start with `\`

PR #79, opened 2022-11-21
header information for cbl_zetar

### LINKED ISSUE #78: Documentation of cbl_zetar MODULE
First attempt at documentation for this module - using the suggested structure under issue 71

har917:
# CABLE

## Description

Fixes #78 

## Type of change

- [ ] Bug fix

## Checklist

### DIFF
--- a/src/science/canopy/cbl_zetar.F90
+++ b/src/science/canopy/cbl_zetar.F90
@@ -68,9 +68,9 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
 
 IMPLICIT NONE
 
-INTEGER, INTENT(IN) :: mp     !! number of land points 
-INTEGER, INTENT(IN) :: NITER  !! number of MO-iterations
-INTEGER, INTENT(IN) :: nrb    !! number of radiation bands
+INTEGER, INTENT(IN) :: mp     !! number of land points (-)
+INTEGER, INTENT(IN) :: NITER  !! number of MO-iterations (-)
+INTEGER, INTENT(IN) :: nrb    !! number of radiation bands (-)
 
 REAL, INTENT(OUT) :: canopy_zetar(mp, NITER)  !!OUT: stability parameter \(\xi\)
 REAL, INTENT(OUT) :: canopy_zetash(mp, NITER) !!OUT: as canopy_zetar for soil
--- a/src/science/canopy/cbl_zetar.F90
+++ b/src/science/canopy/cbl_zetar.F90
@@ -27,33 +27,36 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
   !  [Kowalczyk et al. (2006)](http://www.cmar.csiro.au/e-print/open/kowalczykea_2006a.pdf)
   !  - section 3.1, equations 1-9.  The two outputs of the SUBROUTINE
   !  are the local (in space, time and by iteration counter) value of \(\xi\)
-  !  (Equation 9) `canopy_zetar` and a related quantity `canopy_zetarsh`.
+  !  (Equation 9) `canopy_zetar` and a related quantity `canopy_zetash`.
   !
   !  `canopy_zetar` is evaluated from the total land (soil+canopy)
   !  surface fluxes of momentum, sensible heat and latent heat.
-  !  `canopy_zetarsh` is the equivalent variable evaluated from the soil
-  !  contribution to those fluxes only.  `canopy_zetarsh` is used in conjunction
+  !  `canopy_zetash` is the equivalent variable evaluated from the soil
+  !  contribution to those fluxes only.  `canopy_zetash` is used in conjunction
   !  with the [[sli_main_mod]] soil model to moderate the fluxes from the soil
   !  underneath a canopy.
   !
-  !  `canopy_zetar` and `canopy_zetarsh` quantify the role that the surface
+  !  `canopy_zetar` and `canopy_zetash` quantify the role that the surface
   !  fluxes play in setting the efficiency of turbulent transfer from the land
   !  to the atmosphere, and hence the aerodynamic component of the resistance
   !  network for those same surface fluxes (an implicit problem which requires
-  !  iteration to solve).
+  !  iteration to solve). 
   !
-  !  `canopy_zetar` and `canopy_zetarsh` are initialised to zero in
-  !  [[define_canopy]] and updated NITER(>1) times
+  !  `canopy_zetar` and `canopy_zetash` are initialised to `CZETA0`=0 in
+  !  [[define_canopy]] and updated `NITER`(>1) times
   !  during the calculation of the energy balance.  The value of the variables
-  !  at each iteration are stored in memory to aid in the assessment
+  !  at each iteration are stored in memory to aid in the diagnosis
   !  of convergence.
   !
-  !  The outputs `canopy_zetar` and `canopy_zetarsh` are known as
-  !  `canopy%zetar` and `canopy%zetarsh` in [[define_canopy]] and elsewhere
-  !  in the code.
+  !  The outputs `canopy_zetar` and `canopy_zetash` are known as
+  !  `canopy%zetar` and `canopy%zetash` in [[define_canopy]] and elsewhere
+  !  in the code. `canopy%zetar` is used in [[define_canopy]] and
+  !  [[comp_friction_vel]]; `canopy%zetash` is used in `[[define_canopy]].
   !
-  !  Special cases apply if NITER=2, or if `canopy_zetar` or `canopy_zetarsh`
-  !  exceed the prescribed upper `CZETPOS` or lower `CZETNEG` limits
+  !  Special cases apply if `NITER`=2, or if `canopy_zetar` or `canopy_zetash`
+  !  exceed the prescribed upper `CZETPOS` or lower `CZETNEG` limits. `NITER`(=4)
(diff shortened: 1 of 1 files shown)


src/science/canopy/cbl_zetar.F90
```diff
@@ -13,34 +15,85 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
...
+  !
+  !  The outputs `canopy_zetar` and `canopy_zetash` are known as
+  !  `canopy%zetar` and `canopy%zetash` in [[define_canopy]] and elsewhere
+  !  in the code. `canopy%zetar` is used in [[define_canopy]] and
+  !  [[comp_friction_vel]]; `canopy%zetash` is used in `[[define_canopy]].
+  !
+  !  Special cases apply if `NITER`=2, or if `canopy_zetar` or `canopy_zetash`
+  !  exceed the prescribed upper `CZETPOS` or lower `CZETNEG` limits. `NITER`(=4)
+  !  is defined in [[cable_types_mod]]; `CZETMUL`, `CZET0`, `CZETPOS` and
+  !  `CZETNEG` in [[cable_phys_constants_mod]].
```
ccarouge, line 61:
```suggestion
  !  A special case applies if `NITER`=2.`canopy_zetar` and `canopy_zetash`
  !  are also bounded by the interval `[CZETNEG, CZETPOS]`.
  !
  !`NITER`(=4) is defined in [[cable_types_mod]]; `CZETMUL`, `CZET0`, `CZETPOS` and
  !  `CZETNEG` in [[cable_phys_constants_mod]].
```
I am not sure I am too keen on the end here that tells where the constants come from.

src/science/canopy/cbl_zetar.F90
```diff
@@ -13,34 +15,85 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
...
+  !
+  !  <br></br>
+  !
+  !  *NOTE* The INTENT statements for `canopy_zetar` and `canopy_zetash` need
+  !  sorting - this code would fail strict compilation. Perhaps also need to consider
```
ccarouge, line 66: I'm not sure I follow what you mean with "fail strict compilation"

src/science/canopy/cbl_zetar.F90
```diff
@@ -13,34 +15,85 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
...
+  !
+
 IMPLICIT NONE
 
-INTEGER, INTENT(IN) :: mp
-INTEGER, INTENT(IN) :: NITER
-INTEGER, INTENT(IN) :: nrb
+INTEGER, INTENT(IN) :: mp     !! number of land points 
+INTEGER, INTENT(IN) :: NITER  !! number of MO-iterations
+INTEGER, INTENT(IN) :: nrb    !! number of radiation bands
```
ccarouge, line 74:
```suggestion
INTEGER, INTENT(IN) :: mp     !! number of land points (-)
INTEGER, INTENT(IN) :: NITER  !! number of MO-iterations (-)
INTEGER, INTENT(IN) :: nrb    !! number of radiation bands (-)
```

src/science/canopy/cbl_zetar.F90
```diff
@@ -13,34 +15,85 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
...
 
-REAL, INTENT(OUT) :: canopy_zetar(mp, NITER)
-REAL, INTENT(OUT) :: canopy_zetash(mp, NITER)
-INTEGER, INTENT(IN) :: iter
+REAL, INTENT(OUT) :: canopy_zetar(mp, NITER)  !!OUT: stability parameter \(\xi\)
+REAL, INTENT(OUT) :: canopy_zetash(mp, NITER) !!OUT: as canopy_zetar for soil
+INTEGER, INTENT(IN) :: iter                   !! iteration counter
```
ccarouge, line 78:
```suggestion
REAL, INTENT(OUT) :: canopy_zetar(mp, NITER)  !! stability parameter \(\xi\)
REAL, INTENT(OUT) :: canopy_zetash(mp, NITER) !! as canopy_zetar for soil
INTEGER, INTENT(IN) :: iter                   !! iteration counter
```
If the "INTENT" is specified, we don't need to tell that's the output. Units? I'm guessing that's unitless.

ccarouge (changes requested):
I think I'll have more changes afterwards but let's get this bunch sorted first.

In particular, the subroutine description should start with the purpose, so the current first paragraph needs to be moved to later, I think.

src/science/canopy/cbl_zetar.F90
```diff
@@ -13,34 +15,85 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
                    cable_user_soil_struc, air_rho, met_tk,  met_fsd, &
                    rough_zref_tq, rough_hruff, rough_term6a, rough_z0soilsn,   &
                    canopy_vlaiw, canopy_zetash,  canopy_us, &
-                   canopy_fh, canopy_fe, canopy_fhs, canopy_fes ) 
+                   canopy_fh, canopy_fe, canopy_fhs, canopy_fes )
+  !* This SUBROUTINE forms part of the codebase to evaluate the surface
+  !  energy balance on a sub-diurnal basis (i.e. every CABLE time step).
+  !  It resides in the canopy science directory and is called from
+  !  [[define_canopy]] only.
```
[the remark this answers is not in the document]
har917: Part of the problem with the current code base is that navigation around it is nigh on impossible - hence why I was including comments about where it was located and where routines were called from.

src/science/canopy/cbl_zetar.F90
```diff
@@ -13,34 +15,85 @@ SUBROUTINE update_zetar( mp, NITER, canopy_zetar, iter, nrb, CVONK, CGRAV, CCAPP
...
+  !  `canopy_zetash` is the equivalent variable evaluated from the soil
+  !  contribution to those fluxes only.  `canopy_zetash` is used in conjunction
+  !  with the [[sli_main_mod]] soil model to moderate the fluxes from the soil
+  !  underneath a canopy.
+  !
+  !  `canopy_zetar` and `canopy_zetash` quantify the role that the surface
+  !  fluxes play in setting the efficiency of turbulent transfer from the land
+  !  to the atmosphere, and hence the aerodynamic component of the resistance
+  !  network for those same surface fluxes (an implicit problem which requires
+  !  iteration to solve).
```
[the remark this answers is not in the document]
har917: We could replace `canopy_zetar` and `canopy_zetash` with \(\xi\) and it would make sense.

src/science/canopy/cbl_zetar.F90
har917: Again this was included to aid navigation - at the moment a reader has no idea where these are set (namelists, internal variables, or hard-wired in the code)

src/science/canopy/cbl_zetar.F90
har917: Specifically canopy_zetar and canopy_zetash are INTENT(OUT) when they should be INTENT(INOUT) - as their values are used before they are set if SLI is used.  I'm not sure how this has snuck through compilation testing so far.

src/science/canopy/cbl_zetar.F90
har917: yes \(\xi\) is unitless (strictly it is in m/m)

on the question of whether to include the OUT in the comment - this refers back to the comment about line 65.  I was trying to identify the primary variables that are calculated in the documentation.  I'm happy either way though.

src/science/canopy/cbl_zetar.F90
ccarouge: It won't break compilation as it is currently. What happens is only the values for "iterplus" are correct. All values for preview iterations are not conserved in memory. So the dimension NITER on `canopy_zetar` and `canopy_zetash` is currently useless and `zetash` gets a crazy value for non-vegetated tiles. The initialisations in `cable_canopy` are discarded when entering `update_zetar` for the moment.

src/science/canopy/cbl_zetar.F90
ccarouge: For the documentation or the code, the `INTENT` is what should tell people if the variable is a returned value from the subroutine or not. The `INTENT` is picked up by FORD and appears in the variable table. I really don't think we need more. Except when INTENT isn't specified yet or when there are only the derived-types and not the specific variables listed as arguments.

To make the documentation clearer, we need a standard on the order of the subroutine arguments as FORD lists them in the order they are passed to the subroutine. So having the OUT variables grouped at the end of the list will make this easier to read (UMDP3 standard).

ccarouge (approved):
I have made a few changes. Mainly re-ordering a few elements. But I'm putting this as-is without review from you as I think it's good enough now.

src/science/canopy/cbl_zetar.F90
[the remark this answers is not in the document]
ccarouge: Done

src/science/canopy/cbl_zetar.F90
ccarouge: Left in for now

src/science/canopy/cbl_zetar.F90
ccarouge: change explanation and upgrade to WARNING as that's a bug

PR #80, opened 2022-11-30
Change User Guide structure according to Issue #75

### LINKED ISSUE #75: Suggest different structure under CABLE/documentation/docs/user_guide
To me the structure under **CABLE/documentation/docs/user_guide** is not quite right. I think a more reasonable structure would be (ignoring the alphabetical ordering):

- configurations
- inputs
- outputs
- science
- installation
- uber_quick_guide
- tutorials

**configurations** would contain config.md **and/or** user_guide/tutorial/configuration_desc.md
**outputs** would contain user_guide/tutorial/outputs_desc.md
**installation** would contain user_guide/tutorial/compilation.md and user_guide/tutorial/run_cable.md
**uber_quick_guide** would contain user_guide/tutorial/uber_quick_guide.md
**tutorials** would contain actual tutorials (e.g., sample runs with inputs and outputs)

Of course the above folders can also contain additional files in the future but I just moved around what is already there for now.

Another point worth mentioning is I think we should be explicit about which version of CABLE we are describing, as a lot of this info is version specific (maybe include the version number early in the path name). 

Sorry I did not include all of this stuff in a commit/pull_request but I thought a simple Issue is ok at this stage to get feedback.

Ramzi

ccarouge:
# CABLE

## Description

Reordering of the User Guide to some more logical order

Fixes #75

## Type of change

## Checklist


rkutteh (approved):
- Given that install.md is intended to go into a new folder called "installation", it would make more sense to name it installation.md
- Given that outputs_desc.md is intended to go into a new folder called "outputs", it would make more sense to name it "outputs.md"
- For the same reason I think index.md is better named user_guide.md, as the former is generic
- Finally, as soon as you get a chance to create an "outputs" folder, I can begin moving text from TRAC into it.

ccarouge:
Thanks. 

> For the same reason I think index.md is better named user_guide.md, as the former is generic

It has to be called `index.md` to appear as the landing page of the User Guide section instead of landing on a specific page when opening the section. That's why the structure ends up having quite a few `index.md` files here and there. And that's why the `output/` folder now only has a `index.md` file as I don't know what the structure of that section will be.

For example, in the Inputs section, we may eventually move the "List of inputs" to an index.md file for that section so that it opens when we open the section. But I don't know yet and it works ok as is for now.

PR #105, opened 2022-12-13
first comments on cbl_latent_heat

### LINKED ISSUE #104: Documentation for cbl_latent_heat

har917:
# CABLE

## Description

Documentation for the cbl_latent_heat module

Fixes #104 

## Type of change

- [ ] Bug fix

## Checklist

### DIFF
--- a/src/science/canopy/cbl_latent_heat.F90
+++ b/src/science/canopy/cbl_latent_heat.F90
@@ -234,8 +234,8 @@ ENDDO
 ! <br></br>
 
 !| - The latent heat flux associated with evaporation from puddles is set
-! to the area fraction of `ssnow_potev` provided that sufficient water is
-! available in the puddles.
+! to the area fraction of the potential evaporation (`pwet` * `ssnow_potev`),
+! provided that sufficient water is available in the puddles.
 !  
 canopy_fesp = MIN(ssnow_pudsto/dels*air_rlam,MAX(pwet*ssnow_potev,0.))
 
--- a/src/science/canopy/cbl_latent_heat.F90
+++ b/src/science/canopy/cbl_latent_heat.F90
@@ -141,9 +141,10 @@ INTEGER :: j
 ! is the latent heat flux associated with potential evaporation.
 !
 ! <br></br>
+! THe workflow is
 ! - First, a first estimate for the soil latent heat flux is calculated
 ! assuming that the flux is of liquid water. If `ssnow_potev` is negative
-! any limitation due to soil moisture availability is negected. f
+! any limitation due to soil moisture availability is negected.
 !<br></br>
 WHERE (ssnow_potev < 0. ) ssnow_wetfac(:) = 1.0
 canopy_fess= ssnow_wetfac * ssnow_potev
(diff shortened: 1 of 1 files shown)


src/science/canopy/cbl_latent_heat.F90
```diff
@@ -82,34 +192,50 @@ SUBROUTINE Latent_heat_flux( mp, CTFRZ, dels, soil_zse, soil_swilt,           &
...
-  ! case of sublimation of snow overwrites later
+!|     - If the surface has snow cover, or `ssnow_potev`<0 **and** the soil
+!        temperature is below freezing (frost), then the latent heat flux represents a
+!        conversion between solid and vapour phases of water, \(c_{ls}\)=1.1335.
+!        The first estimate for the soil latent heat flux is updated by the change
+!        in value of \(c_{ls}\).
+!
   IF (ssnow_snowd(j) >=0.1 ) THEN
      ssnow_cls(j) = 1.1335
      canopy_fess(j) = ssnow_cls(j)*ssnow_potev(j)
```
ccarouge, line 222: This has nothing to do with the documentation. Why isn't there "ssnow_wetfac" here to calculate canopy_fess?

src/science/canopy/cbl_latent_heat.F90
```diff
@@ -82,34 +192,50 @@ SUBROUTINE Latent_heat_flux( mp, CTFRZ, dels, soil_zse, soil_swilt,           &
...
 
-  !Ticket 137 - case iv) deposition of frost onto frozen soil, no snow
+!|     - For case 4, deposition of frost onto frozen surface (temperature below
+!        freezing) occurs even if there is no snow - there is no limit
+!        on the magnitude of the soil latent heat flux.
+!
   IF (ssnow_snowd(j) < 0.1 .AND. ssnow_potev(j) < 0. .AND. &
        ssnow_tss(j)<CTFRZ) THEN
      ssnow_cls(j)=1.1335
      canopy_fess(j) = ssnow_cls(j)*ssnow_potev(j)
```
ccarouge, line 232: This has nothing to do with the documentation. Why isn't "ssnow_wetfac" used to calculate canopy_fess?

ccarouge (approved):
I have 2 comments that don't have anything to do with documentation. I just can't make sense of why it isn't a bug.

src/science/canopy/cbl_latent_heat.F90
har917: ssnow_wetfac quantifies the role that the moisture in the surface layer of the soil has in setting evaporation.  It lies between 0 and 1 (i.e. 0 if there's insufficient water to permit any evaporation, to 1 if there's sufficient water in the soil that it behaves like a lake).  With a snow layer, the sublimation flux comes from the snow pack - and this is sully wet (i.e. there's is no limitation).  Given that ssnow_wetfac may not be evaluated properly when snow is present the code just enforces ssnow_wetfac = 1 when calculating canopy_fess.

Importantly canopy_fess is what controls the change in water from the soil/snow pack over the time step - not ssnow_wetfac.

src/science/canopy/cbl_latent_heat.F90
har917: see above - the slight difference here is that the snow pack/soil doesn't play any role in controlling a flux of water when the flux is from the atmosphere onto the soil (it's the atmosphere water content that controls things).  In this set up ssnow_wetfac does get set to 1 earlier on in the SUBROUTINE.

src/science/canopy/cbl_latent_heat.F90
ccarouge: OK. I see. And I see how to make this clearer. To me, it would be clearer if ssnow_wetfac and ssnow_cls were set correctly for all points first. Then have: `canopy_fess = ssnow_wetfac*ssnow_cls*ssnow_potev`. Then have the loop to apply the various limits. So this case and the next one wouldn't be "special" cases but simply the normal equation. It's possible to avoid resetting these values by setting the special cases correctly in the loop.

Or is it a problem to reset ssnow_wetfac to 1 for these points? Or is it a problem not to reset it? Because ssnow_wetfac is used afterwards in the code and without the `if` conditions. So for some points with snow, we enforce ssnow_wetfac=1 here but use a different value later in the code (the one that was maybe wrongly evaluated by CABLE). For example, line 808 in cable_canopy.F90 and after, there is no snow case there.

har917: This section of code has evolved over a period of time with individuals making incremental changes - it's likely that a rewrite/restructure is worth it.  In this case what you suggest appears sensible - however we'd have to check against the various uses/setting of ssnow%wetfac - in define_canopy and soil_snow.  

The place you've identified (qscreen) isn't too much of an issue (a diagnostic that very few people look at).   

The bigger issue is likely the use of wetfac when setting the sensitivity terms (dfe_ddq) and what happens in the case of a change of sign of fes during the time step.  Again, unlikely to be much of an issue for most researchers - but is of relevance to ACCESS.  At the moment I think the code is buggy but internally consistent - changing things to fix the bug(s) could make consistency hard to achieve and might cause problems (or may not, it could solve the remaining mass conservation quirks).

Possibly worth a comment in the docs - that code will have to be looked at for JAC anyway.,

ccarouge: OK. Got it.

PR #112, opened 2022-12-23
Improve guidance for new comers

### LINKED ISSUE #111: Better guidance for new members wanting to contribute to the documentation
We need an index page to the developer guide so new contributors can navigate the guide more easily.

We need more information on how to find the various parts of the documentation especially the split between API and generic documentation.

ccarouge:
# CABLE

## Description

Fixes #111

Describes:

- the location of the scientific documentation
- the location of the other documentation
- how to find the correct file for editing

Add warnings not to include code changes and the explanation to add "Warning" to scientific documentation when a code deficiency is identified.

## Type of change

- [ ] Bug fix

## Checklist


documentation/docs/developer_guide/documentation_guidelines/index.md
```diff
@@ -1,18 +1,37 @@
...
 - the standalone guides such as the User Guide and the Developer Guide. It is written using [Material for MkDocs][material].
 
-All documentation is located in the [CABLE GitHub repository][cable-repo]. Both tools use [Markdown][md-cheatsheets] and [LaTeX Mathematics][latex-maths] for formatting. You can find a cheatsheet for FORD [on this page][ford-cheatsheets].
+Both tools use [Markdown][cheat-sheets] and [LaTeX Mathematics][latex-maths] for formatting. 
+
+## Documentation workflow
+All documentation is located in the [CABLE GitHub repository][cable-repo]. To update the documentation, you need to follow a traditional GitHub contribution process: open an issue and a new branch; work on your changes; submit a pull request for review before publication. The process is explained in more details [on this page][git-process].
+
+## Scientific documentation
+The *scientific documentation* should be added directly into the source code available under the `src/` directory. Please use [these guidelines][api-guidelines] to structure the documentation. You can find a cheatsheet for FORD [on this page][cheat-sheets].
```
tammasloughran, line 16: Move scientific description section of user guide to the top level.

Also what will be in the "Scientific description" section? General description like PFTs? Model layers? Should we say somewhere something like: documentation that does not belong in the source code/API would go there?

documentation/docs/developer_guide/documentation_guidelines/index.md
```diff
@@ -1,18 +1,37 @@
 # Documentation guidelines
 
+First of all, you should become a member of the CABLE-LSM GitHub organisation prior to contribute to the CABLE's documentation. Please ask for membership [on this issue][cable-lsm-join]. All guidelines are written with the assumption you are a member of the organisation.
+
 The documentation for CABLE has two parts:
 
 - inline documentation in the source code. This documentation includes the detailed scientific documentation. It is written using [FORD][ford].
```
tammasloughran, line 7: "and implementation details" ?

tammasloughran (commented):
Also there are some other comments on slightly related stuff:

`documentation/docs/user_guide/uber_quick_guide.md`

- The "more detail" section of the Uber quick guide mentions ACCESS-ESM and -CM versions of CABLE. I'm assuming there will be github tags associated with the ACCESS versions of CABLE. Eventually these should be linked here.
- Regarding hyperlinks, a common frustration with the old documentation is that hyperlinks were often broken pointers. There isn't much that can be done about this for links to external material, but is there any way for Mkdocs to check for broken pointers to it's own content? Maybe checking for broken links should be a github workflow in the future?
- "NB: spinning up the model has been switched off in cable.nml (spinup=.FALSE.)" Why is this important? When would we need to enable spinup, and when should it be disabled?

`documentation/docs/user_guide/installation.md`

- Regarding requirements: what about openmpi or some other mpi library? The build script is written in bash. The netcdf library is actually split into netcdf c library (netcdf) and netcdf fortran (netcdff). Most distributions of netcdf do not include the fortran libraries.
- "To install the latest version of CABLE, you simply need to checkout the trunk" This is not quite installing, just say "to download the source code".
- "In this section, we discuss building CABLE offline for single-site and regional/global applications." So is the serial build for single site mode and the parallel build is for global/regional runs? This is not clear.

documentation/docs/developer_guide/documentation_guidelines/index.md
ccarouge: What will be in the scientific description section of the user guide is still to be determined. Probably once we have written the detailed documentation in the source code, it would be good to have a short overview of the science covered by CABLE. But that's still open to debate.

ccarouge:
@tammasloughran Thanks for the review. To reply to some of your questions:

> but is there any way for Mkdocs to check for broken pointers to it's own content?
It's already in place. Every commit to the repository triggers a link-checker on the mkdocs documentation. It will need to be updated so it's triggered only when the commit changes files under docs/. And it will also need to be started at regular intervals so we know if a link breaks for another reason than an update. All this is planned and is now in an issue (#115 )

For the other comments, I have put them in issues as it is general feedback on the documentation not pertinent to the current PR.

documentation/docs/developer_guide/documentation_guidelines/index.md
ccarouge: I think I will leave the Scientific description section where it is for now. Let's see what goes in there first and then we'll decide where to put it.

tammasloughran (approved):
All good.

PR #193, opened 2023-12-11
Deprecate `serial_cable` and `parallel_cable` scripts

### LINKED ISSUE #189: Move `serial_build` and `parallel_build` functionality into Makefile
The serial_build and parallel_build build scripts can go in the Makefile by creating [rules](https://www.gnu.org/software/make/manual/make.html#Rule-Introduction) specific to serial or mpi. This simplifies the build system by reducing the number of scripts. See [here](https://github.com/CABLE-LSM/CABLE/commit/51f54d249ea24fc0551d1f62522f21b4aa411ebb) as an example.

SeanBryan51:
This change simplifies the CABLE offline build system by deprecating the `serial_cable` and `parallel_cable` build scripts and moving their functionality into the Makefile.

Replace `ls *.o` in `serial_cable` and `parallel_cable` scripts with `$(OBJS)` in the Makefile where `OBJS` is a list of object files common to both serial and MPI. This is done so that:
1. We leverage `make` syntax to simplify the rules for `serial` and `mpi` targets.
2. We can replace the `all` target (which contains a list of object files as dependencies) with `$(OBJS)`. 
2. We clean up the existing definition of `LSRC` in the Makefile and use this to generate a list of the required object files.
See [#193][PR] for a sorted diff of the new `LSRC` variable against the previous `LSRC` and against the list of object file dependencies in the `all` target.

Tools for testing binary equivalence of executables (see `ldd`, `nm` and `objdump` tools), show that the serial and MPI executables are equivalent to their respective executable in the main branch.

Fixes #189

[PR]: https://github.com/CABLE-LSM/CABLE/pull/193

----
📚 Documentation preview 📚: https://cable--193.org.readthedocs.build/en/193/


src/offline/serial_cable
SeanBryan51, line 1: There are references to this file elsewhere in the code base. These should be addressed.

src/offline/parallel_cable
SeanBryan51, line 1: There are references to this file elsewhere in the code base. These should be addressed.

src/offline/Makefile
```diff
@@ -165,7 +168,25 @@ CINC = -I$(NCMOD)
 	$(FC) $(CFLAGS) $(CINC) -c $<
 
 # default target by convention is ``all''
-all : 
+all: serial
+
+serial: supr = -warn nostderrors -diag-disable 10145
```
SeanBryan51, line 173: We probably shouldn't use `supr` here and instead append the flags to either `CFLAGS` or `LDFLAGS`.

src/offline/Makefile
```diff
@@ -165,7 +168,25 @@ CINC = -I$(NCMOD)
...
+mpi: FOBJS = $(shell ls *.o)
+mpi: cable_mpidrv.F90 cable_mpicommon.F90 pop_mpi.F90 cable_mpiworker.F90 cable_mpimaster.F90 common
+	rm -f cable-mpi
+	rm -f cable_mpi*.o
+	rm -f pop_mpi.o
+	$(FC) $(supr) $(CFLAGS) $(LDFLAGS) -c cable_mpicommon.F90 $(FOBJS) $(CINC)
+	$(FC) $(supr) $(CFLAGS) $(LDFLAGS) -c pop_mpi.F90 cable_mpicommon.o $(FOBJS) $(CINC)
+	$(FC) $(supr) $(CFLAGS) $(LDFLAGS) -c cable_mpiworker.F90 cable_mpicommon.o pop_mpi.o $(FOBJS) $(CINC)
+	$(FC) $(supr) $(CFLAGS) $(LDFLAGS) -c cable_mpimaster.F90 cable_mpicommon.o cable_mpiworker.o pop_mpi.o $(FOBJS) $(CINC)
+	$(FC) $(supr) $(CFLAGS) $(LDFLAGS) -o $(PROG_MPI) cable_mpidrv.F90 cable_mpicommon.o cable_mpimaster.o cable_mpiworker.o pop_mpi.o $(FOBJS) $(CINC) $(LD)
```
SeanBryan51, line 189: We should leverage Makefile syntax here so that:
1. We do not have to remove object files or executable files.
2. We have a single build command for the serial and mpi targets that contains the correct list of object files.

SeanBryan51 (commented):
@harshula has tools for checking binary equivalence between executable files, these should be used so we can verify we haven't introduced any unwanted changes from refactoring.

JhanSrbinovsky (approved):
Docs on building might need corresponding revision

src/offline/Makefile
```diff
@@ -350,8 +359,18 @@ landuse3.o: landuse3.F90 cable_define_types.o casa_dimension.o landuse_constant.
 
 landuse_inout.o: landuse_inout.F90 cable_define_types.o landuse_constant.o landuse3.o cable_common.o cable_iovars.o cable_abort.o
 
+# Dependencies for serial compilation:
```
ccarouge, line 362:
```suggestion
# Dependencies for serial-specific files:
```
Makes it clear why it only includes one file

src/offline/Makefile
```diff
@@ -350,8 +359,18 @@ landuse3.o: landuse3.F90 cable_define_types.o casa_dimension.o landuse_constant.
...
+cable_driver.o: cable_driver.F90 common
+
+# Dependencies for mpi compilation:
```
ccarouge, line 365:
```suggestion
# Dependencies for mpi-specific files:
```

ccarouge (approved):
This looks good to me. A couple of minor suggestions on the comments in the Makefile to make them more descriptive.

ccarouge (approved):
Looks good to me.

Just curious: why were added dependencies needed at this point of the work? To fix the order of compilation?

SeanBryan51:
### Changes to `LSRC` variable
A sorted diff of the old `LSRC` variable vs new is as follows:
```diff
--- /home/189/sb8430/LSRC.txt	2023-12-18 13:07:13.869287000 +1100
+++ /home/189/sb8430/LSRC_new.txt	2023-12-19 10:08:10.295839000 +1100
@@ -1,6 +1,4 @@
 bgcdriver
-bgcdriver
-biogeochem_casa
 biogeochem_casa
 cable_abort
 cable_air
@@ -45,7 +43,6 @@
 casa_cnp
 casa_dimension
 casa_feedback
-casa_feedback
 casa_inout
 casa_ncdf
 casa_offline_inout
@@ -55,17 +52,15 @@
 casa_readbiome
 casa_rplant
 casa_sumcflux
-casa_sumcflux
 casa_variable
+cbl_albedo
 cbl_conductivity
 cbl_dryLeaf
 cbl_friction_vel
 cbl_fwsoil
 cbl_GW
-cbl_GW
 cbl_hyd_redistrib
 cbl_init_radiation
-cbl_LAI_canopy_height
 cbl_latent_heat
 cbl_model_driver_offline
 cbl_Oldconductivity
@@ -106,7 +101,6 @@
 POP
 pop_constants
 pop_def
-POPdriver
 pop_io
 POPLUC
 pop_types
```
This change adds one missing file (`cbl_albedo.F90`) and removes two files (`POPdriver.F90` which no longer exists and `cbl_LAI_canopy_height.F90` which is not required for the offline configuration).

A sorted diff of the list of object file dependencies in the `all` target vs the new `LSRC` variable is as follows:
```diff
--- /home/189/sb8430/common.txt	2023-12-18 13:03:52.769319000 +1100
+++ /home/189/sb8430/LSRC_new.txt	2023-12-19 10:08:10.295839000 +1100
@@ -16,6 +16,7 @@
 cable_iovars
 cable_LUC_EXPT
 cable_maths_constants_mod
+cable_metutils
 cable_namelist_input
 cable_other_constants_mod
 cable_output
@@ -54,11 +55,18 @@
 casa_variable
 cbl_albedo
 cbl_conductivity
+cbl_dryLeaf
+cbl_friction_vel
+cbl_fwsoil
 cbl_GW
 cbl_hyd_redistrib
 cbl_init_radiation
+cbl_latent_heat
 cbl_model_driver_offline
 cbl_Oldconductivity
+cbl_photosynthesis
+cbl_pot_evap_snow
+cbl_qsat
 cbl_radiation
 cbl_remove_trans
 cbl_rhoch
@@ -75,13 +83,16 @@
 cbl_soilfreeze
 cbl_soilsnow_data
 cbl_soilsnow_init_special
-cbl_soilsnow_init_special
 cbl_soilsnow_main
 cbl_spitter
 cbl_stempv
+cbl_SurfaceWetness
 cbl_surfbv
 cbl_thermal
 cbl_trimb
+cbl_wetleaf
+cbl_within_canopy
+cbl_zetar
 grid_constants_cbl
 landuse3
 landuse_constant
```
The `OBJS` variable (derived from `LSRC`) includes files that were missing from the all target. This is because `OBJS` must contain the source files that are required at the linking step, otherwise the compiler will complain. It is likely that files which were missing were being compiled somewhere along the dependency chain. The `OBJS` variable also lists the source files in a different order to how it was in the all target. This changes the order of how source files are compiled (hence the changes to `*.o` dependency rules). Note, this does not ensure that target dependencies for all `*.o` files are correct. Only the minimal required changes to make the compilation work have been made.

SeanBryan51:
> Looks good to me.
> 
> Just curious: why were added dependencies needed at this point of the work? To fix the order of compilation?

The changes to dependencies was a result of using the `$(OBJS)` instead of the `all` target. I have updated the PR description and provided an explanation above of the diffs between `OBJS` and the object file dependencies in `all`.

harshula (approved):
Well done! Good work simplifying the build system and making it more maintainable.

PR #197, opened 2024-01-03
Remove `-O0` flag from `LDFLAGS`

### LINKED ISSUE #196: Spurious usage of `-O0` in `LDFLAGS`
In `build3.sh`, the `-O0` flag (which disables optimisations) is included in the `LDFLAGS` variable. See below:

https://github.com/CABLE-LSM/CABLE/blob/1c8a2cb763f0a797f78f23a3dd8296277f26bb86/src/offline/build3.sh#L22

The flag should be removed from `LDFLAGS` because:
1. `-O0` is a compiler flag and not an `ld` flag. If the developer wants to set `-O0`, this should be done by setting it in `CFLAGS`.
2. Adding `-O0` to `LDFLAGS` has the effect that optimisations are disabled when compiling parts of the source code (see [these lines](https://github.com/CABLE-LSM/CABLE/blob/1c8a2cb763f0a797f78f23a3dd8296277f26bb86/src/offline/Makefile#L162-L182) in the Makefile).  Adding in the flag may have had the desired outcome for a developer who added it in (e.g. for debugging purposes), but it is unlikely they wanted the flag to be applied for only a subset of source files and is likely a mistake.
3. The presence of the flag results in the following Makefile rules to become complicated:

https://github.com/CABLE-LSM/CABLE/blob/1c8a2cb763f0a797f78f23a3dd8296277f26bb86/src/offline/Makefile#L171-L182

Removing the flag will simplify the Makefile so that we only use the default suffix rule when compiling object files (the other flags defined in `LDFLAGS` and `LD` are redundant here as `-c` disables linking).

I have not yet tried to run CABLE with the flag removed, but it should run as per usual (if not slightly more optimised as `-O2` will be enabled when compiling all source files).

- [x] Sean: to confirm removing the flag does not produce any undesirable side effects when running CABLE.

SeanBryan51:
In `build3.sh`, the `-O0` flag (which disables optimisations) is included in the `LDFLAGS` variable. This change removes the `-O0` flag from `LDFLAGS` so that:

1. Compiler specific flags (non `ld` flags) are removed from `LDFLAGS`
2. `-O0` flag is not applied for only a subset of source files.
3. The Makefile rules can be simplified.

See #196 for more details on why the flag should be removed.

Tools for testing binary equivalence (see `ldd`, `nm` and `objdump` tools) show that the following build artifacts differ when comparing against the main branch:

- `cable` (serial executable)
- `cable-mpi` (MPI executable)
- `cable_mpicommon.o`
- `cable_mpimaster.o`
- `cable_mpiworker.o`
- `pop_mpi.o`

Regression tests using [benchcab](https://github.com/CABLE-LSM/benchcab) (bitwise comparison of model output via `nccmp`) show that model output is bitwise identical between the current branch and the main branch for serial and MPI model runs.

Fixes #196

## Type of change

- [X] Bug fix

----
📚 Documentation preview 📚: https://cable--197.org.readthedocs.build/en/197/

### DIFF
--- a/src/offline/Makefile
+++ b/src/offline/Makefile
@@ -168,18 +168,10 @@ serial: cable_driver.F90 $(OBJS)
 mpi: cable_mpidrv.F90 cable_mpicommon.o cable_mpimaster.o cable_mpiworker.o pop_mpi.o $(OBJS)
 	$(FC) $(SUPPRESS_FLAGS) $(CFLAGS) $(LDFLAGS) -o $(PROG_MPI) $^ $(CINC) $(LD)
 
-# TODO(Sean): passing LDFLAGS to the compiler ensures bitwise reproducibility
-# for this PR (#193). This is because LDFLAGS contains the -O0 flag (see
-# build3.sh). Ideally, LDFLAGS should not contain -O0 and this should be set in
-# CFLAGS as this would let us use the default suffix rule for *.o files.
 cable_mpicommon.o: cable_mpicommon.F90 $(OBJS)
-	$(FC) $(CFLAGS) $(CINC) $(LDFLAGS) -c $<
 pop_mpi.o: pop_mpi.F90 cable_mpicommon.o $(OBJS)
-	$(FC) $(CFLAGS) $(CINC) $(LDFLAGS) -c $<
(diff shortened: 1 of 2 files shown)


ccarouge (approved):
All good.

PR #199, opened 2024-01-22
Fix LDFLAGS definition and usage

### LINKED ISSUE #198: `ld` flags used in offline build are incorrect
`ld` flags are currently specified incorrectly. For example:
1. `-lnetcdf` is not required since it is a dependency of the netCDF Fortran library - only `-lnetcdff` is needed. **Note**: if linking against a static library then `-lnetcdf` as well as other libraries may be required - see [1.7 Compiling and Linking with the NetCDF Library](https://docs.unidata.ucar.edu/netcdf-fortran/current/f90-use-of-the-netcdf-library.html#:~:text=returned%20status%20value.-,1.7%20Compiling%20and%20Linking%20with%20the%20NetCDF%20Library,-Details%20of%20how)).
https://github.com/CABLE-LSM/CABLE/blob/1c8a2cb763f0a797f78f23a3dd8296277f26bb86/src/offline/build3.sh#L23

2. `ld` flags are being split across two variables: `LDFLAGS` and `LD` when there should only be one (`LD` typically refers to the path  to the `ld` executable).
https://github.com/CABLE-LSM/CABLE/blob/1c8a2cb763f0a797f78f23a3dd8296277f26bb86/src/offline/build3.sh#L22-L23
Having the flags split across two variables allows for the flags to be passed to the compiler in an inconsistent way. For example, the flags in `LDFLAGS` (and `LD` for the serial case) are specified before the source files:
https://github.com/CABLE-LSM/CABLE/blob/1c8a2cb763f0a797f78f23a3dd8296277f26bb86/src/offline/Makefile#L163-L169
`ld` flags should generally be specified after the source files (see this [SO post](https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc) for justification).

A future proof solution would be to not hard code linker flags in the build system and instead use a tool such as [`pkg-config`](https://linux.die.net/man/1/pkg-config) to get the correct flags for a given dependency (see [here](https://docs.unidata.ucar.edu/netcdf-fortran/current/f90-use-of-the-netcdf-library.html#:~:text=If%20it%20is%20installed%20on,netCDF%20libraries%20using%20pkg%2Dconfig%3A) as an example).

SeanBryan51:
`ld` flags are currently specified incorrectly. For example:
1. `-lnetcdf` is not required since it is a dependency of the netCDF Fortran library - only `-lnetcdff` is needed.
2. `ld` flags are being split across two variables: `LDFLAGS` and `LD` when there should only be one (`LD` typically refers to the path  to the `ld` executable). Having the flags split across two variables allows for the flags to be passed to the compiler in an inconsistent way.

See #198 for more details.

This change removes the `LD` variable and any hard-coded linker flags or include flags in the Makefile or build script. Linker flags and include flags are replaced with a query to `pkg-config` so that the correct flags are used. Using `pkg-config` also insulates the build system from changes to library versions and dependencies.

~~This PR should ideally be done after #197 is merged so that we can verify compiled object files have not been altered as a result of this change.~~

Compiled object files are identical between this branch and the main branch. However, the executables differ between the two branches due to changes in the compiler arguments at the linking step.

Regression tests using [benchcab](https://github.com/CABLE-LSM/benchcab)* (bitwise comparison of model output via nccmp) show that model output is bitwise identical between the current branch and the main branch for serial and MPI model runs.

*executables were built manually for the spatial case (see this related issue: https://github.com/CABLE-LSM/benchcab/issues/244).

Fixes #198

----
📚 Documentation preview 📚: https://cable--199.org.readthedocs.build/en/199/

### DIFF
--- a/src/offline/Makefile
+++ b/src/offline/Makefile
@@ -147,26 +147,23 @@ PROG_MPI = cable-mpi
 # from the compiled object files
 OBJS := ${LSRC:.F90=.o}
 
-#compiler switches and flags
-CINC = -I$(NCMOD)
-
 #suffixes we use
 .SUFFIXES:
 .SUFFIXES: .F90 .o
 #default rules for these suffixes
 .F90.o:
-	$(FC) $(CFLAGS) $(CINC) -c $<
+	$(FC) $(CFLAGS) -c $<
 
 SUPPRESS_FLAGS := -warn nostderrors -diag-disable 10145
 
 # the first target is the default target
 .PHONY: serial
 serial: cable_driver.F90 $(OBJS)
-	$(FC) $(SUPPRESS_FLAGS) $(CFLAGS) $(LDFLAGS) $(LD) -o $(PROG_SERIAL) $^ $(CINC)
+	$(FC) $(SUPPRESS_FLAGS) $(CFLAGS) -o $(PROG_SERIAL) $^ $(LDFLAGS)
 
 .PHONY: mpi
 mpi: cable_mpidrv.F90 cable_mpicommon.o cable_mpimaster.o cable_mpiworker.o pop_mpi.o $(OBJS)
-	$(FC) $(SUPPRESS_FLAGS) $(CFLAGS) $(LDFLAGS) -o $(PROG_MPI) $^ $(CINC) $(LD)
+	$(FC) $(SUPPRESS_FLAGS) $(CFLAGS) -o $(PROG_MPI) $^ $(LDFLAGS)
 
 cable_mpicommon.o: cable_mpicommon.F90 $(OBJS)
 pop_mpi.o: pop_mpi.F90 cable_mpicommon.o $(OBJS)
--- a/src/offline/build3.sh
+++ b/src/offline/build3.sh
@@ -15,12 +15,14 @@ host_gadi()
    module purge
    module add intel-compiler/2019.5.281
    module add netcdf/4.6.3
+   # This is required so that the netcdf-fortran library is discoverable by
+   # pkg-config:
+   prepend_path PKG_CONFIG_PATH "${NETCDF_BASE}/lib/Intel/pkgconfig"
 
-   export NCDIR=$NETCDF_ROOT'/lib/Intel'
-   export NCMOD=$NETCDF_ROOT'/include/Intel'
-   export CFLAGS='-O2 -fp-model precise '
-   export LDFLAGS='-L'$NCDIR
-   export LD='-lnetcdf -lnetcdff'
+   if ! pkg-config --exists netcdf-fortran; then
+       echo -e "\nUnable to find netcdf-fortran via pkg-config. Exiting...\n"
+       exit 1
+   fi
 
    if [[ $1 = 'mpi' ]]; then
       module add intel-mpi/2019.5.281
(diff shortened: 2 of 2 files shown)


harshula (approved):

ccarouge:
@harshula Is `pkg-config` available anywhere? Do system administrators have to decide to set it up? I want to make sure this would work on other systems as well without too many modifications.

harshula:
Hi, Good question! My understanding is that it is a default package on GNU/Linux distros.

SeanBryan51:
This is more of a comment around the CMake implementation (#200) but it is still relevant. If `pkg-config` is not a standard tool on most systems, should parts of the build which depend on `pkg-config` be guarded by an "if on Gadi" check? Currently, the CMake implementation needs `pkg-config` to link against the netcdf-fortran library. We could possibly look into linking against netcdf dependency using [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) directly instead of relying on `pkg-config`?

harshula:
With Spack, it's as simple as, `depends_on("pkgconf", type="build")`. Have you come across any systems that Cable is compiled on that does not have `pkg-config`? If you are worried by an unlikely scenario, try `pkg-config` and then fallback to something else.

SeanBryan51:
It also looks like the `find_package(NetCDF)` approach will likely require maintaining a `FindNetCDF.cmake` file which is not ideal: https://stackoverflow.com/questions/41571163/write-cmake-find-netcdf-for-c-and-fortran-library. Using `pkg-config` seems like a better alternative to this.

ccarouge (approved):
I am approving but I need to check the portability of `pkg_config`. We have at least one active developer in Europe who does not use Gadi. And I have no idea what's available on their system.
Depending on what comes out of the enquiry, further adaptation might be required.

Should the `prepend_path` be guarded by a "if gadi" condition? Or do we only worry about that for the CMake version.

SeanBryan51:
> Should the `prepend_path` be guarded by a "if gadi" condition? Or do we only worry about that for the CMake version.

It is implied that anything under `host_gadi` is specific to Gadi.

harshula:
On Rocky Linux:
```
[root@3c6c3cddd8f4 /]# cat /etc/os-release 
NAME="Rocky Linux"
VERSION="8.9 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.9"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.9 (Green Obsidian)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:8:GA"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
SUPPORT_END="2029-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-8"
ROCKY_SUPPORT_PRODUCT_VERSION="8.9"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.9"
```

`pkgconf` is installed by the `Development Tools` group:
```
[root@3c6c3cddd8f4 /]# dnf group info "Development Tools"
Last metadata expiration check: 0:28:59 ago on Thu Feb  1 05:30:48 2024.

Group: Development Tools
 Description: A basic development environment.
 Mandatory Packages:
   autoconf
   automake
   binutils
   bison
   flex
   gcc
   gcc-c++
   gdb
   glibc-devel
   libtool
   make
   pkgconf
   pkgconf-m4
   pkgconf-pkg-config
   redhat-rpm-config
   rpm-build
   rpm-sign
   strace
 Default Packages:
   asciidoc
   byacc
   ctags
   diffstat
   elfutils-libelf-devel
   git
   intltool
   jna
   ltrace
   patchutils
   perl-Fedora-VSP
   perl-Sys-Syslog
   perl-generators
   pesign
   source-highlight
   systemtap
   valgrind
   valgrind-devel
 Optional Packages:
   cmake
   expect
   rpmdevtools
   rpmlint
```
On RHEL clones, it's safe to conclude that pkg-config will be installed by default alongside well-known Development Tools.

ccarouge:
> It is implied that anything under `host_gadi` is specific to Gadi.

Aahh, the diff view in the review didn't show the `host_gadi` line.

SeanBryan51:
> Since I hadn't seen the almost entirety of the build script is within a `host_gadi` section, this seems to make my concerns about `pkg_config` irrelevant. On another machine, people will have to write their own `host_x` anyway and then changing `pkg_config` for something else in that `host_x` section is fine. Or are there planned changes that affect that conclusion?

@ccarouge it will be relevant for CMake as `pkg-config` is a required dependency, in [CMakeLists.txt](https://github.com/CABLE-LSM/CABLE/blob/188-cmake-based-build-system/CMakeLists.txt):
```cmake
find_package(PkgConfig REQUIRED)
```

Issue #192, opened 2023-12-11
Solve dependency problems between targets/dependencies and ensure parallel builds succeed


SeanBryan51:
Many of the rules in the current Makefile are missing dependencies. This causes the compilation to fail when we compile in parallel via `make -j`.

Issue #191, opened 2023-12-11
Improve build3.sh CLI by creating switches for serial Vs mpi and debug Vs non-debug


SeanBryan51:
For example, it is currently not possible to enable debug mode and mpi mode simultaneously.

Issue #190, opened 2023-12-11
Remove temporary source directory by improving the Makefile


SeanBryan51:
The current build system copies the source code into a temporary directory and executes make in that directory. This can make debugging confusing as debug information will point to the source code in the temporary directory rather than the actual source code. The Makefile should be able to handle the compilation without the need for a temporary directory (see a draft implementation of this [here](https://github.com/CABLE-LSM/CABLE/blob/c35dc433c17e77e8d4e840f63ef9acedaba297db/src/offline/TestMakefile)). Removing this from the build system will also simplify the build3.sh build script.

PR #200, opened 2024-01-22
Add CMake based build system for CABLE offline

### LINKED ISSUE #188: Build system improvements for CABLE offline
In preparation for building CABLE with [spack](https://spack.io/), @harshula, @ccarouge and I have found areas where the current build system for CABLE offline can be improved.

- [x] #189
    - The `serial_build` and `parallel_build` build scripts can go in the Makefile by creating [rules](https://www.gnu.org/software/make/manual/make.html#Rule-Introduction) specific to serial or mpi. This simplifies the build system by reducing the number of scripts. See [here](https://github.com/CABLE-LSM/CABLE/commit/51f54d249ea24fc0551d1f62522f21b4aa411ebb) as an example.
- [ ] #190
    - The current build system copies the source code into a temporary directory and executes `make` in that directory. This can make debugging confusing as debug information will point to the source code in the temporary directory rather than the actual source code. The Makefile should be able to handle the compilation without the need for a temporary directory (see a draft implementation of this [here](https://github.com/CABLE-LSM/CABLE/blob/c35dc433c17e77e8d4e840f63ef9acedaba297db/src/offline/TestMakefile)). Removing this from the build system will also simplify the `build3.sh` build script.
- [ ] #191
    - For example, it is currently not possible to enable debug mode and mpi mode simultaneously.
- [ ] #192
    - Many of the rules in the current Makefile are missing dependencies. This causes the compilation to fail when we compile in parallel via `make -j`.

SeanBryan51:
In preparation for building CABLE with [spack](https://spack.io/), @harshula, @ccarouge and I have found areas where the current build system for CABLE offline can be improved (see #188 for more details). This change adds a [CMake](https://cmake.org/) based build system to address these issues:

- #190
  Although this problem can be solved in the Makefile by specifying the relative paths to each source file, it requires updating all the Makefile Fortran module dependency rules which is a pain. CMake automatically handles Fortran module dependencies out of the box so doing this with CMake is much easier.
- #191 
  When invoking `cmake`, custom options, compilers and compiler flags can be configurable by setting the appropriate variables via the [`-D` flag](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-D), for example: `cmake -DCMAKE_Fortran_COMPILER="mpif90" -DCMAKE_Fortran_FLAGS="-O2 -fp-model precise" -S . -B build`. These can be used for doing MPI only builds and enabling specific debug flags.
- #192 
  As mentioned above, CMake automatically handles Fortran module dependencies out of the box. Parallelised builds can be done by specifying the `-j` flag to `cmake`.

The following pull requests should be merged before this pull request so that we can demonstrate binary equivalence between the executables compiled by both Makefile and CMake builds. This ensures no unwanted changes are introduced by the transition.

- [x] #197
- [x] #199

Edit: the above PR's have been merged - the serial and MPI binaries built by CMake (in release mode) are equivalent to the binaries built by the Makefile based build system in the main branch.

Fixes #188, #190, #191, #192

----
📚 Documentation preview 📚: https://cable--200.org.readthedocs.build/en/200/


documentation/docs/user_guide/installation.md
SeanBryan51, line 1: Should we add more detailed documentation describing the build system under `docs/developer_guide`? We should also:

- [ ] Describe how to add source files.
- [ ] Describe how to enable/disable specific compiler flags.

CMakeLists.txt
SeanBryan51, line 1: TODO after this pull request:

- [ ] We should list the source files in alphabetical order (the current order is so that we preserve binary equivalence with the Makefile build).
- [ ] Use [`find_package(MPI REQUIRED)`](https://cmake.org/cmake/help/latest/module/FindMPI.html) instead of specifying `mpif90` when invoking `cmake` for MPI case.

CMakeLists.txt
```diff
@@ -0,0 +1,168 @@
...
+
+# third party libs
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(NETCDF REQUIRED IMPORTED_TARGET "netcdf-fortran")
+
+set(CABLE_INTEL_Fortran_FLAGS -fp-model precise)
+set(CABLE_INTEL_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0)
+set(CABLE_INTEL_Fortran_FLAGS_RELEASE -O2)
+
+set(CABLE_EXE_NAME "$<IF:$<BOOL:${CABLE_MPI}>,cable-mpi,cable>")
```
SeanBryan51, line 18: Maybe add a comment here referencing [CMake generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html).

CMakeLists.txt
```diff
@@ -0,0 +1,171 @@
+# Set minimum CMake version to version on Gadi:
```
SeanBryan51, line 1: Multiple CMake versions are available on Gadi: see https://opus.nci.org.au/display/Help/CMake

documentation/docs/user_guide/installation.md
```diff
@@ -2,7 +2,8 @@
 
 To install CABLE you need to have the following already installed on your system:
 
-- git
+- `git`
+- `cmake`
 - a Fortran compiler
 - the netCDF library
```
SeanBryan51, line 8: Add `pkg-config` or `pkgconf` to dependencies.

SeanBryan51:
Note, some links in the documentation are currently broken as they point to source files that were added in this pull request. To see the source files, replace `main` with `188-cmake-based-build-system` in the url.

SeanBryan51:
@ccarouge is it worth supporting both Makefile and CMake based build systems? Or should we deprecate the Makefile build system in this pull request?

build.bash
```diff
@@ -0,0 +1,81 @@
...
+        module add intel-mpi/2019.5.281
+    fi
+fi
+
+if [[ -n $clean ]]; then
+    rm -r build
+fi
+
+cmake_args=()
+if [[ -n $debug ]]; then
```
harshula, line 64: Can you do this inside the switch statement above?

build.bash
```diff
@@ -0,0 +1,81 @@
...
+    cmake_args+=(-DCMAKE_BUILD_TYPE=Debug)
+else
+    cmake_args+=(-DCMAKE_BUILD_TYPE=Release)
+fi
```
harshula, line 68: Can you do this inside the switch statement above?

build.bash
```diff
@@ -0,0 +1,81 @@
...
+if [[ -n $mpi ]]; then
+    cmake_args+=(-DCABLE_MPI="ON")
+    cmake_args+=(-DCMAKE_Fortran_COMPILER="mpif90")
+fi
```
harshula, line 72: Can you do this inside the switch statement above?

build.bash
```diff
@@ -0,0 +1,81 @@
...
+
+cmake_build_args=()
+if [[ -n $verbose ]]; then
+    cmake_build_args+=(-v)
+fi
```
harshula, line 77: Can you do this inside the switch statement above?

harshula, line 77: Can you do this inside the switch/case statement above?

build.bash
```diff
@@ -0,0 +1,81 @@
...
+        -h|--help)
+            show_help
+            exit
+            ;;
+        ?*)
+            printf 'WARN: Unknown option (ignored): %s\n' "$1" >&2
+            ;;
+    esac
+    shift
+done
```
harshula, line 43: As discussed, can there be a way to pass-through switches to CMake?

ccarouge:
> @ccarouge is it worth supporting both Makefile and CMake based build systems? Or should we deprecate the Makefile build system in this pull request?

For the record: we only want to support one build system. So this PR should deprecate the Makefile build system.

ccarouge (approved):
Some minor stuff. I don't think I need to review again but I'd like the `-j` option for cmake to be an option for `build.bash`.

build.bash
```diff
@@ -0,0 +1,70 @@
...
+# Argument parsing adapted and stolen from http://mywiki.wooledge.org/BashFAQ/035#Complex_nonstandard_add-on_utilities
+while [ $# -gt 0 ]; do
+    case $1 in
+        --clean)
+            rm -r build
+            ;;
+        --mpi)
+            mpi=1
+            cmake_args+=(-DCABLE_MPI="ON")
+            cmake_args+=(-DCMAKE_Fortran_COMPILER="mpif90")
```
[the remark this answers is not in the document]
SeanBryan51: I plan to remove these lines in the next changes after this PR when we use `find_package` to link against MPI libraries. But to answer your question, MPI compiler wrappers should work for most fortran compilers. Compilers can be overwritten by setting `FC` environment variable (see https://www.open-mpi.org/doc/v4.0/man1/mpifort.1.php#toc9).

build.bash
```diff
@@ -0,0 +1,70 @@
...
+    # This is required so that the netcdf-fortran library is discoverable by
+    # pkg-config:
+    prepend_path PKG_CONFIG_PATH "${NETCDF_BASE}/lib/Intel/pkgconfig"
+    if [[ -n $mpi ]]; then
+        module add intel-mpi/2019.5.281
+    fi
+fi
+
+cmake -S . -B build "${cmake_args[@]}" &&\
+cmake --build build -j 4 &&\
```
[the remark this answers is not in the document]
SeanBryan51: We can make this configurable by setting this CMake environment variable: https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html

documentation/docs/developer_guide/other_resources/build_system.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Adding external dependencies (libraries)
+
+When linking against external third-party libraries, we strongly recommend not to hard code library paths, include paths, link flags or compiler flags in the CMakeLists.txt file. This is ensures that the build system is portable and insulated from changes to library versions and dependencies.
+
+Instead, CMake should query the required flags for a given library from the library itself. To do this, we recommend using either CMake's [`find_package` mechanism](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Finding%20Packages.html) or a tool such as [`pkg-config`](https://en.wikipedia.org/wiki/Pkg-config) (see the [FindPkgConfig](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) module for using `pkg-config` with CMake.). The dependencies should provide exported targets which can be used via [`target_link_libraries`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html).
```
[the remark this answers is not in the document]
SeanBryan51: What do you mean by using the Gadi setup? This section is about adding 3rd party libraries which is a separate issue to porting to other HPC setups.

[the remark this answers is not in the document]
SeanBryan51: Or do you mean how to make 3rd party libraries discoverable to CMake?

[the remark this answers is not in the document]
ccarouge: I mean how it is done for netcdf in CMakeLists.txt and build.bash. And use the Gadi version as a template as that's the one we test currently.

build.bash
```diff
@@ -0,0 +1,80 @@
...
+
+  To enable more verbose output from Makefile builds, specify the CMake option
+  -DCMAKE_VERBOSE_MAKEFILE=ON when invoking $script_name.
+
+Parallel compilation:
+
+  By default, the number of parallel jobs used in the compilation is
+  $nproc_default. This value can be overwritten by setting the environment
+  variable CMAKE_BUILD_PARALLEL_LEVEL.
+
```
ccarouge, line 34: Can we add this to the main documentation? It can be a note box under installation.md with something like:
The compilation is done in parallel by default. See `build.bash -h` for more information on the default number of parallel jobs and how to customise it.

I think it's important that the default behaviour is clearly stated in the doc.

harshula (approved):
Well done! Potentially a few things we can look at later.

PR #214, opened 2024-02-28
Correction to the numerical value of the specific heat of water

rkutteh:
## Description

Change the descriptions for the specific heat for water, snow, ice and air to:
- remote "capacity"
- add a unit
- specify the temperature for the specific heat of water since it is used at 0°C and not the standard 20°C.

This is done for CABLE and SLI.

Reference:   https://www.engineeringtoolbox.com/specific-heat-capacity-water-d_660.html

- [X ] Bug fix

## Checklist

- [ X] The new content is accessible and located in the appropriate section.
- [X ] I have checked that links are valid and point to the intended content.
- [X ] I have checked my code/text and corrected any misspellings

----
📚 Documentation preview 📚: https://cable--214.org.readthedocs.build/en/214/

### DIFF
--- a/src/params/cable_phys_constants_mod.F90
+++ b/src/params/cable_phys_constants_mod.F90
@@ -35,11 +35,11 @@ REAL, PARAMETER :: grav   = 9.8086        ! gravity acceleration (m/s2)
 REAL, PARAMETER :: rgas   = 8.3143        ! universal gas const  (J/mol/K)
 REAL, PARAMETER :: rmair  = 0.02897       ! molecular wt: dry air (kg/mol)
 REAL, PARAMETER :: rmh2o  = 0.018016      ! molecular wt: water (kg/mol)
-REAL, PARAMETER :: cgsnow = 2090.0        ! specific heat capacity for snow
+REAL, PARAMETER :: cgsnow = 2090.0        ! specific heat for snow (J/kg/K)
 REAL, PARAMETER :: cs_rho_ice = 1.9341e6  !heat capacity * density ice
 REAL, PARAMETER :: cs_rho_wat = 4.218e6   ! heat capacity * density  water
-REAL, PARAMETER :: csice = 2.100e3        ! specific heat capacity for ice
-REAL, PARAMETER :: cswat = 4.182e3        ! specific heat for water (J/kg/K)
+REAL, PARAMETER :: csice = 2.100e3        ! specific heat for ice (J/kg/K)
+REAL, PARAMETER :: cswat = 4.218e3        ! specific heat for water at 0°C (J/kg/K)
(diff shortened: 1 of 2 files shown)


ccarouge:
@rkutteh  Since the specific heat is temperature-dependent, how do you know 4.218 is wrong? 4.218 would be correct at ~0°C according to the website you link. A lot of the usage of `cswat` seems to have to do with snow, so in cold conditions. Except for `gammzz` maybe. Some of the gammzz calculations are done when snow is present and when not present. So it seems we are approximating `cswat` to be temperature-independent in the code, and we have to decide at what temperature we take that value. 

Am I correct here? Or do you see something else?

rkutteh:
@ccarouge @gabsun What you say is possible of course. Typically though when people speak of specific heat of water without mentioning the temperature explicitly, they mean room temperature ~25 C. Either way, if this gets sorted, I think this is a good occasion to start enforcing the inclusion of units in the code (which was missing in this case) and things like temperature in the present case.

ccarouge:
@har917 With Ramzi we are wondering about the correct value for the specific heat of water in CABLE.

It's currently 4.218e3 J/kg/K. Ramzi is wondering if it should be 4.182e3 J/kg/K. The value is temperature-dependent. The current value is valid around 0°C while Ramzi's value is around 25°C. Which one should we use in CABLE?

From a quick look, it seems the specific heat of water is used mainly for snow stuff, pointing to a cold temperature value, although maybe not only. 

Do you know anything about that?

har917:
I don't really know for sure - however, as far as I can tell all of the uses of `cswat` or `Ccswat` relate to occurrences when there's a change in phase going on.  That strongly indicates that we should be using the specific heat of water at freezing. (There is a small impact of atmospheric pressure on this number - since it's enthalpy that's really conserved not energy - see [the wikipedia page](https://en.wikipedia.org/wiki/Specific_heat_capacity), particularly the Theoretical Estimation section)

see also [graph](https://www.chegg.com/homework-help/questions-and-answers/specific-heat-water-first-decrease-increase-rise-temperature-q78479617)

and yes the comments need in the code need to be more precise where sensible.

If we do change the value then this needs to be applied in both the `cable_phys_constant_mod` and `cable_sli_numbers`

ccarouge:
@har917 Thanks Ian. 

Considering your conclusions agree with mine, I would keep the number but add to the descriptions:
- the unit
- the temperature for that value so 0°C.

Do you want to modify the code for this PR accordingly @rkutteh or do you want me to?

ccarouge (approved):
Modified as agreed in the discussion on the PR. Changed "specific heat capacity" to "specific heat" for water, ice, snow and air.

PR #216, opened 2024-03-01
Additional CMake changes

### LINKED ISSUE #215: Additional CMake changes
See #200 for the initial implementation of a CMake based build system. This issue documents changes that should follow after this pull request:

- [ ] List source files in alphabetical order in CMakeLists.txt. The order of source files was chosen so that we could demonstrate binary equivalence in executables between the CMake build and the Makefile build.
- [ ] Use [`find_package(MPI REQUIRED)`](https://cmake.org/cmake/help/latest/module/FindMPI.html) instead of specifying `mpif90` when invoking `cmake` for MPI case. This is done so that
    1. We can create separate targets for the serial and MPI executable and only link against the MPI libraries when needed without relying on the MPI compiler wrapper. This lets us compile all executables (serial and parallel) with a single invocation of CMake.
    2. We can use object libraries to compile the object files common to
both serial and MPI builds. This saves compiling these object files for
each executable (serial and MPI), reducing compilation time.
- [ ] The previous task also lets us remove the MPI option from build script. The same can be achieved by specifying `--target cable-mpi` when invoking CMake.

_Originally posted by @SeanBryan51 in https://github.com/CABLE-LSM/CABLE/pull/200#discussion_r1462694603_

SeanBryan51:
This pull request should follow after the CMake implementation (see #200). This change adds the following build system changes:

- List source files in alphabetical order in CMakeLists.txt. The order of source files was previously chosen so that we could demonstrate binary equivalence in executables between the CMake build and the Makefile build.
- Use [`find_package(MPI REQUIRED)`](https://cmake.org/cmake/help/latest/module/FindMPI.html) instead of specifying `mpif90` when invoking `cmake` for MPI case. This is done so that
    1. We can create separate targets for the serial and MPI executable and only link against the MPI libraries when needed without relying on the MPI compiler wrapper. This lets us compile all executables (serial and parallel) with a single invocation of CMake.
    2. We can use object libraries to compile the object files common to both serial and MPI builds. This saves compiling these object files for each executable (serial and MPI), reducing compilation time.
- Portability fixes and improvements to build.bash.
- Add `--compiler` flag to build.bash and GNU compiler support.
- Add Matthias's configuration for example.

Regression tests using [benchcab](https://github.com/CABLE-LSM/benchcab)* (bitwise comparison of model output via nccmp) show that model output is bitwise identical between the current branch and the main branch for serial and MPI model runs.

*executables were built manually as benchcab does not yet support the recent build system changes (see related issue: https://github.com/CABLE-LSM/benchcab/issues/258).

Fixes #215

----
📚 Documentation preview 📚: https://cable--216.org.readthedocs.build/en/216/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,17 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
     )
 endif()
 
+set(CABLE_GNU_Fortran_FLAGS -cpp -ffree-form -ffixed-line-length-132)
+set(CABLE_GNU_Fortran_FLAGS_DEBUG -O -g -pedantic-errors -Wall -W -Wno-maybe-uninitialized -fbacktrace -ffpe-trap=zero,overflow,underflow -finit-real=nan)
+set(CABLE_GNU_Fortran_FLAGS_RELEASE -O3 -Wno-aggressive-loop-optimizations)
+if(CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+    add_compile_options(
+        ${CABLE_GNU_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_GNU_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_GNU_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
+
 add_library(
     cable_common_objlib
     OBJECT
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,31 +9,26 @@ project(
 option(CABLE_MPI "Build the MPI executable" OFF)
 
 # third party libs
+if(CABLE_MPI)
+    find_package(MPI REQUIRED COMPONENTS Fortran)
+endif()
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(NETCDF REQUIRED IMPORTED_TARGET "netcdf-fortran")
 
-set(CABLE_INTEL_Fortran_FLAGS -fp-model precise)
-set(CABLE_INTEL_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0)
-set(CABLE_INTEL_Fortran_FLAGS_RELEASE -O2)
+set(CABLE_Intel_Fortran_FLAGS -fp-model precise)
+set(CABLE_Intel_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0)
+set(CABLE_Intel_Fortran_FLAGS_RELEASE -O2)
+if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
+    add_compile_options(
+        ${CABLE_Intel_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_Intel_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_Intel_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
 
-# CMake $<...> syntax is explained here:
-# https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#introduction
-set(CABLE_EXE_NAME "$<IF:$<BOOL:${CABLE_MPI}>,cable-mpi,cable>")
-
-set(
-    CABLE_SRCS_SERIAL
-    src/offline/cable_driver.F90
-)
-set(
-    CABLE_SRCS_MPI
-    src/offline/cable_mpidrv.F90
-    src/offline/cable_mpicommon.F90
-    src/offline/cable_mpimaster.F90
-    src/offline/cable_mpiworker.F90
-    src/science/pop/pop_mpi.F90
-)
-set(
-    CABLE_SRCS_COMMON
+add_library(
+    cable_common_objlib
+    OBJECT
     src/science/casa-cnp/bgcdriver.F90
     src/science/casa-cnp/biogeochem_casa.F90
     src/offline/cable_abort.F90
@@ -143,29 +138,26 @@ set(
     src/science/roughness/roughnessHGT_effLAI_cbl.F90
     src/offline/spincasacnp.F90
 )
-set(
-    CABLE_SRCS
-    "$<IF:$<BOOL:${CABLE_MPI}>,${CABLE_SRCS_MPI},${CABLE_SRCS_SERIAL}>"
-    ${CABLE_SRCS_COMMON}
-)
+target_link_libraries(cable_common_objlib PRIVATE PkgConfig::NETCDF)
 
-add_executable(cable ${CABLE_SRCS})
-
-set_target_properties(cable PROPERTIES OUTPUT_NAME ${CABLE_EXE_NAME})
-
-if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
-    target_compile_options(
-        cable
-        PRIVATE
-        ${CABLE_INTEL_Fortran_FLAGS}
-        "$<$<CONFIG:Release>:${CABLE_INTEL_Fortran_FLAGS_RELEASE}>"
-        "$<$<CONFIG:Debug>:${CABLE_INTEL_Fortran_FLAGS_DEBUG}>"
-    )
-endif()
-
-target_link_libraries(
+add_executable(
     cable
-    PkgConfig::NETCDF
+    src/offline/cable_driver.F90
+    "$<TARGET_OBJECTS:cable_common_objlib>"
 )
-
+target_link_libraries(cable PRIVATE PkgConfig::NETCDF)
 install(TARGETS cable RUNTIME)
+
+if(CABLE_MPI)
+    add_executable(
+        cable-mpi
+        src/offline/cable_mpidrv.F90
+        src/offline/cable_mpicommon.F90
+        src/offline/cable_mpimaster.F90
+        src/offline/cable_mpiworker.F90
+        src/science/pop/pop_mpi.F90
+        "$<TARGET_OBJECTS:cable_common_objlib>"
+    )
+    target_link_libraries(cable-mpi PRIVATE PkgConfig::NETCDF MPI::MPI_Fortran)
+    install(TARGETS cable-mpi RUNTIME)
+endif()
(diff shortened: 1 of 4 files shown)


mcuntz:
Dear @SeanBryan51 ,

very good that the build system got an update.

Being a community model, it would be good if it would work on more than one computing system and with more than one compiler
I looked into build.bash and CMakeLIsts.txt and it is not obvious to me to add a new compiler system.
I am running CABLE on three different clusters and on my local Mac. Appart from my strong conviction that one should be compiling and running its computer code with different compilers, new Macs (with ARM processors) simply do not have access to Intel compilers anymore, but they can use gfortran, for example.

I would be good if you included the possibility to compile CABLE with gfortran on Gadi. That would help to port the script to other systems.

I also think that it would be good to use less environment variables. The build script should help me and not make me have to remember various environment variables.
So it would be good to have a parallel compilation switch, for example, with the build.script, starting it like `./build.bash -j 8`.
And the debug mode could be switched on with `./build.bash -j 8 -d`, for example.

Lastly, I do not think that it is a good idea to remove the --mpi switch. Sometimes an MPI library gets loaded due to other modules, such as a Python environment that uses mpi4py. But I still want to compile the serial version.

Kind regards,
Matthias

SeanBryan51:
Hi Matthias, thank you for the feedback.

- We can definitely add a case using gfortran. I can look into adding a flag to specify the compiler (e.g. `./build.bash --compiler=gnu`) when invoking build.bash.
- I agree that an environment variable for the number of parallel jobs is a bit inconvenient. We can wrap this around a flag.
- Currently build.bash supports CMake flags so we don't need to define flags which have a CMake analogue. I have included some useful flags / CMake specifics in the help of build.bash.
- I agree with your comment around MPI. CMake supports a `--target` option to select a target to build, this can be used to build either the MPI target or serial target. I can look into adding support for this to build.bash.

mcuntz (changes requested):
Dear @SeanBryan51 ,

I hence tried to port the new build script to my Mac mini with ARM processor, which has the short hostname _mcmini_. Attached is the script that worked in a basic version. I don't know how to do the fancy editing directly so I attach the file to this comment. (I had to add .txt to be able to attach it)
[build.bash.txt](https://github.com/CABLE-LSM/CABLE/files/14485741/build.bash.txt)

Here are several remarks:
1. Could you use the shebang line `#!/usr/bin/env bash` rather than `#!/bin/bash`, please. My bash is at `/opt/homebrew/bin/bash`, for example.
This is the preferred way for all scripting languages, also for python, perl, etc. With these languages, one immediately understands the reason.
2. `prepend_path` is part of the module's perl script. No module, no prepend_path. I put a version `iprepend_path` into the build script, which I use in my `.bash_profile`, and use this with the non-module Mac.
3. I added the config for _mcmini_. I do not know if I need the two PKG_CONFIG_PATH but it is to show you that netcdf-c and netcdf-fortran are in very different places. One of the cluster I am using has a weird, non-homogenous architecture and there a static build is preferred. There I might need the second addition to PKG_CONFIG_PATH.
4. You can see that I had to add the Fortran compiler for CMake because it found the NAG compiler first.
5. With all this, it started compiling but failed with _landuse3.f90_. This is because mland is a module variable and defined as a local variable in the subroutine `landuse_allocate_mland`. So I renamed mland into mmland in this routine (lines 235-487) and then it compiled.
6. I have not put in compiler flags for gfortran and neither had you. Did you try the setup with the GNU compiler?
7. Have you looked into build.ksh of the branch CABLE-POP_TRENDY? We use more compiler flags also on GADI, for example, -O3 instead of -O2, or the optimisation for the CPUs. And we use much, much more flags for the debug version for it to be useful.
8. I find it a bit _ugly_ that the script searches for MPI even if one does not want to do an MPI run. It tells me three lines about "could not find".
9. And I think I had told before that I think the logic is flawed to compile the MPI version if the MPI compiler is found. This can be to several reasons. Just as on my system where it finds the NAG compiler and tries with this.

Kind regards,
Matthias

SeanBryan51:
@mcuntz thanks for testing out the build system!

> Could you use the shebang line #!/usr/bin/env bash rather than #!/bin/bash, please.

Done.

> prepend_path is part of the module's perl script. No module, no prepend_path. I put a version iprepend_path into the build script, which I use in my .bash_profile, and use this with the non-module Mac.

I would like to avoid defining helper functions in build.bash to keep it minimal. If possible, please source the relevant scripts in your `mcmini` section.

> I added the config for mcmini. I do not know if I need the two PKG_CONFIG_PATH but it is to show you that netcdf-c and netcdf-fortran are in very different places. One of the cluster I am using has a weird, non-homogenous architecture and there a static build is preferred. There I might need the second addition to PKG_CONFIG_PATH.

I recommend trying the build without modifying `PKG_CONFIG_PATH` to see if it is necessary. You can also try `pkg-config --path netcdf-fortran` to see if the correct netcdf-fortran path is being used.

Static linking is possible but it will require modifying CMakeLists.txt. Static linking via the [`FindPkgConfig`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html) module also isn't that well supported (see https://gitlab.kitware.com/cmake/cmake/-/issues/21714). It does look like there are work-arounds by using the `NETCDF_STATIC_*` variables and [`target_link_options`](https://cmake.org/cmake/help/latest/command/target_link_options.html). I'm happy to help with this.

> I have not put in compiler flags for gfortran and neither had you. Did you try the setup with the GNU compiler?

Compiler flags for the release and debug configurations are defined in CMakeLists.txt. Currently this has only been done for the Intel compiler:
https://github.com/CABLE-LSM/CABLE/blob/9cc7be60d7004a45a1c87880f808144df9f58254/CMakeLists.txt#L14-L24
I would be happy to add the flags used in your TRENDY branch for the GNU case.

> I find it a bit ugly that the script searches for MPI even if one does not want to do an MPI run. It tells me three lines about "could not find". And I think I had told before that I think the logic is flawed to compile the MPI version if the MPI compiler is found. This can be to several reasons. Just as on my system where it finds the NAG compiler and tries with this.

Currently the default is to build both serial and MPI applications simultaneously. You can choose which application to build by specifying the target to build.bash.

I see now that an explicit switch for MPI would be useful. Currently the default is to build both serial and MPI applications simultaneously. I will change the default to build only the serial application and add back `--mpi` to enable the MPI compilation. I previously had a `CABLE_MPI` flag in CMakeLists.txt, we can use this to guard the build from MPI specifics.

SeanBryan51:
> Have you looked into build.ksh of the branch CABLE-POP_TRENDY? We use more compiler flags also on GADI, for example, -O3 instead of -O2, or the optimisation for the CPUs. And we use much, much more flags for the debug version for it to be useful.

@mcuntz I see [here](https://github.com/CABLE-LSM/CABLE/blob/6680580a0df07ca7040b4ae059fa458ffe8d7dee/offline/build.ksh#L31) you use different flags for the release build as well as using intel-compiler-llvm instead of the regular intel-compiler. I think it is best we discuss these changes in a separate issue as optimisation improvements are outside the scope of this pull request. This pull request is focussed on putting the foundations in place for a build system using CMake.

ccarouge (approved):
I have one comment on the documentation but otherwise I'm fine with the current version.

SeanBryan51:
Hi @mcuntz, have your requested changes been addressed? This pull request requires your approval to be merged.

CMakeLists.txt
```diff
@@ -9,163 +9,171 @@ project(
 option(CABLE_MPI "Build the MPI executable" OFF)
 
 # third party libs
+if (CABLE_MPI)
```
harshula, line 12: You use `if(` in the rest of the file. I think that's the CMake convention.

SeanBryan51:
@mcuntz thank you for reviewing. I'm happy to add the configuration you attached as an example.

> You can also see that I need to pass the argument -DCMAKE_Fortran_COMPILER=gfortran to cmake otherwise it would pick up the NAG compiler.

CMake also recognises ["standard" environment variables](https://cmake.org/cmake/help/latest/manual/cmake-env-variables.7.html#environment-variables-for-languages) like `FC` for the fortran compiler - this is probably how CMake finds the right compiler on Gadi as the `FC` variable is set when module loading the Intel or GNU compilers.

> One little demand: could you make the compiler names either lowercase of case-insensitive, please. I can never remember Intel/intel/ifort or GNU/gnu/gcc/gfortran.

Done.

> I was just wondering if you actually compiled the code with the GNU compiler? Because I had lots of (small) errors. The one I told you before (in landuse3.F90) in release mode but plenty more in debug mode.

Yep I've tested out the GNU compiler on my local machine and had to work through the bug fixes you encountered to get it to work. As you guessed I would prefer this to be done in a separate pull request as it is a bit out of scope.

CMakeLists.txt
```diff
@@ -9,163 +9,171 @@ project(
...
+set(CABLE_Intel_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0)
+set(CABLE_Intel_Fortran_FLAGS_RELEASE -O2)
+if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
+    set(
+        CABLE_Fortran_FLAGS
+        ${CABLE_Intel_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_Intel_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_Intel_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
```
harshula, line 28: Shouldn't CMake do some of this stuff automatically?

CMakeLists.txt
```diff
@@ -9,163 +9,171 @@ project(
...
+set(CABLE_GNU_Fortran_FLAGS_DEBUG -O -g -pedantic-errors -Wall -W -Wno-maybe-uninitialized -fbacktrace -ffpe-trap=zero,overflow,underflow -finit-real=nan)
+set(CABLE_GNU_Fortran_FLAGS_RELEASE -O3 -Wno-aggressive-loop-optimizations)
+if(CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+    set(
+        CABLE_Fortran_FLAGS
+        ${CABLE_GNU_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_GNU_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_GNU_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
```
harshula, line 40: Shouldn't CMake do some of this stuff automatically?

harshula (approved):

PR #218, opened 2024-03-05
Remove vegin%froot1 to vegin%froot6 from the code

rkutteh:
## Description

vegin%froot1 to vegin%froot6 serve no purpose in the code (dead code). I have removed them from /offline/cable_pft_params.F90 and from /offline/pft_params.nml

There is a separate clean-up related to vegin%froot  (a different 2-d array) that I will address in a separate pull request.

I have not edited the corresponding documentation yet. I will wait first to see the outcome of this request.

----
📚 Documentation preview 📚: https://cable--218.org.readthedocs.build/en/218/


JhanSrbinovsky:
froot is the "root fraction" and is used throughout  the code as veg%froot. It looks like offline is currently ONLY using a default value it's getting (possibly) through grid info. Online, there is no grid info. We certainly set this from vegin%froot. This should probably be setup again for offline. vegin is passed to READ the namelist pft_params.nml. froot* are  in there . As it stands the pft_params.nml is identical to the one in ACCESS, which is crucial to hopefully soon run  expensive spin up for online, in offline setting. 

it is probably better to include froot in  cable_parametrs.F90:
` SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zee )`, IDK why it seems deliberately absent (maybe @har917  remembers)

rkutteh:
@JhanSrbinovsky I think we are talking about different arrays here. froot1 to froot6 are 1-d arrays that do nothing in the code (offline or online). It appears that they were replaced at some point with froot which is a 2-d array, which absorbed the 1 to 6 layers in the 1-d arrays into a second index. I was planning on dealing with froot (the 2-d array) in a separate pull request, but since it came up it is worth mentioning that in cable_parameters:

    ! MMY @Oct2022 note replace the look-up table froot with Jackson eq
    ! calculate vegin%froot from using rootbeta and soil depth
    ! (Jackson et al. 1996, Oceologica, 108:389-411)

In any case, I just wanted feedback regarding the 1-d arrays froot1 to froot6 for now. Thanks.

JhanSrbinovsky:
yep thats right. there are 6 soil layers. 

```
veg%froot(i,1) = vegin%froot1(j)
veg%froot(i,2) = vegin%froot2(j)
veg%froot(i,3) = vegin%froot3(j)
veg%froot(i,4) = vegin%froot4(j)
veg%froot(i,5) = vegin%froot5(j)
veg%froot(i,6) = vegin%froot6(j)
```

where "i" is an element  of "mp" and "j"is the PFT. We have to map the PFT onto "mp" - see that subr referenced above.  the pft parameters are ALL 1-D.

har917:
@JhanSrbinovsky @rkutteh We have to be a bit careful here - this `vegin%froot` may still be used in ACCESS-ESM1.5 - so there is a potential need to test any removal of code in the ESM.  

It is/was all a bit confusing to me as to the true state of the code - 
- there's this capability to specify fraction of roots per soil layer (which, I agree, I think is obsolete in offline and CM2)
- the other capability that specifies a rooting distribution profile according to Jackson (1996) with PFT parameters which are given via the veg_params file (`%rootbeta` perhaps)
- various hardwiring in various code bases (e.g. ACCESS1.3 hardwired the Jackson parameter)

We got feedback after ACCESS1.3 that we had to look at this - so CM2 uses the Jackson formulae with PFT varying parameters. I implemented that revision so that CABLE-in-CM2 uses the Jackson parameterisation (by default) and there's a namelist option to revert back to ACCESS1.3 version if  `l_access13_roots=.TRUE.`

ccarouge:
@har917 @rkutteh I had a look in the offline version as well as in JULES and this is what I understand.

## CABLE offline
This change effectively deletes unused variables and has no impact on the results.

## Coupled
ACCESS-ESM1.5, CM2, ACCESS-AM3 do not use the namelist but use hard-wired values per PFT. These do not include the variables `froot1` to `froot6` so the propose change would have no effect on these. @JhanSrbinovsky Can you confirm the `vegin` values are hard-wired for these model versions?

## JULES
JULES is using the `pft_params.nml` file and as such, it expects the namelist to have the `froot1` to `froot6` variables in. So a corresponding change is needed for JULES offline to stay in sync. The question is how to implement that change. 

## Conclusion
I think before we go ahead with this, we need to finalise how to manage the synchronisation with JULES. And this is probably urgent as other incoming changes will have the same requirement.

JhanSrbinovsky:
> ## Coupled
> ACCESS-ESM1.5, CM2, ACCESS-AM3 do not use the namelist but use hard-wired values per PFT. These do not include the variables `froot1` to `froot6` so the propose change would have no effect on these. @JhanSrbinovsky Can you confirm the `vegin` values are hard-wired for these model versions?

For models:
ACCESS-ESM1.5-CABLE3          - hard-wired 
CM2                                                  - hard-wired  
ACCESS-AM3                                - namelist - the same pft_params.nml as JULES
offline                                               - namelist - the same pft_params.nml as JULES, JAC,                                                              AM3, (CM3, ESM3 dont exist yet but will be built on AM3)
ACCESS-ESM1.5-CABLE2          - dont know (whilst this was a CMIP6 model, I did this part back in ACCESS1.3 - CMIP5 - I have to go digging)

My intention has always been to have a single pft_params.nml across applications. Going forward this is not CM2, etc. My point was that veg%froot is an integral part of CABLE. It seems unlikely that we go back to using initialisation from froot1:6, but as you suggested it is going to require a corresponding JULES Ticket to at least update the namelist in there, unless we are OK with different pft_params files, and for the sake of removing some redundant code. Although, we should be doing all of these things. I guess we should keep track of them (even branches where the work has been done, so we can attach it with a larger submission, or just push through this small ticket I guess)

har917:
okay - we have to be a bit careful here - there's variation by hard-wiring/name lists/veg_params.txt and variation by the general approach (directly specifying `%froot1-6` vs specifying `%rootbeta` and letting the Jackson (1996) code work `%froot1-6` out)

I think we have that
- ACCESS1.3 and ACCESS-ESM1.5 is hard-wired (PFT independent) `%froot1-6`
- CM2 is hard-wired (PFT varying) `%frootbeta`  (but with a name list switch to revert back to ACCESS1.3
- CABLE3-offline and AM3 uses a name list to provide values of `%frootbeta` by PFT
- offline TRENDY branch is probably using veg_params.txt to provide `%frootbeta'

I think/hope that the actual/default values in offline TRENDY, CABLE3-offline, CM2 and AM3 are all the same.

An earlier comment of mine (to Calre and Ramzi that seems to have vanished - maybe I didn't hit comment?) was that it could be useful to retain the means of overriding the Jackson (1996) formulation.  this function given is not really designed to work with a really thin surface soil layer.  It puts a lot of roots in the surface (2.2cm) which is likely impacting soil hydrology, infiltration, run off, ET etc - that's science for later though.   As a result - my inclination would be to work the other way and reactivate a means to read in `%froot1-6` or `%rootbeta` and a switch to go between the two.

For now that may mean commenting code not removing? 

Note that you cannot give a value to `%frootbeta` which results in the same root distribution as ACCESS1.3

ccarouge:
> For now that may mean commenting code

No. We either do nothing or do the work (whether we remove or add new capability). Going in and commenting code while leaving it in is just bad practice in general and unnecessary. The code that is in for the moment does nothing, it is confusing for developers but not harmful to the code.

From your comment @har917, I guess we should close the issue and PR and open a new issue to implement a switch between `rootbeta` and `froot1-6` which will be worked on when someone feels like doing it.

ccarouge:
Issue to tackle this properly is finally opened: #334 . Closing this PR.

PR #222, opened 2024-03-13
Add python version of split_landmask.R script

Whyborn:
# CABLE

## Description

Added a python version of the split_landmask.R script built by @ccarouge (with minor modifications to open directories before we write to them). Added lines to the run_TRENDY.sh script demonstrating their usage (lines currently commented out, so original R behaviour is preserved for now). Involves addition of -e <extent> to the split_landmask execution line. Can also utilise the python module contained in the hh5 project, if user has access to said project.

## Type of change

- [ ] Bug fix
- [x] Minor addition

## Checklist

### DIFF
--- a/scripts/TRENDY_runs/v12/run_TRENDY.sh
+++ b/scripts/TRENDY_runs/v12/run_TRENDY.sh
@@ -24,7 +24,7 @@ module load geos/3.8.0
 module load intel-compiler/2021.8.0
 export R_LIBS=/g/data/x45/R/libs
 export PYTHONPATH=/g/data/x45/python3.10.4/lib/python3.10/site-packages
-
+# module load conda_concept/analysis3       # Can load python libraries from this module, supplied by project hh5
 
 #-------------------------------------------------------
 # Settings
@@ -56,6 +56,7 @@ cablecode="/home/599/jk8585/CABLE_code/SHARE/CABLE-POP_TRENDY"
 rundir="/home/599/jk8585/CABLE_run/TRENDY_v12"
 # Scripts
 landmask_script="${rundir}/scripts/split_landmask.R"
+#landmask_script="${rundir}/scripts/split_landmask.py"
 run_script="${rundir}/scripts/run_cable.sh"
 merge_script="${rundir}/scripts/merge_outputs.sh"
 cleanup_script="${rundir}/scripts/cleanup.sh"
@@ -81,6 +82,7 @@ GlobalLandMaskFile="${aux}/landmasks/glob_ipsl_1x1.nc"
 # -----------------------------------------------------------------------
 if [[ ${create_landmasks} -eq 1 ]] ; then
    $landmask_script $GlobalLandMaskFile $nruns $outpath $extent
+   #$landmask_script $GlobalLandMaskFile $nruns $outpath -e $extent     # Python uses arg -e for extent
 fi
 
 # -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


scripts/TRENDY_runs/v12/run_TRENDY.sh
```diff
@@ -24,7 +24,7 @@ module load geos/3.8.0
 module load intel-compiler/2021.8.0
 export R_LIBS=/g/data/x45/R/libs
 export PYTHONPATH=/g/data/x45/python3.10.4/lib/python3.10/site-packages
-
+# module load conda_concept/analysis3       # Can load python libraries from this module, supplied by project hh5
```
mcuntz, line 27: Can you add in the comment that access to hh5 can be asked on [https://my.nci.org.au/](https://my.nci.org.au/)?

scripts/TRENDY_runs/v12/split_landmask.py
```diff
@@ -0,0 +1,109 @@
...
+
+def generate_parser():
+    """Generate the argument parser"""
+    args = argparse.ArgumentParser(
+        description="Generate land masks for CABLE to split a global run in a series of serial runs."
+    )
+    args.add_argument("landmask_file", help="Initial global landmask")
+    args.add_argument("nmasks", help="Number of runs for CABLE-POP", type=int)
+    args.add_argument("outpath", help="Root path for output files")
+    args.add_argument(
```
mcuntz, line 22: There seems to be default argument so that 'global' is taken if -e not given.

scripts/TRENDY_runs/v12/split_landmask.py
```diff
@@ -0,0 +1,109 @@
...
+        landmask_per_run = landmask_in.where(run_mask == i + 1, 0)
+        os.makedirs(f"{args.outpath}/run{i+1}/landmask", exist_ok = True)
+        landmask_per_run.to_netcdf(
+            f"{args.outpath}/run{i+1}/landmask/landmask{i+1}.nc",
+            encoding={landmask_per_run.name: {"dtype": np.byte, "_FillValue": 0}},
+        )
+
+
+def create_run_mask(
+    landmask_in: xr.DataArray,
```
mcuntz, line 71: Do you want/need type annotation here? I heard mixed things about it.

mcuntz (changes requested):
Basically mostly setting the extent to global as default.

scripts/TRENDY_runs/v12/split_landmask.py
Whyborn: As far as I'm aware, they don't actually do anything in standard python code (https://docs.python.org/3/library/typing.html). It's at least some form of documentation, and if it's already there, I don't think we need to remove it.

juergenknauer:
I cannot comment much on the code as a I am not a python person, but I would definitely do is to run both the python and R script and create an arbitrary number of land masks to see if they are the same.

juergenknauer (commented):
I cannot comment much on the code as a I am not a python person, but I would definitely do is to run both the python and R script and create an arbitrary number of land masks to see if they are the same.

Whyborn:
Yep, Claire and I are doing that now.

ccarouge:
@juergenknauer I have run both the R and Python scripts and they give the same masks. 
With Lachlan we are now checking that we can reproduce the TRENDY v12 runs with the original input files (including the R script) and the files copied to one common place (rp23) and this Python script.

juergenknauer:
> @juergenknauer I have run both the R and Python scripts and they give the same masks. With Lachlan we are now checking that we can reproduce the TRENDY v12 runs with the original input files (including the R script) and the files copied to one common place (rp23) and this Python script.

That's great news. Let me know if you encounter issues with the runs.

mcuntz (approved):

PR #224, opened 2024-03-17
Introduce SUBROUTINE GWspatialParameters

rkutteh:
# CABLE ground water hydrology work of Mengyuan Mu

This pull request is part of integrating the ground water hydrology work of Mengyuan Mu into CABLE, which will take place in a sequence of pull requests. The actual integration project itself was completed back in September 2023. The current work involves just a gradual merge into the main branch. 

## Description

1. A new SUBROUTINE GWspatialParameters is introduced in **src/offline/cable_parameters.F90**
2. This subroutine will eventually be called only once, from **src/offline/cable_input.F90**. I will add this call at the appropriate time, but for now the subroutine is inactive
3. Dependencies are introduced in **src/offline/cable_parameters.F90** and **src/offline/cable_define_types.F90** in order for the code to compile successfully (confirmed)
4. This pull request involves only code additions, no deletions or modifications of existing code were carried out (some minor typos I came across in existing comments were fixed)
5. *PLEASE* do not delete any of my in-code comments (tagged rk4417) for now. Once the integration of the ground water hydrology work is complete and its functionality is confirmed to reproduce the results obtained already outside of the repo, I will clean-up all of my comments in a single pull request. Deleting any of my comments now will just make it much more difficult for me to track down and fix any issues that might arise during this process.

----
📚 Documentation preview 📚: https://cable--224.org.readthedocs.build/en/224/

### DIFF
--- a/src/offline/cable_parameters.F90
+++ b/src/offline/cable_parameters.F90
@@ -125,6 +125,15 @@ MODULE cable_param_module
   ! vars intro for Ticket #27
   INTEGER, DIMENSION(:, :),     ALLOCATABLE :: inSoilColor
 
+  INTERFACE get_gw_data                          ! rk4417 - phase2
+     MODULE PROCEDURE get_gw_2d_var_constdef  
+     MODULE PROCEDURE get_gw_3d_var_constdef
+     MODULE PROCEDURE get_gw_4d_var_constdef
+     MODULE PROCEDURE get_gw_2d_var
+     MODULE PROCEDURE get_gw_3d_var
+     MODULE PROCEDURE get_gw_4d_var
+  END INTERFACE
+  
 CONTAINS
 
   SUBROUTINE get_default_params(logn, vegparmnew, LUC_EXPT)
@@ -980,10 +989,6 @@ CONTAINS
 
   !=============================================================================
 
-
-
-
-
   SUBROUTINE countPatch(nlon, nlat, npatch)
     ! count the total number of active patches and
     ! fill the index variable 'landpt'
@@ -1010,6 +1015,12 @@ CONTAINS
     REAL :: lon2, distance, newLength
     INTEGER :: ii, jj, kk, tt, ncount
 
+! mpatch setting below introduced by rk4417 - phase2    
+    !set mpatch, the publically available number of patches
+    mpatch = npatch   ! MMY@13April it makes sense. mpatch is set in cable_define_types_mod    
+                      !     as a public variable and is used in subroutine GWspatialParameters 
+                      !     and in get_gw_data
+    
     ! range of longitudes from input file (inLon) should be -180 to 180,
     ! and longitude(:) has already been converted to -180 to 180 for CABLE.
     landpt(:)%ilon = -999
@@ -1092,6 +1103,11 @@ CONTAINS
 
     ! Set the maximum number of active patches to that read from met file:
     max_vegpatches = MAXVAL(landpt(:)%nap)
+
+! mpatch setting below introduced by rk4417 - phase2
+    mpatch         = max_vegpatches   ! MMY@13April keep this line, it is an update of mpatch to consider
+                                      !        force_npatches_as's change on mpatch
+    
     !CLN    IF (max_vegpatches /= nmetpatches) THEN
     IF (max_vegpatches /= nmetpatches .AND. npatch == 1) THEN
        PRINT *, 'Error! Met file claiming to have more active patches than'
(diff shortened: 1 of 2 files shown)


ccarouge:
@rkutteh I know you were hoping for a quick pull request. But I don't think I can give you that. From what I've seen during the review the GWspatialParameters add to the spaghetti mess of cable_parameters.F90 instead of helping untangling things. So I'm not particularly inclined to accept as is. I'll have a deeper look into it and see if we can adapt things quickly to get the code to a better standard.

rkutteh:
@ccarouge As you might guess from my earlier comments, I prefer that all clean-up of code occurs **after** all the GW work is in and validated against the copy I have outside the repo. If you change things around while I am moving code in and the final code does not agree with my results, it will be nearly impossible for me to track down the source(s) of the errors. As it stands, we know that the code that I am introducing works with GW on, so my own strategy would be to mirror it in the repo, check that all is good as expected, and only THEN clean-up and check, clean-up and check,...Keep in mind that GWspatialParameters is presently inactive.

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+         integer, intent(in)               :: ncfile_id
+         real(r_2), dimension(mp) :: data_vec
+
+         real, dimension(nlon,nlat) :: GW2d_data
+         real, dimension(nlon,nlat) :: default_data
+         integer :: i,j,k,n
+         integer :: varid
+         integer :: varinq_status,varget_status
+
+         print *, "MMY in get_gw_2d_var_constdef" ! MMY add to test whether get_gw_2d_var_constdef is necessary
```
ccarouge, line 3084:
```suggestion
```

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+         default_data(:,:) = default_const
+
+         if (.not.cable_user%gw_model) then
+
+            GW2d_data(:,:) = default_data(:,:)
+
+         else
```
ccarouge, line 3091: I would like to remove this case. This should be done outside these routines. Keep these generic.

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+         real, dimension(nlon,nlat) :: GW2d_data
+         integer :: i,j,k,n
+         integer :: varid
+         integer :: varinq_status,varget_status
+
+         if (.not.cable_user%gw_model) then
+
+            GW2d_data(:,:) = default_data(:,:)
+
+         else
```
ccarouge, line 3139: I would like to remove this case. This should be done outside these routines. Keep these generic.

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+         real, dimension(nlon,nlat) :: default_data
+
+          default_data(:,:) = default_const
+
+         if (.not. cable_user%gw_model) then
+            do k=1,ms
+               GW3d_data(:,:,k) = default_data(:,:)
+            end do
+
+         else
```
ccarouge, line 3191: I would like to remove this case. This should be done outside these routines. Keep these generic.

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+         integer :: i,j,k,n
+         integer :: varid
+         integer :: varinq_status,varget_status
+
+         if (.not. cable_user%gw_model) then
+            do k=1,ms
+               GW3d_data(:,:,k) = default_data(:,:)
+            end do
+
+         else
```
ccarouge, line 3245: I would like to remove this case. This should be done outside these routines. Keep these generic.

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+
+         default_data(:,:) = default_const
+
+         if (.not.cable_user%gw_model) then
+            do k=1,ms
+               do n=1,npatch
+                  GW4d_data(:,:,k,n) = default_data(:,:)
+               end do
+            end do
+         else
```
ccarouge, line 3304: I would like to remove this case. This should be done outside these routines. Keep these generic.

src/offline/cable_parameters.F90
```diff
@@ -2740,11 +2756,660 @@ SUBROUTINE init_veg_from_vegin(ifmp,fmp, veg, soil_zse )
...
+         real, dimension(nlon,nlat,ms)        :: GW3d_data
+
+
+         if (.not.cable_user%gw_model) then
+            do k=1,ms
+               do n=1,npatch
+                  GW4d_data(:,:,k,n) = default_data(:,:)
+               end do
+            end do
+         else
```
ccarouge, line 3370: I would like to remove this case. This should be done outside these routines. Keep these generic.

ccarouge (approved):
I have written down what I would change (at least what I could see here). 
But I have seen that `cable_parameters.F90` already has code to read in the GW parameters and assign the values to the CABLE variables (in spatialSoil() and write_default_params()), so we need to ensure there is only one way to do that in the final code.

This clean up will be done separately once the GW code is committed in to ensure testing is possible.

rkutteh:
@ccarouge thanks for all the comments. Some I suspect will be resolved "automatically", once I am finished pushing all the remaining changes to src/offline/cable_parameters.F90 . The remaining I agree we can address up after testing.

Issue #234, opened 2024-04-12
GW merge: evaluation of changes outside GW


ccarouge:
The merge of the groundwater branch (GW) has uncovered changes made in that branch that will affect all simulations even with the GW turned off. This issue is to inform users about the evaluation work that is to start. Discussions about each change will happen in individual PRs.

Changes that will be kept will be recorded in their own issues once the decision to keep them is made.

ccarouge:
We will implement forced consistency of ice points in CABLE #280

PR #235, opened 2024-04-12
Ensure "ice" veg type is also permanent ice soil - #234

ccarouge:
# CABLE

## Description

The GW work introduces this change. It ensures that points with "ice" for vegetation type are also classified as "permanent ice" for the soil type.

#234

## Type of change

- [ ] Bug fix
- [X] New 

## Checklist

----
📚 Documentation preview 📚: https://cable--235.org.readthedocs.build/en/235/

### DIFF
--- a/src/offline/cable_parameters.F90
+++ b/src/offline/cable_parameters.F90
@@ -1759,7 +1759,7 @@ CONTAINS
   SUBROUTINE derived_parameters(soil, sum_flux, bal, ssnow, veg, rough)
     ! Gives values to parameters that are derived from other parameters.
     TYPE (soil_snow_type),      INTENT(INOUT)    :: ssnow
-    TYPE (veg_parameter_type),  INTENT(IN)    :: veg
+    TYPE (veg_parameter_type),  INTENT(INOUT)    :: veg
     TYPE (soil_parameter_type), INTENT(INOUT) :: soil
     TYPE (sum_flux_type),       INTENT(INOUT) :: sum_flux
     TYPE (balances_type),       INTENT(INOUT) :: bal
@@ -1774,8 +1774,10 @@ CONTAINS
     REAL(r_2), DIMENSION(mp,ms) :: psi_tmp
     REAL(r_2), DIMENSION(ms) :: soil_depth
 
+    ! ccc. Ensure consitency between permanent ice soil type and ice vegetation 
     ! line below inserted by rk4417 - phase2
     where(veg%iveg .eq. 17) soil%isoilm = 9   ! 
+    where(soil%isoilm .eq. 9) veg%iveg = 17
     
     soil_depth(1) = REAL(soil%zse(1),r_2)
     DO klev=2,ms
--- a/src/offline/cable_parameters.F90
+++ b/src/offline/cable_parameters.F90
@@ -1774,6 +1774,9 @@ CONTAINS
     REAL(r_2), DIMENSION(mp,ms) :: psi_tmp
     REAL(r_2), DIMENSION(ms) :: soil_depth
 
+    ! line below inserted by rk4417 - phase2
+    where(veg%iveg .eq. 17) soil%isoilm = 9   ! 
+    
     soil_depth(1) = REAL(soil%zse(1),r_2)
     DO klev=2,ms
        soil_depth(klev) = soil_depth(klev-1) + REAL(soil%zse(klev),r_2)


ccarouge:
@har917 Mengyuan is introducing this change in her work for the groundwater and I'm not sure if we want it. Typically, it ensures all patches with ice for "vegetation" also have permanent ice for the soil. 

I know we discussed something similar for CM3 but didn't necessarily plan on implementing it. Additionally, she puts it in `derived_parameters()` and I'm wondering if it's too late. Would any of the soil properties (soil%) need to be modified as well? I haven't checked if some initialisation for permanent ice points happen before that point or not, I thought I'd ask your opinion early before investing too much time on it. 

The spatial case in benchcab shows differences in some points in the Arctic so we can use that to document the effect if needed.

ccarouge:
If I look at `cable_soilparm.nml` (which is not used) to give me an idea of typical values for permanent ice, it seems that `css` and `rhosoil` were different from other soil types. Should their values be updated as well if we force a point to be permanent ice? Does it make sense to always use the same value for these 2 parameters for all types of permanently frozen soils?

ccarouge:
@rkutteh you might be interested in this discussion since it pertains to GW work.

JhanSrbinovsky:
> @har917 Mengyuan is introducing this change in her work for the groundwater and I'm not sure if we want it. Typically, it ensures all patches with ice for "vegetation" also have permanent ice for the soil.

I'm just curious -  when would this NOT be the case?

> I know we discussed something similar for CM3 but didn't necessarily plan on implementing it. 

Effectively we HAVE been doing this since CMIP5. The UM doesn't have a SoilType map, it doesn't need one. The Soil parameters are all on spatial grids.  Also note for the next comment/reply that the UM/JULES has no concept of soil density. I have comments on this several times. A lot of this is similar to my discussion with David Hutchinson. In the UM CMIPs we use VegType=ICE to specify SoilType=ICE, and use "our" per SoilType params (same as in tour offline params). Whether this describes the soil under 1km of ICE and how much it would impact surface physics, I won't comment. So - other than SoilType=ICE (Greenland,Antarctica) SoilType=2. i.e. the rest of the globe is covered in clay. This doesn't matter so much because we are actually using UM spatial maps to describe the soil anyway. Except for - der der deeeeerrrr - soil density - because as I said JULES does not have it. Also as I have said before (sorry for repeating) it is only used a few times and always as the same product, which from memory IS available. Further, I can't put my finger on it but I vaguely remember that we had to get a soil type/texture map for CASA, and I can't remember what we did.

har917:
I can't actually remember where `isoil==frozen` gets used (outside of the UM) - so the following could be incorrect.

In principle I can see the need to apply consistency between `veg%iveg` and `soil%isoil` when it comes to frozen land points - however I would have this the other way around, i.e.
- if soil is frozen then ensure veg is frozen
- if veg is frozen and soil is not frozen then change the iveg to bare soil

We do need to be mindful of the order of operations (enforce consistency between iveg and isoil and then fill the other veg% and soil% properties) which could require larger changes.

also the change set uses hardwired indices - can we avoid creating more of these please (if at all possible)

ccarouge:
@har917 

> if soil is frozen then ensure veg is frozen

Possibly but below is the change in soil type from applying Mengyuan's addition compared to without. I would think all the points that were changed should indeed be permanent ice. So doing it this way makes sense to me.

I'll apply the reverse as well (not changing the properties yet) and see where we get changes (if any).

Screenshot 2024-04-18 at 12 44 30 pm

har917:
> I would think all the points that were changed should indeed be permanent ice. So doing it this way makes sense to me.

I would say not - these points are where soil is frozen, not ice.  In effect the only permanent ice grid cells should be those in the centre of Greenland and Antarctica.

... and to address other points in thread - the 'soil' properties for ice points should be that of frozen water (i.e. the fabric of the substrate below any snow cover), whereas the soil properties for the other soils should be that of the soil type.  In the UM context there is an issue because CABLE is written to ingest soil density and soil specific heat capacity separately whereas as JULES/UM only has provision to read in volumetric heat capacity (the product of the two) - but I think we have a handle on this.

... activating CASA in AM3 will likely mean a return to this question as - from memory - additional spatial fields (soil%clay, soil%sand etc.) are used and I'm not sure whether the current build has all of those available [but that's a separate topic]

har917:
> I would say not - these points are where soil is frozen, not ice. In effect the only permanent ice grid cells should be those in the centre of Greenland and Antarctica.

caveat to this on further thought - some of these `veg=frozen` but `soil/=frozen` could be representing glaciers.  In this case there is a variation in the effective soil type within the grid cell and hence isoil should be changed to be consistent with the veg type (not the other way around).

JhanSrbinovsky:
we actually had a case not so long ago (from memory it was only one cell) on the coast of Greenland where the PFT map was non-ICE but all the soil parameters there were effectively rmdi. Resulted in a messy bug to track down

har917:
Hi - so tried to make a comment inline but bit blocked me.

I think that we will have to move this enforcement from `derived_parameters` elsewhere _or_ extend the enforcement to the entire `veg%` and `soil%` TYPE.

At the moment the `veg%` and `soil%` TYPEs are largely populated via `get_default_params` or the restart - which are earlier in the sequence than `derived_parameters`.  Enforcing consistency only in the veg and soil types here will then lead to the parameters being inconsistent with the surface index - e.g. you could end up with an ice point wih non-zero LAI.

ccarouge:
> we will have to move this enforcement from derived_parameters

I know, this was just to see how many points would be impacted by the change.

But before we go into more implementation details I think there are at least 2 questions we haven't answered.

# Do we want to put this in CABLE?

Do we want CABLE to change the input maps of soil and vegetation types and parameters? Or should that be part of preparing the input files? It would be really easy to have a program that reads in a gridinfo file, plot the locations where permanent ice /= ice "vegetation", apply the consistency rules to the types and the parameters and output a consistent gridinfo file.

I am not certain CABLE should be in charge of ensuring the consistency, mainly because it obscures things.

# What are the frozen ice parameters?

That can probably be found on the internet. But I can see that the JAC cable_soilparm.nml file has 

```
css = 2100
rhosoil = 910
```

While the gridinfo file in the benchcab testcase (`/g/data/rp23/data/no_provenance/gridinfo/crujra_accessN96_gridinfo.nc`) has (values taken over Greenland):

```
css = 2100
rhosoil = 1455
```
The internet tells me the namelist is correct and not the gridinfo file. 

So in addition to ensuring consistency in types, should we ensure permanent ice points have the correct parameters as well? If we add permanent ice points to that dataset, they will have a different density from the existing one.

har917:
> What are the frozen ice parameters?

I prefer the namelist values - cf [wikipedia](https://en.wikipedia.org/wiki/Table_of_specific_heat_capacities) and compare th3 volumetric heat capacity with the product of the two namelist values.

> Do we want to put this in CABLE?

Harder to answer as this requires knowing all the use cases and how the veg type and soil type are/will be used.  Ideally it would be up to the user to sort out their inputs to be consistent with each other.  However (I think) we still need a sanity check step somewhere to catch stuff that could (should) cause the model problems (such as a veg land point on ice) - but that may be better placed as separate executable rather than inside the codebase. 

JULES has a sanity check step on the science options but not (from memory) on parameter values.

ccarouge:
> we still need a sanity check

OK. I'll figure something more comprehensive than what Mengyuan has put then.

har917:
re sanity check - on further thinking - I think it should be possible to do this outside of the model but it would have to have offline/ACCESS/CCAM? variants.  This is because the UM-JULES soil and vegetation cover ancillaries are read in as separate files whereas gridinfo is one.  

So it may be best if it could be done inside CABLE.

ccarouge:
For JULES, it will have to be done differently in JULES and offline anyway, even if in CABLE, because that's in the part of the code that is not shared between the 2 codes. But I'm happy to put something in CABLE. It will have to be more rudimentary than outside CABLE. Outside CABLE, one could think of giving some flexibility while I think inside CABLE it is better if it is prescriptive (a real sanity check not a "a la carte" option).

ccarouge:
I've opened an issue for that: #280 . I've asked @rkutteh to have a look if he can find the right implementation for it. I'm closing this as it served its purpose to clarify what we wanted to do on that front.

PR #238, opened 2024-04-18
Port branch CABLE-POP_TRENDY to CMake

SeanBryan51:
PR for porting the CABLE-POP_TRENDY branch to CMake.

**Context:**

- We want to consolidate the build systems across CABLE to use CMake to improve ease of testing with tools such as [benchcab][benchcab] and [spack][spack].
- With the current CABLE-4 development around CABLE-POP_TRENDY, it makes sense to port this branch first.

**Checklist:**

- [x] Review compiler flags for default, release and debug configurations for the following compilers in `CABLE-POP_TRENDY` and compare against build configurations used in `main` (see #268):
    1. Intel compiler (classic)
    2. Intel compiler (llvm)
    3. GNU compiler
- [x] Ensure model outputs are bitwise identical between Makefile and CMake builds of CABLE-POP_TRENDY on Gadi.
  - Model outputs were generated using the [TRENDY_V12][TRENDY_V12] configuration for a reduced domain (`extent="64.0,66.0,60.0,62.0"` in `run_TRENDY.sh`). Model outputs were compared using `nccmp -d <file_a> <file_b>`. The model outputs being compared are all `cru_out_cable_1901_2022.nc` and `cru_out_casa_1901_2022.nc` outputs (as these two files are the final products) for each serial CABLE run.
- [x] Ensure compiled object files are identical between Makefile and CMake builds of CABLE-POP_TRENDY on Gadi.

[benchcab]: https://github.com/CABLE-LSM/benchcab
[spack]: https://spack.io/
[TRENDY_V12]: https://github.com/CABLE-LSM/TRENDY_V12

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,129 @@
+# Set minimum CMake version to latest version on Gadi:
+cmake_minimum_required(VERSION 3.24.2)
+
+project(
+    CABLE
+    LANGUAGES Fortran
+)
+
+option(CABLE_MPI "Build the MPI executable" OFF)
+
+if(CABLE_MPI)
+    # The MPI implementation is currently not supported in CABLE-POP_TRENDY:
+    message(FATAL_ERROR "MPI compilation not supported.")
+endif()
+
+# third party libs
+if(CABLE_MPI)
+    find_package(MPI REQUIRED COMPONENTS Fortran)
+endif()
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(NETCDF REQUIRED IMPORTED_TARGET "netcdf-fortran")
+
+set(CABLE_Intel_Fortran_FLAGS -fp-model precise -fpp -nofixed -assume byterecl -D__INTEL__ -D__INTEL_COMPILER__ -D__NETCDF3__)
+set(CABLE_Intel_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0 -debug extended -check all -warn all -fp-stack-check)
+set(CABLE_Intel_Fortran_FLAGS_RELEASE -O3 -ip)
+if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
+    add_compile_options(
+        ${CABLE_Intel_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_Intel_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_Intel_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
+
+set(CABLE_GNU_Fortran_FLAGS -cpp -ffree-form -ffixed-line-length-132 -D__GFORTRAN__ -D__gFortran__ -D__NETCDF3__)
+set(CABLE_GNU_Fortran_FLAGS_DEBUG -O -g -pedantic-errors -Wall -W -Wno-maybe-uninitialized -fbacktrace -ffpe-trap=zero,overflow,underflow -finit-real=nan)
+set(CABLE_GNU_Fortran_FLAGS_RELEASE -O3 -Wno-aggressive-loop-optimizations)
+if(CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+    add_compile_options(
+        ${CABLE_GNU_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_GNU_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_GNU_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
+
+add_library(
+    cable_common_objlib
+    OBJECT
+    core/biogeochem/cable_adjustJVgm.F90
+    core/biogeochem/cable_optimiseJVratio.F90
+    core/biogeochem/cable_phenology.F90
+    core/biogeochem/casa_cable.F90
+    core/biogeochem/casa_cnp.F90
+    core/biogeochem/casa_inout.F90
+    core/biogeochem/CASAONLY_LUC.F90
+    core/biogeochem/casa_variable.F90
+    core/biogeochem/POP.F90
+    core/biogeochem/pop_io.F90
+    core/biogeochem/POPLUC.F90
+    core/biogeochem/spincasacnp.F90
+    core/biogeophys/cable_air.F90
+    core/biogeophys/cable_albedo.F90
+    core/biogeophys/cable_c13o2_def.F90
+    core/biogeophys/cable_c13o2.F90
+    core/biogeophys/cable_canopy.F90
+    core/biogeophys/cable_carbon.F90
+    core/biogeophys/cable_cbm.F90
+    core/biogeophys/cable_climate.F90
+    core/biogeophys/cable_common.F90
+    core/biogeophys/cable_data.F90
+    core/biogeophys/cable_define_types.F90
+    core/biogeophys/cable_diag.F90
+    core/biogeophys/cable_radiation.F90
+    core/biogeophys/cable_roughness.F90
+    core/biogeophys/cable_sli_main.F90
+    core/biogeophys/cable_sli_numbers.F90
+    core/biogeophys/cable_sli_roots.F90
+    core/biogeophys/cable_sli_solve.F90
+    core/biogeophys/cable_sli_utils.F90
+    core/biogeophys/cable_soilsnow.F90
+    core/biogeophys/mo_c13o2_photosynthesis.f90
+    core/biogeophys/mo_constants.f90
+    core/biogeophys/mo_isotope.f90
+    core/biogeophys/mo_isotope_luc_model.f90
+    core/biogeophys/mo_isotope_pool_model.f90
+    core/biogeophys/mo_kind.f90
+    core/biogeophys/mo_utils.f90
+    core/blaze/blaze_driver.F90
+    core/blaze/blaze.F90
+    core/blaze/simfire.F90
+    core/utils/minpack.f90
+    offline/cable_abort.F90
+    offline/cable_bios_met_obs_params.F90
+    offline/cable_checks.F90
+    offline/cable_cru_TRENDY.F90
+    offline/cable_initialise.F90
+    offline/cable_input.F90
+    offline/cable_iovars.F90
+    offline/cable_LUC_EXPT.F90
+    offline/cable_output.F90
+    offline/cable_parameters.F90
+    offline/cable_plume_mip.F90
+    offline/cable_read.F90
+    offline/cable_site.F90
+    offline/cable_weathergenerator.F90
+    offline/cable_write.F90
+)
+target_link_libraries(cable_common_objlib PRIVATE PkgConfig::NETCDF)
+
+add_executable(
+    cable
+    offline/cable_driver.F90
+    "$<TARGET_OBJECTS:cable_common_objlib>"
+)
+target_link_libraries(cable PRIVATE PkgConfig::NETCDF)
+install(TARGETS cable RUNTIME)
+
+if(CABLE_MPI)
+    add_executable(
+        cable-mpi
+        src/offline/cable_mpicommon.F90
+        src/offline/cable_mpidrv.F90
+        src/offline/cable_mpimaster.F90
+        src/offline/cable_mpiworker.F90
+        src/science/pop/pop_mpi.F90
+        "$<TARGET_OBJECTS:cable_common_objlib>"
+    )
+    target_link_libraries(cable-mpi PRIVATE PkgConfig::NETCDF MPI::MPI_Fortran)
+    install(TARGETS cable-mpi RUNTIME)
+endif()
--- a/build.bash
+++ b/build.bash
@@ -1,0 +1,132 @@
+#!/usr/bin/env bash
+
+ncpus_default=4
+
+script_name=$(basename "${0}")
+
+show_help() {
+    cat << EOF
+Usage: ./${script_name} [OPTIONS]
+
+Build script wrapper around CMake. Supplied arguments that do not match the
+options below will be passed to CMake when generating the build system.
+
+Options:
+  -c, --clean   Delete build directory before invoking CMake.
+  -m, --mpi     Compile MPI executable.
+  -C, --compiler <compiler>
+                Specify the compiler to use.
+  -n, --ncpus <ncpus>
+                Specify the number of parallel jobs in the compilation. By
+                default this value is set to ${ncpus_default}.
+  -h, --help    Show this screen.
+
+Enabling debug mode:
+
+  The release build is default. To enable debug mode, specify the CMake option
+  -DCMAKE_BUILD_TYPE=Debug when invoking ${script_name}.
+
+Enabling verbose output from Makefile builds:
+
+  To enable more verbose output from Makefile builds, specify the CMake option
+  -DCMAKE_VERBOSE_MAKEFILE=ON when invoking ${script_name}.
+
+EOF
+}
+
+cmake_args=(-DCMAKE_BUILD_TYPE=Release -DCABLE_MPI=OFF)
+
+# Argument parsing adapted and stolen from http://mywiki.wooledge.org/BashFAQ/035#Complex_nonstandard_add-on_utilities
+while [ ${#} -gt 0 ]; do
+    case ${1} in
+        -c|--clean)
+            rm -r bin build
+            exit
+            ;;
+        -m|--mpi)
+            mpi=1
+            cmake_args+=(-DCABLE_MPI="ON")
+            ;;
+        -C|--compiler)
+            compiler=${2}
+            shift
+            ;;
+        -n|--ncpus)
+            CMAKE_BUILD_PARALLEL_LEVEL=${2}
+            shift
+            ;;
+        -h|--help)
+            show_help
+            exit
+            ;;
+        ?*)
+            cmake_args+=("${1}")
+            ;;
+    esac
+    shift
+done
+
+if hostname -f | grep gadi.nci.org.au > /dev/null; then
+    : "${compiler:=intel}"
+
+    . /etc/bashrc
+    module purge
+    module add cmake/3.24.2
+    module add netcdf/4.9.2
+    case ${compiler} in
+        intel)
+            module add intel-compiler-llvm/2023.0.0
+            compiler_lib_install_dir=Intel
+            [[ -n ${mpi} ]] && module add intel-mpi/2019.5.281
+
+            # This is required to use ifort instead of ifx (default):
+            cmake_args+=(-DCMAKE_Fortran_COMPILER="ifort")
+
+            # This is required to build with runtime dispatch for all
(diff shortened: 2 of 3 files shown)


SeanBryan51:
I've managed to compile the serial executable successfully with the Intel compiler on Gadi (`intel-compiler/2019.5.281`).

To test the build, run:

```
./build.bash -DCMAKE_Fortran_FLAGS="-fpp -O3 -nofixed -assume byterecl -fp-model precise -ip -diag-disable=10382,15009 -march=broadwell -D__INTEL__ -D__INTEL_COMPILER__ -D__NETCDF3__"
```

Note: compiler flags included above were taken from [`offline/build.ksh` L49-L54](https://github.com/CABLE-LSM/CABLE/blob/d534a3f51129c5b2b0eaa1279eb84eb02d4225d9/offline/build.ksh#L49-L54) (with the exception of `-axSKYLAKE-AVX512,CASCADELAKE,SAPPHIRERAPIDS`).

Whyborn:
I assume the usage message is just taken from the CMake build on the main branch? Worth removing the references to the MPI build?

SeanBryan51:
> I assume the usage message is just taken from the CMake build on the main branch? Worth removing the references to the MPI build?

Yep, the build script was taken from the main branch. It probably makes sense to remove MPI specifics or at least error with a warning when enabling MPI for the CABLE-POP_TRENDY branch considering it is currently not supported.

SeanBryan51:
FYI I've started documenting the differences in build configurations between CABLE-POP_TRENDY and main:
- https://github.com/CABLE-LSM/CABLE/issues/268

ccarouge:
Now that we have a green light to change the CABLE-POP_TRENDY build to CMake, I was reviewing this. Thanks for the comparison between the TRENDY compilation and main. 

I assume in this PR, we will want to go for a build that reproduces the old build as much as possible. Can you tell us why you changed the intel compiler and the compiler flags (removing all the -ax flags)?

SeanBryan51:
@ccarouge I agree, let's aim to reproduce the existing build configuration as close as possible. I've tested the CMake build only with the Intel (classic) compiler which is what is currently used in the `main` branch. The `CABLE-POP_TRENDY` Gadi build configuration uses the Intel LLVM compiler which supports the `-ax*` optimisation flags. I will try adding support for the Intel LLVM compiler.

SeanBryan51:
I have omitted some compiler flags in the CMake build - these are documented below:
- [Intel] Omit `-diag-disable=10382` as the warning was not produced during the build for either release or debug configurations.
- [Intel] Omit `-fpe-all=0 -no-ftz -ftrapuv -init=arrays,snan` as these flags are [commented out in the existing build script](https://github.com/CABLE-LSM/CABLE/blob/01db36403cc740394a544ff98be1077875188b74/offline/build.ksh#L135C194-L135C224).
- [Intel] Omit `-xHost` as the flag is not common to all build configurations and looks to be optional (see [`build.ksh`](https://github.com/CABLE-LSM/CABLE/blob/CABLE-POP_TRENDY/offline/build.ksh)).

ccarouge:
> [Intel] Omit -xHost as the flag is not common to all build configurations and looks to be optional (see [build.ksh](https://github.com/CABLE-LSM/CABLE/blob/CABLE-POP_TRENDY/offline/build.ksh)).

I'm not sure on this one. This flag allows for machine-specific optimisations to be applied. For me, it is the equivalent (with less flexibility) of "-march=broadwell -axSKYLAKE-AVX512,CASCADELAKE,SAPPHIRERAPIDS" that you added for intel-llvm. Same with `-march=native`. So I'm not sure why we would include the `-ax` line but remove the `-xHost`  option, I would include both as required. Happy to discuss this as I may have missed something.

ccarouge:
> [Intel] Omit -fpe-all=0 -no-ftz -ftrapuv -init=arrays,snan as these flags are [commented out in the existing build script](https://github.com/CABLE-LSM/CABLE/blob/01db36403cc740394a544ff98be1077875188b74/offline/build.ksh#L135C194-L135C224).

Sometimes these are kept commented out because the flags are sometimes needed but not always. I don't know if it is the case here or just old history being left behind. 
@mcuntz These flags appear for the `mcin` host setup, do you still switch them on from time to time?

SeanBryan51:
@ccarouge I'm happy with `-xHost` being set in `build.bash` on a per machine basis or specified at build time (via [`CMAKE_Fortran_FLAGS_RELEASE`](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG.html)) but it should not be part of the default flags defined in CMakeLists.txt (which cannot be overwritten at build time).  I think in general it is best to leave architecture specific optimisation flags outside the CMakeLists.txt file so that the build system is portable. Setting the flag in CMakeLists.txt could also prevent us from specifying optimisation flags at build time that target a processor other than the compilation host processor (like what is done with `-ax*` flags).

SeanBryan51:
@mcuntz requesting your review.

CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+# third party libs
+if(CABLE_MPI)
+    find_package(MPI REQUIRED COMPONENTS Fortran)
+endif()
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(NETCDF REQUIRED IMPORTED_TARGET "netcdf-fortran")
+
+set(CABLE_Intel_Fortran_FLAGS -fp-model precise -fpp -nofixed -assume byterecl -D__INTEL__ -D__INTEL_COMPILER__ -D__NETCDF3__)
+set(CABLE_Intel_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0 -debug extended -check all -warn all -fp-stack-check)
+set(CABLE_Intel_Fortran_FLAGS_RELEASE -O3 -ip)
```
mcuntz, line 25: These are my compiler flags. There is nothing fancy. I compared them to the main branch
```
set(CABLE_Intel_Fortran_FLAGS -fp-model precise)
set(CABLE_Intel_Fortran_FLAGS_DEBUG -O0 -g -traceback -fpe0)
set(CABLE_Intel_Fortran_FLAGS_RELEASE -O2)

```
 and I was wondering why the main branch does not have these switches. The debug version did not exist before so why not putting all checks and warnings? I also do not know why main has only -O2.

CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+    add_compile_options(
+        ${CABLE_Intel_Fortran_FLAGS}
+        "$<$<CONFIG:Release>:${CABLE_Intel_Fortran_FLAGS_RELEASE}>"
+        "$<$<CONFIG:Debug>:${CABLE_Intel_Fortran_FLAGS_DEBUG}>"
+    )
+endif()
+
+set(CABLE_GNU_Fortran_FLAGS -cpp -ffree-form -ffixed-line-length-132 -D__GFORTRAN__ -D__gFortran__ -D__CRU2017__ -D__NETCDF3__ -D__C13DEBUG__)
+set(CABLE_GNU_Fortran_FLAGS_DEBUG -O -g -pedantic-errors -Wall -W -Wno-maybe-uninitialized -fbacktrace -ffpe-trap=zero,overflow,underflow -finit-real=nan)
+set(CABLE_GNU_Fortran_FLAGS_RELEASE -O3 -Wno-aggressive-loop-optimizations)
```
mcuntz, line 36: One can remove `-D__CRU2017__`, which is old and probably not used anymore and `-D__C13DEBUG__`, which is only for me when working on 13C.

build.bash
```diff
@@ -0,0 +1,134 @@
...
+EOF
+}
+
+cmake_args=(-DCMAKE_BUILD_TYPE=Release)
+
+# Argument parsing adapted and stolen from http://mywiki.wooledge.org/BashFAQ/035#Complex_nonstandard_add-on_utilities
+while [ $# -gt 0 ]; do
+    case $1 in
+        --clean)
+            rm -r build
```
mcuntz, line 42: This is counter-intuitive. I just wanted to clean the run, i.e. removing the directories `bin` and `build`, something like `make clean` but it removed "only" `build` and started compiling.

build.bash
```diff
@@ -0,0 +1,134 @@
...
+            cmake_args+=(-DCMAKE_Fortran_FLAGS_RELEASE_INIT="-march=broadwell -axSKYLAKE-AVX512,CASCADELAKE,SAPPHIRERAPIDS -diag-disable=15009")
+            ;;
+        gnu)
+            module add gcc/13.2.0
+            compiler_lib_install_dir=GNU
+            [[ -n $mpi ]] && module add openmpi/4.1.4
+            ;;
+        ?*)
+            echo -e "\nError: compiler ${compiler} is not supported.\n"
+            exit 1
```
mcuntz, line 101: `;;` missing

build.bash
```diff
@@ -0,0 +1,134 @@
...
+        --clean)
+            rm -r build
+            ;;
+        --mpi)
+            mpi=1
+            cmake_args+=(-DCABLE_MPI="ON")
+            ;;
+        --compiler)
+            compiler=$2
+            shift
```
mcuntz, line 50: The standard in *nix tools is that the short form of options have the option, a space and then the argument, i.e. `-c gnu`, and the long form is with equal sign `--compiler=gnu`. Like this one can also pass arguments with a minus sign such as `--undef=-9999`. This is different here and confused me.
Why are you not using the `getopts` functionality of `bash`?

build.bash
```diff
@@ -0,0 +1,134 @@
...
+# Argument parsing adapted and stolen from http://mywiki.wooledge.org/BashFAQ/035#Complex_nonstandard_add-on_utilities
+while [ $# -gt 0 ]; do
+    case $1 in
+        --clean)
+            rm -r build
+            ;;
+        --mpi)
+            mpi=1
+            cmake_args+=(-DCABLE_MPI="ON")
+            ;;
```
mcuntz, line 47: I tried `./build.bash --mpi` and it rightfully told me that MPI compilation is not supported. I then removed `--mpi` and started `./build.bash`  but it kept telling me that MPI compilation is not supported. I had to add `--clean` to get it working again.

build.bash
```diff
@@ -0,0 +1,134 @@
...
+            [[ -n $mpi ]] && module add intel-mpi/2019.5.281
+            ;;
+        intel-llvm)
+            module add intel-compiler-llvm/2023.0.0
+            compiler_lib_install_dir=Intel
+            [[ -n $mpi ]] && module add intel-mpi/2019.5.281
+
+            # This is required to use ifort instead of ifx (default):
+            cmake_args+=(-DCMAKE_Fortran_COMPILER="ifort")
+
```
mcuntz, line 87: What is the difference between ifort and ifx? Why not using ifx?

build.bash
```diff
@@ -0,0 +1,134 @@
...
+        ?*)
+            cmake_args+=("$1")
+            ;;
+    esac
+    shift
+done
+
+if hostname -f | grep gadi.nci.org.au > /dev/null; then
+    : "${compiler:=intel-llvm}"
+
```
mcuntz, line 69: I just used the latest version of the Intel compiler so that I can do the same on all machines. I do not care if it is llvm or not. Last time I checked, the branch gives identical results for the GNU and the Intel compiler (output and restart files, i.e. bit-wise identical to double precision). So I guess and I hope that it is the same for different Intel compilers.

build.bash
```diff
@@ -0,0 +1,134 @@
...
+            # This is required to use ifort instead of ifx (default):
+            cmake_args+=(-DCMAKE_Fortran_COMPILER="ifort")
+
+            # This is required to build with runtime dispatch for all
+            # architectures on Gadi (see
+            # https://opus.nci.org.au/display/Help/Sapphire+Rapids+Compute+Nodes
+            # for more information):
+            cmake_args+=(-DCMAKE_Fortran_FLAGS_RELEASE_INIT="-march=broadwell -axSKYLAKE-AVX512,CASCADELAKE,SAPPHIRERAPIDS -diag-disable=15009")
+            ;;
```
mcuntz, line 93: Why is this not set also for the other Intel compiler?

build.bash
```diff
@@ -0,0 +1,134 @@
...
+            mpi=1
+            cmake_args+=(-DCABLE_MPI="ON")
+            ;;
+        --compiler)
+            compiler=$2
+            shift
+            ;;
+        -j)
+            CMAKE_BUILD_PARALLEL_LEVEL=$2
+            shift
```
mcuntz, line 54: I changed my opinion ;-)  I think that `-j` is only talkative if you know make. But who knows make? It is now also the only option that has a single dash. Perhaps `--ncpus`? But see my comment about single dash-double dash options.

mcuntz (changes requested):
Worked in principal on my machine (mc16). That is already great :-)

mcuntz:
> > [Intel] Omit -fpe-all=0 -no-ftz -ftrapuv -init=arrays,snan as these flags are [commented out in the existing build script](https://github.com/CABLE-LSM/CABLE/blob/01db36403cc740394a544ff98be1077875188b74/offline/build.ksh#L135C194-L135C224).
> 
> Sometimes these are kept commented out because the flags are sometimes needed but not always. I don't know if it is the case here or just old history being left behind. @mcuntz These flags appear for the `mcin` host setup, do you still switch them on from time to time?

Yes, I switch them on from time to time when I am looking for hard-to-find errors.
But I can do that in build.bash similar as optimisation is done with `cmake_args+=()`.

build.bash
SeanBryan51: This behaviour is also common to the main branch so I have made a new issue: https://github.com/CABLE-LSM/CABLE/issues/292

CMakeLists.txt
SeanBryan51: The flags in the main branch were left as is when doing the transition from Makefile to CMake (with the exception of https://github.com/CABLE-LSM/CABLE/issues/198 and https://github.com/CABLE-LSM/CABLE/issues/196). It is difficult to track down the origin of these flags in the commit history so I do not know why the flags are the way they are.

@mcuntz we should continue this discussion in this issue thread: https://github.com/CABLE-LSM/CABLE/issues/268 as this focuses on the merging of build configurations between main and CABLE-POP_TRENDY.

I would imagine including most of these flags in the main branch would be trivial but some flags will require testing.

build.bash
SeanBryan51: I did look into `getopt` and `getopts` when writing the build script. A problem I had with `getopts` was that long-named options were not supported. `getopt` on the other hand supports long-named options however I was discouraged from using it for other reasons (see [here](http://mywiki.wooledge.org/ComplexOptionParsing)). In the end I went with the approach suggested here: http://mywiki.wooledge.org/BashFAQ/035#Complex_nonstandard_add-on_utilities.

build.bash
SeanBryan51: Thanks for pointing this out, this is due to the `CABLE_MPI` option being cached and since the option is specified only when the `--mpi` flag is given, the cached value is not overwritten. We can fix this by specifying the `CABLE_MPI` option in the case that `--mpi` is not given and set the option to `OFF` in that case.

build.bash
SeanBryan51: This was done to reproduce the existing build script in CABLE-POP_TRENDY. I've attempted the compilation with `ifx` - it warns the `-ip` flag is no longer supported but the compilation succeeds so it will likely require us to update the flags for the release configuration.

I personally would try to reproduce the existing build as close as possible in this PR and have the update to `ifx` done as a separate pull request.

build.bash
SeanBryan51: It would be a good sanity check to verify results are identical across different intel compiler versions, I would be happy to do this. In the case that there are no issues, we can eliminate the separate compiler case for `intel-llvm` and use the latest version of the intel compiler when `$compiler == intel`.

build.bash
SeanBryan51: The [NCI documentation](https://opus.nci.org.au/display/Help/Sapphire+Rapids+Compute+Nodes) states to use these flags with the `intel-compiler-llvm` module. The intel classic compiler used in the `intel-compiler/2019.5.281` module does not support the `-ax*` optimisation flags.

build.bash
SeanBryan51: I'd be happy to go with `--ncpus` as it is more descriptive.

build.bash
ccarouge: The link to the approach you gave shows it is possible to have a mix of short and long options and to have long options as `--option value` or `--option=value`. How we implement it comes down to personal choice. I don't have a preference either way, but I agree with @mcuntz that a lot of command-line tools use the form `--option=value`. 

I wouldn't support both `--option value` and `--option=value` because it makes the arguments parsing quite a lot longer for very little added advantage.

SeanBryan51:
I have done a sanity check to verify results are identical across different intel compiler versions so that we can eliminate the separate compiler case for `intel-llvm` in `build.bash` and use the latest version of the intel compiler when `$compiler == intel`.

In short, I found no differences when using `ifort (IFORT) 2021.8.0 20221119` vs `ifort (IFORT) 19.0.5.281 20190815`. However, I found large differences (greater than 10%) across multiple variables (e.g. see `Ebal` plot below) when enabling the optimisation flags `-march=broadwell -axSKYLAKE-AVX512,CASCADELAKE,SAPPHIRERAPIDS -diag-disable=15009` for the latest Intel compiler (`intel-compiler-llvm/2023.0.0` module):

download

It is probably safe to upgrade the intel compiler to use the `intel-compiler-llvm/2023.0.0` module (`ifort (IFORT) 2021.8.0 20221119`) but I can't say the same for the architecture specific optimisations.

ccarouge:
@SeanBryan51 Is the curve showing the worsening imbalance over time for the compilation with the `-ax` option enabled?

SeanBryan51:
@ccarouge I have updated the plot to include a legend. Curve A shows the run without the optimisation flags and curve B shows the run with the optimisation flags.

SeanBryan51:
I will create a separate issue for the apparent divergence when using the architecture specific flags as it is out of the scope of this PR. These flags were already present in the existing Makefile build and the CMake build reproduces this exactly.

SeanBryan51:
Hi @mcuntz, a gentle reminder that I have re-requested your review. Your approval is required to merge this PR.

CMakeLists.txt
mcuntz: I would rather not comment in #268 but will do in #307 .

build.bash
mcuntz: All good to have long options with space. Makes the script the easiest to read. I guess:
```
-n)
    CMAKE_BUILD_PARALLEL_LEVEL=${2}
    shift
    ;;
--ncpus=?*)
    CMAKE_BUILD_PARALLEL_LEVEL=${1#*=}
    ;;

```
is not for everyone.

A last attempt from my side ;-) : why not having always a short option as well, such as -n for --ncpus, -m for --mpi, etc.

mcuntz (approved):
All good.
You can think a bout my last suggestion to have also single character options.
I would also be nitpicking and put really all variables in bash scripts in curly brackets, e.g. ${1}. They are firstly easier spotted/identified and it is needed when enhancing with variable substitution, default values, etc. I just made it a habit for me.

ccarouge (approved):
LGTM

Issue #240, opened 2024-04-19
Compiling the MPI executable using the GNU compiler fails


SeanBryan51:
Compiling the MPI executable for CABLE offline using the GNU compiler fails with the following error.

```
/home/189/sb8430/cable/src/offline/cable_mpiworker.F90:5686:49:

 5686 |                   //' (SUBROUTINE get_met_data)')
      |                                                 1
Error: Too many arguments in call to ‘abort’ at (1)
```

Steps to reproduce:

```
./build.bash --compiler gnu --mpi
```

The above command was run on commit: a6e0dccd19ee16158d64891979033f92d98197c5

SeanBryan51:
The error lies inside the `worker_time_update` function:

https://github.com/CABLE-LSM/CABLE/blob/46830b3773f1932680af158cab27ae223fd8685a/src/offline/cable_mpiworker.F90#L5650-L5842

The function looks to be unused in the code base.

ccarouge:
This is likely to change with the work to bring CABLE-POP_TRENDY in main and thus is deemed low priority

SeanBryan51:
This patch fixes the issue:
```diff
diff --git a/src/offline/cable_mpiworker.F90 b/src/offline/cable_mpiworker.F90
index f240e55..bf5e64d 100644
--- a/src/offline/cable_mpiworker.F90
+++ b/src/offline/cable_mpiworker.F90
@@ -5655,6 +5655,7 @@ USE cbl_soil_snow_init_special_module

   SUBROUTINE worker_time_update (met, kend, dels)

+    USE cable_abort_module, ONLY: abort
     USE cable_common_module, ONLY: ktau_gl
     USE cable_def_types_mod
     USE cable_IO_vars_module
```

SeanBryan51:
This was fixed in https://github.com/CABLE-LSM/CABLE/pull/410

PR #241, opened 2024-04-19
Add support for MPI builds with the GNU compiler

### LINKED ISSUE #239: Add support for MPI builds with the GNU compiler on Gadi
The build script currently does not support MPI builds with the GNU compiler on Gadi. Executing `./build.bash --compiler gnu --mpi` on Gadi produces the following error when compiling MPI source files:

```
f951: Fatal Error: Reading module ‘/apps/intel/compilers_and_libraries_2019.5.281/linux/mpi/intel64/include/mpi.mod’ at line 1 column 2: Unexpected EOF
compilation terminated.
gmake[2]: *** [CMakeFiles/cable-mpi.dir/build.make:75: CMakeFiles/cable-mpi.dir/src/offline/cable_mpicommon.F90.o] Error 1
```

This is due to the `intel-mpi` module being loaded when we should be using an MPI module that is compatible with the chosen compiler. See below:

https://github.com/CABLE-LSM/CABLE/blob/46830b3773f1932680af158cab27ae223fd8685a/build.bash#L92-L94

SeanBryan51:
The build script currently does not support MPI builds with the GNU compiler on Gadi. This is due to the `intel-mpi` module being loaded when we should be using an MPI module that is compatible with the chosen compiler. This change loads the `openmpi` module when GNU compilation is enabled in the build script. 

Fixes #239

----
📚 Documentation preview 📚: https://cable--241.org.readthedocs.build/en/241/

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -75,10 +75,12 @@ if hostname -f | grep gadi.nci.org.au > /dev/null; then
         intel)
             module add intel-compiler/2019.5.281
             compiler_lib_install_dir=Intel
+            [[ -n $mpi ]] && module add intel-mpi/2019.5.281
             ;;
         gnu)
             module add gcc/13.2.0
             compiler_lib_install_dir=GNU
+            [[ -n $mpi ]] && module add openmpi/4.1.4
             ;;
         ?*)
             echo -e "\nError: compiler ${compiler} is not supported.\n"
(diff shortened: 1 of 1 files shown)


SeanBryan51:
After fixing the compiler issue documented in #240, I've verified the build succeeds with:
```bash
./build.bash --compiler gnu --mpi --clean
```

ccarouge (approved):
Looks good.

Issue #242, opened 2024-04-19
Merging CABLE-POP into CABLE3 - Part1


JhanSrbinovsky:
I will populate this over the next few days with a branch containing already partially completed tasks and a description of what has been done. 

For now I have included some notes.

# Background

* A decade ago the ultimate/latest version of CABLE was CABLE1.4b.

* CMIP5 version of CABLE in ACCESS1.3 corresponds to CABLE1.4c.

* Following CMIP5, we created CABLE2 to resolve 2 issues: 
     1. The significant divergence between 1.4b and 1.4c
     2. Establishing CABLE as a version controlled community model 

* Concurrently, we included land biogeochemistry into ACCESS1.3 to create ACCESS1.4.

* Ocean biogeochemistry was then included and ACCESS1.4 beacame ESM1.5. 

* CMIP6 version of CABLE in ESM1.5 corresponds to CABLE2.4
* CMIP6 version of CABLE in ACCESS-CM2 corresponds to CABLE2.5

* The JAC(JULES and CABLE) project represents a long term aim to include CABLE as an alternative LSM to JULES in the UKMO framework. CABLE2.5 was the basis for JAC. 

* CABLE3 is the resultant(modified) JAC version of CABLE2.5, but essentially reproduces CABLE2.5

# Issue

Below are listed the proposed CMIP7 variants (as per my understanding today) ::

 ### ESM1.6 

This model would cover our fast-track submissions and include: 

    - UM7.3 
    - MOM6 (at higher resolution) 
    - CABLE3  
    - WOMBAT 
    - CASA (with thinning/harvesting & potentially revised LUC scheme)

 ### ACCESS-CM3

This model would cover our coupled physical climate submissions:

    - UM13.1
    - MOM6 (at higher resolution) 
    - CABLE3  
 
 ### ACCESS-ESM3

This model would cover our coupled ESM submissions and include: 

    - UM13.1
    - MOM6 (at higher resolution) 
    - CABLE4  
    - WOMBAT 
    - CASA (with thinning/harvesting & potentially revised LUC scheme)

All of the above CMIP7 variants (except ACCESS-ESM3) use CABLE3. CABLE3 being the most recent version of CABLE. We have (currently being reviewed) a working version of (ACCESS-)AM3, an atmosphere only version of CM3, which in turn logically forms the basis of an ESM3. The clear difference here is that ESM3 is proposed to use CABLE4. 

[...]

PR #243, opened 2024-04-19
Do not merge to main - 242 merging cable pop into cable3 part1

JhanSrbinovsky:
## Description

Do not merge to main 

See #242 for full description. Included here are notes per revision


JhanSrbinovsky:
Here the trunk refers to CABLE-POP. new is CABLE3 merged into CABLE-POP. up until CALL define_canopy in cbm. Also call to radiation from canopy is converted. Also uses CABLE3 style of sinbet() in cable_input.  

Screenshot 2024-04-21 at 12 03 52 AM

PR #274, opened 2024-04-26
Reconcile versions of rplant

### LINKED ISSUE #163: ESM1.5 reconciliation: casa_rplant
`type_defect`   |    by rml599

___

casa_rplant.F90 does not perform as expected. Although it is intended to reproduce the behaviour of ESM1.5 it produces plant respiration (as output in field1390) that is reduced by around 30%.

casa_rplant.F90 contains two similar subroutines, casa_rplant and casa_rplant1. casa_rplant contains additional code for acclimation and to accomodate POP. casa_rplant1 is similar but not identical to the casa_rplant subroutine in CABLE2.4.

As currently written these subroutines are called from biogeochem_casa

```
65	#ifdef ESM15
66	CALL casa_rplant(veg,casabiome,casapool,casaflux,casamet,climate)
67	#else
68	CALL casa_rplant1(veg,casabiome,casapool,casaflux,casamet)
69	#endif
```

So casa_rplant is called if trying to reproduce ESM1.5 behaviour and casa_rplant1 in all other cases.

Tests running casa_rplant1 instead corrects the reduced plant respiration.

Most of the additional code in casa_rplant is not run for ESM1.5-type runs since it requires CALL_climate to be true. When CALL_climate is not true, there are two sections which are modified for an ESM15 compile.

```
227	#             ifdef ESM15
228	                casaflux%crmplant(:,wood)  = casabiome%rmplant(veg%iveg(:),wood) &
229	#             else
230	                casaflux%crmplant(:,wood)  =  resp_coeff * casaflux%frac_sapwood(:) * &
231	                     casabiome%rmplant(veg%iveg(:),wood) &
232	#             endif
```

```
246	#         ifdef ESM15
247	            casaflux%crmplant(:,froot) = casabiome%rmplant(veg%iveg(:),froot) &
248	#         else
249	            casaflux%crmplant(:,froot) =  resp_coeff * casabiome%rmplant(veg%iveg(:),froot) &
250	#         endif
```

However, ESM15 is not currently defined (either YES or NO) in the start of the file and defaults to NO. This means the ESM15 specific code was not run when the plant respiration tests were done. For the second instance (lines 246-250), this makes no difference as resp_coeff # 1 for CALL_climate=.false. In the first case, the two options should be the same if frac_sapwood 1 which should be the case unless POP is being run.

frac_sapwood is initialized in casa_inout.F90 in the casa_init subroutine.
```
156	    casaflux%frac_sapwood = 0.
```

It is then conditionally modified but only if it is read in from an offline restart file, or casa_fromzero is .true. Neither will be the case for an ESM15 compilation/run.

```
214	    IF (initcasa==1) THEN
215	       IF (.NOT.cable_user%casa_fromzero) THEN
216	#ifndef UM_CBL
217	          CALL READ_CASA_RESTART_NC (  casamet, casapool, casaflux, phen )
218	#endif
219	       ELSE
220	          WRITE(*,*)'casa_init: not using restart file!'
221	          WRITE(*,*)'Using input from readbiome.!'
222	          WRITE(*,*) 'initialising frac_sapwood=1 and sapwood_area = 0)'
223	          casaflux%frac_sapwood(:) = 1.0
224	          casaflux%sapwood_area(:) = 0.0
225	       ENDIF
226	    ENDI
```

Hence frac_sapwood=0 is passed into the casa_rplant subroutine with the consequence that the plant respiration term has no contribution from the wood pool, only from the labile pool and fine roots. This explains why the reduced plant respiration was only seen as a reduction for woody pfts.

The solution is to initialise frac_sapwood to 1 at line 156. I think this should be appropriate for all applications, not just ESM15. 

___
_Issue migrated from [trac:362](https://trac.nci.org.au/trac/cable/362) at 2023-11-27 11:44:53 +1100_

JhanSrbinovsky:
Fixes #163 

initialization to 1 so is applicable even  IF not over-written by specific case applications. i.e. Will be correct for ESM1.5

# CABLE

## Description

Fixes #(issue)

## Type of change

- [x] Bug fix

## Checklist

----
📚 Documentation preview 📚: https://cable--274.org.readthedocs.build/en/274/

### DIFF
--- a/src/science/casa-cnp/casa_rplant.F90
+++ b/src/science/casa-cnp/casa_rplant.F90
@@ -80,11 +80,11 @@ USE casa_cnp_module, ONLY : vcmax_np
     Ygrow        = 0.0
 
     WHERE(casapool%Nplant>0.0)
-       ratioPNplant = casapool%Pplant/(casapool%Nplant+ 1.0e-10)
+      ratioPNplant = casapool%Pplant / casapool%Nplant                   
     ENDWHERE
 
-    Ygrow(:) = 0.65+0.2*ratioPNplant(:,leaf)/(ratioPNplant(:,leaf)+1.0/15.0)
-    Ygrow(:) = min(0.85,max(0.65,Ygrow(:)))
+    Ygrow(:) = 0.65 + 0.2 * ratioPNplant(:,leaf) /                             &
+                      ( ratioPNplant(:,leaf) + 1.0 / 15.0 )
 
     casaflux%crmplant(:,wood) = 0.0
     casaflux%crmplant(:,froot) = 0.0
--- a/src/science/casa-cnp/casa_rplant.F90
+++ b/src/science/casa-cnp/casa_rplant.F90
@@ -1,3 +1,4 @@
+!#define ESM15 YES
 !# define UM_CBL YES
 !==============================================================================
 ! This source code is part of the
--- a/src/science/casa-cnp/casa_rplant.F90
+++ b/src/science/casa-cnp/casa_rplant.F90
@@ -1,4 +1,3 @@
-!#define ESM15 YES
 !# define UM_CBL YES
 !==============================================================================
 ! This source code is part of the
@@ -225,12 +224,8 @@ USE casa_cnp_module, ONLY : vcmax_np
           WHERE(casamet%tairk >250.0)
              WHERE(casapool%cplant(:,wood)>1.0e-6)
 
-#             ifdef ESM15 
-                casaflux%crmplant(:,wood)  = casabiome%rmplant(veg%iveg(:),wood) &
-#             else
                 casaflux%crmplant(:,wood)  =  resp_coeff * casaflux%frac_sapwood(:) * &
                      casabiome%rmplant(veg%iveg(:),wood) &
-#             endif
                      * casapool%nplant(:,wood)             &
                      * EXP(308.56*(1.0/56.02-1.0           &
                      / (casamet%tairk(:)+46.02-tkzeroc)))
@@ -244,11 +239,7 @@ USE casa_cnp_module, ONLY : vcmax_np
           ENDWHERE
           WHERE(casamet%tsoilavg >250.0.AND.casapool%cplant(:,froot)>1.0e-6)
 
-#         ifdef ESM15 
-            casaflux%crmplant(:,froot) = casabiome%rmplant(veg%iveg(:),froot) &
-#         else
             casaflux%crmplant(:,froot) =  resp_coeff * casabiome%rmplant(veg%iveg(:),froot) &
-#         endif
                   * casapool%nplant(:,froot)             &
                   * EXP(308.56*(1.0/56.02-1.0            &
                   / (casamet%tsoilavg(:)+46.02-tkzeroc)))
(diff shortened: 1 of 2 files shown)


rml599gh (commented):
I think these changes will fix the immediate bug that results in reduced plant respiration for ESM1.5 cases - but it would be better if we can do the longer-term fix that properly reconciles the rplant versions and doesn't need the ifdef ESM15 conditions in casa_rplant.F90 or biogeochem_casa.F90

JhanSrbinovsky:
> I think these changes will fix the immediate bug that results in reduced plant respiration for ESM1.5 cases - but it would be better if we can do the longer-term fix that properly reconciles the rplant versions and doesn't need the ifdef ESM15 conditions in casa_rplant.F90 or biogeochem_casa.F90

Following the comments about Grow in that TRAC ticket:

`The final difference is in the treatment of cases where Cnpp < 0. This section needs some further work as the logic in the casa_rplant1 version seems flawed (for setting Cnpp to zero when the correction still has not solved the negative NPP). It looks like the zero-ed NPP then gets recalculated anyway outside the ENDWHERE statement.`

....seems to suggest that rplant1 is necessary to retain for ESM1.5? Waiting on /g/data/ to determine how extensively #def ESM15 is even used. From memory it was more than just two files

JhanSrbinovsky:
Gadi has returned - so ESM15 in general exists across 11 files:

```
util/cable_common.F90:          um_radiation = .FALSE., um_hydrology = .FALSE., esm15 = .FALSE.
coupled/ESM1.5/casa_landuse.F90:# define ESM15 YES
science/casa-cnp/biogeochem_casa.F90:!#define ESM15 YES
science/casa-cnp/casa_cnp.F90:!#define ESM15 YES
science/casa-cnp/casa_param.F90:  REAL(r_2), PARAMETER :: ratioPCstrfix = ratioNCstrfix/25.0 !ESM15:casa_landuse
science/casa-cnp/casa_rplant.F90:!#define ESM15 YES
science/misc/cable_climate.F90:# ifndef ESM15
science/soilsnow/cbl_surfbv.F90:    IF( .NOT. cable_runtime%UM  .OR. cable_runtime%esm15 ) THEN
coupled/ESM1.5/CABLEfilesFromESM1.5/cable_data.F90:#define ESM15 YES
coupled/ESM1.5/CABLEfilesFromESM1.5/cable_explicit_driver.F90:   cable_runtime%esm15= .TRUE.
coupled/ESM1.5/CABLEfilesFromESM1.5/cable_implicit_driver.F90:      cable_runtime%esm15= .TRUE.
```

However note in the first instance above it is the declaration of a cable_runtime% switch that is switched on in explicit/implicit. I  vaguely recall now that there were instances where it had to be there, and others where there was still work to be done. Like this one around rplant. Of course the others dont HAVE to be there either, only to retain the behaviour (physical at least) of the model.  

Sorry, re-reading above comment - so With initialisation change AND using rplant() instead of rplant1() then rplant1() can disappear?

ccarouge:
In my read of the ticket, I think we also need to change `casa_rplant`:

- remove the 1.e-10 from:
https://github.com/CABLE-LSM/CABLE/blob/3d539e962ebc26e52995978f07d17a4f9325040a/src/science/casa-cnp/casa_rplant.F90#L82-L84

- remove this line:
https://github.com/CABLE-LSM/CABLE/blob/3d539e962ebc26e52995978f07d17a4f9325040a/src/science/casa-cnp/casa_rplant.F90#L87

And I have a question: what configuration(s) are you doing to evaluate the changes?

rml599gh:
No evaluation has been done since the initial trac ticket. They were short (~1 month) ESM1.5-CABLE3 runs and should probably be confirmed now that we are working under git. Is there an offline case with CASA that can also be used for testing?

ccarouge:
> Is there an offline case with CASA that can also be used for testing?

The only CASA case I know of is the one for TRENDY but that's quite different to what we'd want here.

JhanSrbinovsky:
I'v'e been using this GSWP2 for global benchmarking since GSWP2 was the latest and greatest. The three single sites also work. They cover a few different scenarios, including snow which is traditionally the problematic one:

`https://svnserv.csiro.au/svn/CABLE/dev/users/jxs599/GenericCABLETicketDev`

JhanSrbinovsky:
Sorry, thats CSIRO server. I'll co to /scratch/public/jxs599

rml599gh:
If we could get grid-cell timeseries output from ACCESS for the relevant forcing variables, perhaps there is value in setting up a test case to run offline at those grid-cells (i.e. like the benchcab site test) with casa. I guess we'd need to work out how to extract data from a restart file to provide the initial conditions for the offline test.

JhanSrbinovsky:
I included the changes suggested by @ccarouge. It seems that unless  climate rplant1

src/science/casa-cnp/casa_rplant.F90
```diff
@@ -1,3 +1,4 @@
+!#define ESM15 YES
```
ccarouge, line 1: Actually, the intent is to remove ESM15 here. So let the `#define` out. And remove the `#ifdef` cases. They are not needed anymore since `casaflux%frac_sapwood` is initialised to 1.

ccarouge (changes requested):
I am finally looking at this.

I think we are getting to the end here. Just need:
- a change for ESM15 (remove the pre-processor cases)
- remove `casa_rplant1` (@rml599gh are we ready to remove `rplant1` or do we want more tests?`

@rml599gh Do we want to add at the end of `casa_rplant` a reset of Cnpp to 0 if negative? Something like:

```fortran
WHERE(casaflux%Cnpp < 0.0)
      casaflux%cnpp(:) = 0.0
ENDWHERE
```
Because someone added that in rplant1 (even if misplaced), saying the delcrm* corrections would not necessarily always work.

ccarouge:
@JhanSrbinovsky It'd be good to have @rml599gh view about the negative Cnpp before putting this through. Otherwise, I'm happy with it. If you talk to Rachel before me, maybe you can ask her.

JhanSrbinovsky:
no problem. I think she is on leave for school holidays

rml599gh:
As Claire noted - if we do want to include these lines
```
      WHERE(casaflux%Cnpp < 0.0)
         casaflux%cnpp(:) = 0.0
      ENDWHERE
```
they would need to come right at the end of the subroutine since Cnpp is re-calculated outside of the final WHERE loop. Presumably if the correction hasn't quite worked and we just reset to zero, I'm guessing we may be impacting carbon conservation.
Is it worth bringing the lines across into rplant but leaving them commented out for now, then starting a new issue to resolve Cnpp<0 cases throughout CASA. For example casa_allocation in casa_cnp.F90 also has negative NPP tests in two blocks, one apparently Ying-Ping's and one Vanessa's.

```
          ! added in for negative NPP and one of biomass pool being zero ypw 27/jan/2014
            WHERE(casaflux%Cnpp<0.0)
               casaflux%fracCalloc(:,leaf)  = casaflux%Crmplant(:,leaf)/SUM(casaflux%Crmplant,2)
               casaflux%fracCalloc(:,wood)  = casaflux%Crmplant(:,wood)/SUM(casaflux%Crmplant,2)
               casaflux%fracCalloc(:,froot) = casaflux%Crmplant(:,froot)/SUM(casaflux%Crmplant,2)
            ENDWHERE

          ! vh_js !
          ! as long as biomass is positive, adjust allocation to be
          ! proportional to stock when NPP -ve   (Ticket#108)
            WHERE(casaflux%Cnpp<0.0 .AND. SUM(casapool%Cplant,2)>0  )
               casaflux%fracCalloc(:,leaf)  = casapool%Cplant(:,leaf)/SUM(casapool%Cplant,2)
               casaflux%fracCalloc(:,wood)  = casapool%Cplant(:,wood)/SUM(casapool%Cplant,2)
               casaflux%fracCalloc(:,froot) = casapool%Cplant(:,froot)/SUM(casapool%Cplant,2)
            ENDWHERE

```

JhanSrbinovsky:
There is almost more commented code in CASA than operational code, so the thought of adding to it makes me shiver :). Although I see the merit in it. I think it is worth **raising as an issue to be fixed later and then point back to here?**

Specifically, the fact that it goes -ve at all is a worry. Apart from possibly justifiable avoiding /0, sqrt(-ve) etc it all seems a bit VW modelling-ish.  ie MPG = MAX(6.4, MPG); MPG=MIN(MPG,6.4) - Hooray MPG=6.4. It certainly isnt just here though, just a comment :) 

That said, I prefer to limit it with a Cnpp = MAX(0, Cnpp), even though it is effectively the same thing

ccarouge:
I have added the issue #337 . Let's not keep comment code in, I've put the code in the issue so we know it used to be here but without effect. I'm accepting this now.

ccarouge (approved):

rkutteh:
@JhanSrbinovsky @ccarouge While trying to use main for the ice consistency check I noticed that the code does not compile anymore with error:
CABLE/src/science/casa-cnp/biogeochem_casa.F90(23): error #6580: Name in only-list does not exist or is not accessible.   [CASA_RPLANT1]
USE casa_rplant_module, ONLY: casa_rplant1
------------------------------^

JhanSrbinovsky:
IF you change that _rplant1 to plant it will. That should be the only change you need. Can you check t hat it is  and I'll upload the change formally.

PR #287, opened 2024-05-08
Range checks cable parameters

### LINKED ISSUE #225: Range checks on state variables/parameters
- Subroutine that bound-checks ranges for every timestep (inclusive)
- Some parameters will be constant after processing input, so we need to check them only once. First step would be identifying and separating different types of variables.
- Ask on the forums for what bounds we would need for each of those variables
- Switch for various ways for error-handling - (a) no checks for bounds, if out-of-bounds either (b) exit or (c) put warnings in logger (for now, by default throw necessary error and exit the program) <- TODO: check whether there is an existing/similar option for the same.

abhaasgoyal:
# CABLE

## Description

Fixes #225 

Add 2 flags to range checks

```
check%ranges (0, 1, 2) - When to do the range checks NO_CHECK, ON_TIMESTEP, ON_WRITE
check%exit (.true., .false.) - Whether to exit the program if range checks are done, and out of range - the other option is to provide a warning
```

- Checks are being done in the following stages
   1. Parameters of `soil` and `veg` before running the spinloop
   2. Parameters when creating the NetCDF output file (some of them clash with `soil` and `veg` but code can be refactored in the future to account for that)
   3. Landuse variables I/O
   4. Variables in creating restart file
   5. Variables at every timestep
- Redesign `cable_output` with introducing `generate_out_write_acc`, and `check_and_write`
- Change the range checks to be done in `cable_write` during `cable_output`.
- Make `output_inclusion_type` dependencies clear between group output and individual values

Tested with experiments in `bench_example` for `main` vs current commit.

----
📚 Documentation preview 📚: https://cable--287.org.readthedocs.build/en/287/

### DIFF
--- a/src/offline/cable_abort.F90
+++ b/src/offline/cable_abort.F90
@@ -18,9 +18,9 @@
 !
 ! ==============================================================================
 
-
 MODULE cable_abort_module
 
+  USE cable_IO_vars_module, ONLY: check, logn
   IMPLICIT NONE
 
 CONTAINS
@@ -60,13 +60,12 @@ CONTAINS
   !
   !==============================================================================
 
-
-  SUBROUTINE abort( message )
+  SUBROUTINE abort(message)
 
     ! Input arguments
     CHARACTER(LEN=*), INTENT(IN) :: message
 
-    WRITE(*, *) message
+    WRITE (*, *) message
     STOP 1
 
   END SUBROUTINE abort
@@ -109,8 +108,7 @@ CONTAINS
   !
   !==============================================================================
 
-
-  SUBROUTINE nc_abort( ok, message )
+  SUBROUTINE nc_abort(ok, message)
 
     USE netcdf
 
@@ -118,8 +116,8 @@ CONTAINS
     CHARACTER(LEN=*), INTENT(IN) :: message
     INTEGER, INTENT(IN) :: ok
 
-    WRITE(*,*) message ! error from subroutine
-    WRITE(*,*) NF90_STRERROR(ok) ! netcdf error details
+    WRITE (*, *) message ! error from subroutine
+    WRITE (*, *) NF90_STRERROR(ok) ! netcdf error details
 
     STOP
 
@@ -140,59 +138,55 @@ CONTAINS
   !
   !==============================================================================
 
-
-
-  SUBROUTINE range_abort(message,ktau,met,value,var_range,                       &
-       i,xx,yy)
+  SUBROUTINE range_abort(vname, ktau, met, value, var_range, i, xx, yy)
 
     USE cable_def_types_mod, ONLY: met_type
-    USE cable_IO_vars_module, ONLY: latitude,longitude,landpt,lat_all,lon_all
+    USE cable_IO_vars_module, ONLY: latitude, longitude, &
+                                    landpt, lat_all, lon_all
 
     ! Input arguments
-    CHARACTER(LEN=*), INTENT(IN) :: message
+    CHARACTER(LEN=*), INTENT(IN) :: vname
 
-    INTEGER, INTENT(IN) ::                                                      &
-         ktau, & ! time step
-         i       ! landpt number of erroneous grid square
+    INTEGER, INTENT(IN) :: &
+      ktau, & ! time step
+      i       ! tile number along mp
 
-    INTEGER,INTENT(IN),OPTIONAL ::                                              &
-         xx, & ! coordinates of erroneous grid square
-         yy    ! coordinates of erroneous grid square
+    REAL, INTENT(IN) :: &
+      xx, & ! coordinates of erroneous grid square
+      yy    ! coordinates of erroneous grid square
 
+    TYPE(met_type), INTENT(IN) :: met  ! met data
 
-    TYPE(met_type),INTENT(IN) :: met  ! met data
+    REAL(4), INTENT(IN) :: value ! value deemed to be out of range
 
-    REAL,INTENT(IN) :: value ! value deemed to be out of range
+    REAL, INTENT(IN) :: var_range(2) ! appropriate var range
 
-    REAL,DIMENSION(2),INTENT(IN) :: var_range ! appropriate var range
+    INTEGER :: iunit
 
-    WRITE(*,*) "in SUBR range_abort"
-    WRITE(*,*) message ! error from subroutine
-
-    IF( PRESENT(yy) ) THEN ! i.e. using rectangular land/sea grid
-
-       WRITE(*,*) 'Site lat, lon:',lat_all(xx,yy),lon_all(xx,yy)
-       WRITE(*,*) 'Output timestep',ktau,                                       &
-            ', or ', met%hod(landpt(i)%cstart),' hod, ',                  &
-            INT( met%doy( landpt(i)%cstart) ),'doy, ',                   &
-            INT( met%year(landpt(i)%cstart) )
-
-    ELSE ! i.e. using compressed land only grid
-
-       WRITE(*,*) 'Site lat, lon:', latitude(i), longitude(i)
-       WRITE(*,*) 'Output timestep', ktau,                                       &
-            ', or ', met%hod( landpt(i)%cstart ), ' hod, ',               &
-            INT( met%doy( landpt(i)%cstart) ),'doy, ',                    &
-            INT( met%year( landpt(i)%cstart) )
-
+    IF (check%exit) THEN
+      iunit = 6
+    ELSE
+      iunit = logn ! warning
     END IF
 
-    WRITE(*,*) 'Specified acceptable range (checks.f90):', var_range(1),        &
-         'to',var_range(2)
+    WRITE (iunit, *) "in SUBR range_abort: Out of range"
+    WRITE (iunit, *) "for var ", vname ! error from subroutine
 
-    WRITE(*,*) 'Value:',value
+    ! patch(i)%latitude, patch(i)%longitude
+    WRITE (iunit, *) 'Site lat, lon:', xx, yy
+    WRITE (iunit, *) 'Output timestep', ktau, &
+      ', or ', met%hod(i), ' hod, ', &
+      INT(met%doy(i)), 'doy, ', &
+      INT(met%year(i))
 
-    STOP
+    WRITE (iunit, *) 'Specified acceptable range (cable_checks.f90):', &
+      var_range(1), 'to', var_range(2)
+
+    WRITE (iunit, *) 'Value:', value
+
+    IF (check%exit) THEN
+      STOP
+    END IF
 
   END SUBROUTINE range_abort
 
--- a/src/offline/cable_checks.F90
+++ b/src/offline/cable_checks.F90
@@ -30,160 +30,434 @@ MODULE cable_checks_module
   ! particular sections of the code - largely for diagnostics/fault finding.
   ! rh_sh - converts relative to sensible humidity if met file units require it
   !
+  USE cable_IO_vars_module, ONLY: patch
+  USE cable_abort_module, ONLY: range_abort
   USE cable_def_types_mod
   USE cable_common_module, ONLY: cable_user
 
   IMPLICIT NONE
 
   PRIVATE
-  PUBLIC ranges_type, ranges, mass_balance, energy_balance, rh_sh
-  
+  PUBLIC constant_check_range, check_range, ranges_type, ranges, mass_balance, energy_balance, rh_sh
+
   TYPE units_type
-     CHARACTER(LEN=1) :: Rainf ! 's' (mm/s) or 'h' (mm/h)
-     CHARACTER(LEN=1) :: PSurf  ! 'h'(hPa or mbar) or 'P'(Pa)
-     CHARACTER(LEN=1) :: Tair  ! 'C' or 'K'
-     CHARACTER(LEN=1) :: Qair  ! '%' or 'g' (spec hum)
-     CHARACTER(LEN=1) :: CO2air ! 'p' (ppmv)
-     CHARACTER(LEN=1) :: Wind ! 'm'(m/s)
+    CHARACTER(LEN=1) :: Rainf ! 's' (mm/s) or 'h' (mm/h)
+    CHARACTER(LEN=1) :: PSurf  ! 'h'(hPa or mbar) or 'P'(Pa)
+    CHARACTER(LEN=1) :: Tair  ! 'C' or 'K'
+    CHARACTER(LEN=1) :: Qair  ! '%' or 'g' (spec hum)
+    CHARACTER(LEN=1) :: CO2air ! 'p' (ppmv)
+    CHARACTER(LEN=1) :: Wind ! 'm'(m/s)
   END TYPE units_type
   TYPE(units_type) :: units
 
   TYPE ranges_type
-     REAL, DIMENSION(2) ::                                               &
-          nav_lon = (/-360.0,360.0/),         &
-          nav_lat = (/-90.0,90.0/),           &
-          time,                               &
-          timestp,                            &
-                                ! possible forcing variables for CABLE
-          SWdown = (/0.0,1360.0/),            & ! W/m^2
-          LWdown = (/0.0,950.0/),             & ! W/m^2
-          Rainf = (/0.0,0.1/),               & ! mm/s
-          Snowf = (/0.0,0.1/),             & ! mm/s
-          PSurf = (/500.0,1100.0/),           & ! mbar/hPa
-          Tair = (/200.0,333.0/),             & ! K
-          Qair = (/0.0,0.1/),                & ! g/g
-          Tscrn = (/-70.0,70.0/),             & ! oC - INH
-          Qscrn = (/0.0,0.1/),                & ! kg/kg
-          CO2air = (/160.0,2000.0/),          & ! ppmv
-          Wind = (/0.0,75.0/),                & ! m/s
-          Wind_N = (/-75.0,75.0/),            & ! m/s
-          Wind_E = (/-75.0,75.0/),            & ! m/s
-                                ! possible output variables
-          Qmom = (/-10.0,8000.0/),            & ! kg/m/s2 - (INH generous range)
-          Qh = (/-2000.0,2000.0/),            & ! W/m^2
-          Qle = (/-2500.0,2500.0/),           & ! W/m^2
-          Qg = (/-4000.0,4000.0/),            & ! W/m^2
-          SWnet = (/0.0,1350.0/),             & ! W/m^2 (YP oct07)
-                                ! SWnet = (/0.0,1250.0/),            & ! W/m^2
-          LWnet = (/-500.0,510.0/),           & ! W/m^2
-          Rnet = (/-500.0,1250.0/),           & ! W/m^2
-          Evap = (/-0.0045,0.0045/),         &  ! note this is also used for snow melt !
-          Ewater = (/-0.0005,0.0005/),        &
-          ESoil = (/-0.0015,0.0015/),         &
-          TVeg = (/-0.0003,0.0003/),          &
-          ECanop = (/-0.0003,0.0003/),        &
-          PotEvap = (/-0.0006,0.0006/),       &
-          ACond = (/0.0,1.0/),                &
-          SoilWet = (/-0.4,1.2/),             &
-          Albedo = (/0.0,1.0/),               &
-          visAlbedo = (/0.0,1.0/),            & ! vars intro for Ticket #27
-          nirAlbedo = (/0.0,1.0/),            & ! vars intro for Ticket #27
-          VegT = (/213.0,353.0/),             &
-          SoilTemp = (/213.0,353.0/),         &
-          SoilMoist = (/0.0,2000.0/),         &
-          Qs = (/0.0,15.0/),                   &
-          Qsb = (/0.0,15.0/),                  &
-          DelSoilMoist  = (/-2000.0,2000.0/), &
-          DelSWE  = (/-2000.0,2000.0/),       &
-          DelIntercept = (/-100.0,100.0/),    &
-          SnowT  = (/213.0,280.0/),           &
-          BaresoilT = (/213.0,343.0/),        &
-          AvgSurfT = (/213.0,333.0/),         &
-          RadT = (/200.0,373.0/),             &
-                                ! vh_js !
-          SWE = (/0.0,4000.0/),               &
-          RootMoist = (/0.0,2000.0/),         &
-          CanopInt = (/0.0,100.0/),           &
-          NEE = (/-70.0,50.0/),               & ! umol/m2/s
-          NPP = (/-20.0,75.0/),               & ! umol/m2/s
-          GPP = (/-20.0,100.0/),              & ! umol/m2/s
-          PAR = (/-1000.0,5000.0/),              & ! umol/m2/s
-          AutoResp = (/-50.0,20.0/),          & ! umol/m2/s
-          LeafResp = (/-50.0,20.0/),          & ! umol/m2/s
-          HeteroResp = (/-50.0,20.0/),        & ! umol/m2/s
-          HSoil = (/-1000.0,1000.0/),         &
-          HVeg = (/-1000.0,1000.0/),          &
-          SnowDepth = (/0.0,50.0/),           & ! EK nov07
-          Wbal = (/-999999.0,999999.0/),      &
-          Ebal = (/-999999.0,999999.0/),      &
-                                ! vh_js !
-          CanT = (/213.0,333.0/),      &
-          Fwsoil = (/0.0,1.0/),      &
-                                ! parameters:
-          albsoil = (/0.0,0.9/),              &
-          isoil = (/1.0,30.0/),               &
-          iveg = (/1.0,30.0/),                &
-          bch = (/2.0,15.0/),                 &
-          latitude = (/-90.0,90.0/),          &
-          c3 = (/0.0,1.0/),                   & ! EK nov07
-          clay = (/0.0,1.0/),                 &
-          css = (/700.0,2200.0/),             &
-          rhosoil = (/300.0,3000.0/),         &
-          hyds = (/5.0E-7,8.5E-3/),           & ! vh_js ! sep14
-          rs20 = (/0.0,10.0/),                &
-          sand = (/0.0,1.0/),                 &
-          sfc = (/0.1,0.5/),                  &
-          silt = (/0.0,1.0/),                 &
-          ssat = (/0.35,0.5/),                &
-          sucs = (/-0.8,-0.03/),              &
-          swilt = (/0.05,0.4/),               &
-          froot = (/0.0,1.0/),                &
-          zse = (/0.0,5.0/),                  &
-          canst1 = (/0.05,0.15/),             &
-          dleaf = (/0.005,0.4/),              &
-          ejmax = (/1.0E-5,3.0E-4/),          &
-          frac4 = (/0.0,1.0/),                &
-          hc = (/0.0,100.0/),                 &
-          lai = (/0.0,8.0/),                  &
-          rp20 = (/0.0,10.0/),                &
-          vbeta =(/-999999.0,999999.0/),      &
-          g0 = (/-0.5,0.5/),                  & ! Ticket #56 (must find better range)
-          g1 = (/0.0,20.0/),                  & ! Ticket #56 (must find better range)
-          xalbnir = (/0.0,1.5/),              &
-          meth = (/0.0,1.0/),                 &
-          za =(/0.0,150.0/),                  &
-          rpcoef = (/0.05,1.5/),              &
-          shelrb = (/1.0,3.0/),               &
-          vcmax = (/5.0E-6,1.5E-4/),          &
-          xfang = (/-1.0,0.5/),               &
-          ratecp = (/0.01,3.0/),              &
-          ratecs = (/0.01,3.0/),              &
-          refsbare = (/0.0,0.5/),             &
-          taul = (/0.0,0.3/),                 &
-          refl = (/0.0,0.5/),                 &
-          tauw = (/0.0,0.1/),                 &
-          refw = (/0.0,0.5/),                 &
-          extkn = (/0.0,10.0/),               & ! YP oct07
-          wai = (/0.0,5.0/),                  & ! YP oct07
-          vegcf = (/0.0,100.0/),              & ! YP oct07
-          tminvj = (/-20.0,15.0/),            &
-          tmaxvj = (/-15.0,30.0/),            &
-          rootbeta = (/0.7,1.0/),             & ! YP oct07
-          veg_class = (/1.0,20.0/),           &
-          soil_class = (/1.0,20.0/)  , &
-          TotLivBiomass =  (/0.0, 1000./),      &
-          TotSoilCarb =  (/0.0, 1000./),      &
-          TotLittCarb =  (/0.0, 1000./), &
-          Area = (/0.0, 5000./),&
-                                !MD
-          WatTable = (/0.0,1.0e10/),          &
-          GWwb = (/0.0,1.0/),              &
-          SatFrac = (/0.0,1.0/),              &
-          Qrecharge = (/-9999.0,9999.0/)
+    REAL, DIMENSION(2) :: &
+      nav_lon = [-360.0, 360.0], &
+      nav_lat = [-90.0, 90.0], &
+      time, &
+      timestp, &
+      ! possible forcing variables for CABLE
+      SWdown = [0.0, 1360.0], & ! W/m^2
+      LWdown = [0.0, 950.0], & ! W/m^2
+      Rainf = [0.0, 0.1], & ! mm/s
+      Snowf = [0.0, 0.1], & ! mm/s
+      PSurf = [500.0, 1100.0], & ! mbar/hPa
+      Tair = [200.0, 333.0], & ! K
+      Qair = [0.0, 0.1], & ! g/g
+      Tscrn = [-70.0, 70.0], & ! oC - INH
+      Qscrn = [0.0, 0.1], & ! kg/kg
+      CO2air = [160.0, 2000.0], & ! ppmv
+      Wind = [0.0, 75.0], & ! m/s
+      Wind_N = [-75.0, 75.0], & ! m/s
+      Wind_E = [-75.0, 75.0], & ! m/s
+      ! possible output variables
+      Qmom = [-10.0, 8000.0], & ! kg/m/s2 - (INH generous range)
+      Qh = [-2000.0, 2000.0], & ! W/m^2
+      Qle = [-2500.0, 2500.0], & ! W/m^2
+      Qg = [-4000.0, 4000.0], & ! W/m^2
+      SWnet = [0.0, 1350.0], & ! W/m^2 (YP oct07)
+      ! SWnet = [0.0, 1250.0],            & ! W/m^2
+      LWnet = [-500.0, 510.0], & ! W/m^2
+      Rnet = [-500.0, 1250.0], & ! W/m^2
+      Evap = [-0.0045, 0.0045], &  ! note this is also used for snow melt !
+      Ewater = [-0.0005, 0.0005], &
+      ESoil = [-0.0015, 0.0015], &
+      TVeg = [-0.0003, 0.0003], &
+      ECanop = [-0.0003, 0.0003], &
+      PotEvap = [-0.0006, 0.0006], &
+      ACond = [0.0, 1.0], &
+      SoilWet = [-0.4, 1.2], &
+      Albedo = [0.0, 1.0], &
+      visAlbedo = [0.0, 1.0], & ! vars intro for Ticket #27
+      nirAlbedo = [0.0, 1.0], & ! vars intro for Ticket #27
+      VegT = [213.0, 353.0], &
+      SoilTemp = [213.0, 353.0], &
+      SoilMoist = [0.0, 2000.0], &
+      Qs = [0.0, 15.0], &
+      Qsb = [0.0, 15.0], &
+      DelSoilMoist = [-2000.0, 2000.0], &
+      DelSWE = [-2000.0, 2000.0], &
+      DelIntercept = [-100.0, 100.0], &
+      SnowT = [213.0, 280.0], &
+      BaresoilT = [213.0, 343.0], &
+      AvgSurfT = [213.0, 333.0], &
+      RadT = [200.0, 373.0], &
+      ! vh_js !
+      SWE = [0.0, 4000.0], &
+      RootMoist = [0.0, 2000.0], &
+      CanopInt = [0.0, 100.0], &
+      NEE = [-70.0, 50.0], & ! umol/m2/s
+      NPP = [-20.0, 75.0], & ! umol/m2/s
+      GPP = [-20.0, 100.0], & ! umol/m2/s
+      PAR = [-1000.0, 5000.0], & ! umol/m2/s
+      AutoResp = [-50.0, 20.0], & ! umol/m2/s
+      LeafResp = [-50.0, 20.0], & ! umol/m2/s
+      HeteroResp = [-50.0, 20.0], & ! umol/m2/s
+      HSoil = [-1000.0, 1000.0], &
+      HVeg = [-1000.0, 1000.0], &
+      SnowDepth = [0.0, 50.0], & ! EK nov07
+      SnowMelt = [-99999.0, 999999.0], &
+      Wbal = [-999999.0, 999999.0], &
+      Ebal = [-999999.0, 999999.0], &
+      ! vh_js !
+      CanT = [213.0, 333.0], &
+      Fwsoil = [0.0, 1.0], &
+      ! parameters:
+      albsoil = [0.0, 0.9], &
+      isoil = [1.0, 30.0], &
+      iveg = [1.0, 30.0], &
+      bch = [2.0, 15.0], &
+      latitude = [-90.0, 90.0], &
+      c3 = [0.0, 1.0], & ! EK nov07
+      clay = [0.0, 1.0], &
+      css = [700.0, 2200.0], &
+      rhosoil = [300.0, 3000.0], &
+      hyds = [5.0E-7, 8.5E-3], & ! vh_js ! sep14
+      rs20 = [0.0, 10.0], &
+      sand = [0.0, 1.0], &
+      sfc = [0.1, 0.5], &
+      silt = [0.0, 1.0], &
+      ssat = [0.35, 0.5], &
+      sucs = [-0.8, -0.03], &
+      swilt = [0.05, 0.4], &
+      froot = [0.0, 1.0], &
+      zse = [0.0, 5.0], &
+      canst1 = [0.05, 0.15], &
+      dleaf = [0.005, 0.4], &
+      ejmax = [1.0E-5, 3.0E-4], &
+      frac4 = [0.0, 1.0], &
+      hc = [0.0, 100.0], &
+      lai = [0.0, 8.0], &
+      rp20 = [0.0, 10.0], &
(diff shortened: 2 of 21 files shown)


src/offline/cable_iovars.F90
```diff
@@ -413,5 +421,136 @@ MODULE cable_IO_vars_module
 
   ! For threading:
   !$OMP THREADPRIVATE(landpt,patch)
+   CONTAINS
+      SUBROUTINE set_group_output_values
```
ccarouge, line 425: Let's add some documentation.

```suggestion
      SUBROUTINE set_group_output_values

      !*#Purpose: 
      ! Set individual variables to output according to the values of the group options from the namelist entries in `output%`.
```

src/offline/cable_driver.F90
```diff
@@ -337,6 +338,8 @@ PROGRAM cable_offline_driver
   ENDIF
 
   ! INITIALISATION depending on nml settings
+  CALL set_group_output_values()
```
ccarouge, line 341:
```suggestion
  ! INITIALISATION depending on nml settings
  ! Initialise flags to output individual variables according to group
  ! options from the namelist file
  CALL set_group_output_values()
```

src/offline/cable_abort.F90
```diff
@@ -140,58 +138,56 @@ END SUBROUTINE nc_abort
...
+      xx, & ! coordinates of erroneous grid square
+      yy    ! coordinates of erroneous grid square
 
+    TYPE(met_type), INTENT(IN) :: met  ! met data
 
-    TYPE(met_type),INTENT(IN) :: met  ! met data
+    REAL, INTENT(IN) :: value ! value deemed to be out of range
 
-    REAL,INTENT(IN) :: value ! value deemed to be out of range
+    REAL, DIMENSION(2), INTENT(IN) :: var_range ! appropriate var range
```
ccarouge, line 162:
```suggestion
    REAL, INTENT(IN) :: var_range(2) ! appropriate var range
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+  END INTERFACE check_range
+
 CONTAINS
 
+  ! Make for integers as well: Example isoilm
+  SUBROUTINE check_range_i1_d1(vname, parameter_i1, parameter_range, ktau, met)
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, DIMENSION(:), INTENT(IN) :: parameter_i1
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
```
ccarouge, line 229:
```suggestion
     INTEGER, INTENT(IN) :: parameter_i1(:)
     REAL, INTENT(IN) :: parameter_range(2)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+  END SUBROUTINE check_range_i1_d1
+
+  SUBROUTINE check_range_r2_d1(vname, parameter_r2, parameter_range, ktau, met)
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     REAL(r_2), DIMENSION(:), INTENT(IN) :: parameter_r2
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
```
ccarouge, line 252:
```suggestion
     REAL(r_2), INTENT(IN) :: parameter_r2(:)
     REAL, INTENT(IN) :: parameter_range(2)
```

src/offline/cable_output.F90
```diff
@@ -725,66 +747,66 @@ SUBROUTINE open_output_file(dels, soil, veg, bgc, rough)
...
             'dummy', xID, yID, zID, landID, patchID, tID)
        ALLOCATE(out%Wbal(mp))
        out%Wbal = 0.0 ! initialise
     END IF
     ! Define carbon variables in output file and allocate temp output vars:
-    IF(output%carbon .OR. output%AutoResp) THEN
+    IF(output%AutoResp) THEN
        CALL define_ovar(ncid_out, ovid%AutoResp, 'AutoResp', 'umol/m^2/s',     &
             'Autotrophic respiration', patchout%AutoResp,          &
             'dummy', xID, yID, zID, landID, patchID, tID)
```
abhaasgoyal, line 812: Configuration : `AU-Tum_2002-2017_OzFlux_Met_R0_S0_out.nc`

Change: L816 uses the default configuration of range checks to include `StemResp` and `LeafResp` in nc headers if `output%carbon` is `true` (since `AutoResp` is being set in `set_group_output_values`). So the effective change is that L816 and L824 are using the condition `IF(output%casa .OR. output%carbon .OR. output%AutoResp)`. Based on the behaviour by other functions, I think this is the supposed functionality that we want

ccarouge:
@abhaasgoyal I'm not sure what happened but for some reason, some comments got posted before I finished the review.. I'm still working on it but I'm getting to the end.

src/offline/cable_checks.F90
ccarouge: This came in before I was finished. I changed this because the UM/JULES coding standard asks not to use the DIMENSION keyword.

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+  END SUBROUTINE check_range_r2_d1
+
+  SUBROUTINE check_range_r1_d1(vname, parameter_r1, parameter_range, ktau, met)
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     REAL, DIMENSION(:), INTENT(IN) :: parameter_r1
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
```
ccarouge, line 272:
```suggestion
     REAL, INTENT(IN) :: parameter_r1(:)
     REAL, INTENT(IN) :: parameter_range(2)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     INTEGER, DIMENSION(:,:), INTENT(IN) :: parameter_i2
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
+
+     INTEGER :: max_val, min_val
+     INTEGER, DIMENSION(2) :: index
```
ccarouge, line 296:
```suggestion
     INTEGER, INTENT(IN) :: parameter_i2(:,:)
     REAL, INTENT(IN) :: parameter_range(2)

     INTEGER :: max_val, min_val
     INTEGER :: index(2)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     REAL(r_2), DIMENSION(:,:), INTENT(IN) :: parameter_r2
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
+
+     REAL(r_2) :: max_val, min_val
+     INTEGER, DIMENSION(2) :: index
```
ccarouge, line 327:
```suggestion
     REAL(r_2), INTENT(IN) :: parameter_r2(:,:)
     REAL, INTENT(IN) :: parameter_range(2)

     REAL(r_2) :: max_val, min_val
     INTEGER :: index(2)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     REAL, DIMENSION(:,:), INTENT(IN) :: parameter_r1
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
+
+     REAL :: max_val, min_val
+     INTEGER, DIMENSION(2) :: index
```
ccarouge, line 358:
```suggestion
     REAL, INTENT(IN) :: parameter_r1(:,:)
     REAL, INTENT(IN) :: parameter_range(2)

     REAL :: max_val, min_val
     INTEGER :: index(2)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     REAL(r_2), DIMENSION(:,:,:), INTENT(IN) :: parameter_r2
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
+
+     INTEGER, DIMENSION(3) :: index
+     REAL(r_2) :: max_val, min_val
```
ccarouge, line 390:
```suggestion
     REAL(r_2), INTENT(IN) :: parameter_r2(:,:,:)
     REAL, INTENT(IN) :: parameter_range(2)

     INTEGER :: index(3)
     REAL(r_2) :: max_val, min_val
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+  SUBROUTINE check_range_r1_d3(vname, parameter_r1, parameter_range, ktau, met)
+
+     CHARACTER(LEN=*) :: vname
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     REAL, DIMENSION(:,:,:), INTENT(IN) :: parameter_r1
+     REAL, DIMENSION(2), INTENT(IN) :: parameter_range
+     INTEGER, DIMENSION(3) :: index
```
ccarouge, line 419:
```suggestion
     REAL, INTENT(IN) :: parameter_r1(:,:,:)
     REAL, INTENT(IN) :: parameter_range(2)
     INTEGER :: index(3)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     END IF
+
+     IF (SIZE(index) > 0) THEN
+          CALL range_abort(vname, ktau, met, parameter_r1(index(1), index(2), index(3)), parameter_range, & 
+               index(1), patch(index(1))%latitude, patch(index(1))%longitude)
+     END IF
+
+  END SUBROUTINE check_range_r1_d3
+
+  ! TODO add for integer
```
ccarouge, line 441: Why add for integer? Is that currently needed by the code? If not, I don't see the point of adding it now. It can be added if it is ever needed.

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+   ! In future, it's better to have a callback for a derived type having both ranges and values
+
+     TYPE (soil_parameter_type), INTENT(IN) :: soil
+     TYPE (veg_parameter_type), INTENT(IN) :: veg
+
+     INTEGER, INTENT(IN) :: ktau
+     TYPE(met_type), INTENT(IN) :: met
+
+     ! Soil
+     !! 1D
```
ccarouge, line 457: `!!` is part of FORD markup and signals a comment line that gets picked up for documentation. So replacing with `~`.

```suggestion
     !~ 1D
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+     !!! Integer
```
ccarouge, line 459:
```suggestion
     !~~ Integer
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     ! CALL  check_range(soil%isoilm, ranges%isoilm) ! Review: isoilm vs isoil
+     ! CALL  check_range(soil%nhorizons, ranges%nhorizons)
+
+     !!! r1
```
ccarouge, line 463:
```suggestion
     !~~ r1
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     ! CALL  check_range(soil%GWz, ranges%GWz, ktau, met)
+     ! CALL  check_range(soil%GWdz, ranges%GWdz, ktau, met)
+     ! CALL  check_range(soil%GWrhosoil_vec, ranges%GWrhosoil_vec, ktau, met)
+
+     ! CALL  check_range(soil%clitt, ranges%clitt, ktau, met)
+     ! CALL  check_range(soil%zeta, ranges%zeta, ktau, met)
+     ! CALL  check_range(soil%fsatma, ranges%fsatma, ktau, met)
+
+     !! 2D
+     !!! Integer
```
ccarouge, line 512:
```suggestion
     !~ 2D
     !~~ Integer
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     ! CALL  check_range(soil%ishorizon, ranges%ishorizon, ktau, met)
+
+     !!! Real
```
ccarouge, line 515:
```suggestion
     !~~ Real
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     !!!! CALL  check_range(soil%bch_vec, ranges%bch, ktau, met)
+     !!!! CALL  check_range(soil%clay_vec, ranges%clay, ktau, met)
+     !!!! CALL  check_range(soil%sand_vec, ranges%sand, ktau, met)
+     !!!! CALL  check_range(soil%silt_vec, ranges%silt, ktau, met)
+     ! CALL  check_range(soil%org_vec, ranges%org, ktau, met)
+     !!!! CALL  check_range(soil%rhosoil_vec, ranges%rhosoil, ktau, met)
+     !!!! CALL  check_range(soil%ssat_vec, ranges%ssat, ktau, met)
+     ! CALL  check_range(soil%watr, ranges%watr, ktau, met)
+     !!!! CALL  check_range(soil%sfc_vec, ranges%sfc, ktau, met)
+     !!!! CALL  check_range(soil%swilt_vec, ranges%swilt, ktau, met)
```
ccarouge, line 537:
```suggestion
     !~~ Real r2
     !~~~ REVIEW: soil%field_vec vs ranges%field
     !~~~ without vec failing on 1x1 - zse and css
     !~~~ CALL  check_range(soil%zse_vec, ranges%zse_vec, ktau, met)
     !~~~ CALL  check_range(soil%css_vec, ranges%css_vec, ktau, met)
     !~~~ CALL  check_range(soil%cnsd_vec, ranges%cnsd, ktau, met)

     !~~~ CALL  check_range(soil%sucs_vec, ranges%sucs, ktau, met)
     !~~~ CALL  check_range(soil%hyds_vec, ranges%hyds, ktau, met)
     !~~~ CALL  check_range(soil%bch_vec, ranges%bch, ktau, met)
     !~~~ CALL  check_range(soil%clay_vec, ranges%clay, ktau, met)
     !~~~ CALL  check_range(soil%sand_vec, ranges%sand, ktau, met)
     !~~~ CALL  check_range(soil%silt_vec, ranges%silt, ktau, met)
     ! CALL  check_range(soil%org_vec, ranges%org, ktau, met)
     !~~~ CALL  check_range(soil%rhosoil_vec, ranges%rhosoil, ktau, met)
     !~~~ CALL  check_range(soil%ssat_vec, ranges%ssat, ktau, met)
     ! CALL  check_range(soil%watr, ranges%watr, ktau, met)
     !~~~ CALL  check_range(soil%sfc_vec, ranges%sfc, ktau, met)
     !~~~ CALL  check_range(soil%swilt_vec, ranges%swilt, ktau, met)
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+
+
+     ! Vegetation
+     !! 1D
+     !!! Integer
```
ccarouge, line 542:
```suggestion
     !~ 1D
     !~~ Integer
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     CALL  check_range("iveg", veg%iveg, ranges%iveg, ktau, met)
+     ! CALL  check_range(veg%iLU, ranges%iLU, ktau, met)
+
+     !!! Real
```
ccarouge, line 546:
```suggestion
     !~~ Real
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     ! CALL  check_range("cfrd", veg%cfrd, ranges%cfrd, ktau, met)
+     ! CALL  check_range("gswmin", veg%gswmin, ranges%gswmin, ktau, met)
+     ! CALL  check_range("conkc0", veg%conkc0, ranges%conkc0, ktau, met)
+     ! CALL  check_range("conko0", veg%conko0, ranges%conko0, ktau, met)
+     ! CALL  check_range("ekc", veg%ekc, ranges%ekc, ktau, met)
+     ! CALL  check_range("eko", veg%eko, ranges%eko, ktau, met)
+     CALL  check_range("g0", veg%g0, ranges%g0, ktau, met)
+     CALL  check_range("g1", veg%g1, ranges%g1, ktau, met)
+
+     !!! Real r_2
```
ccarouge, line 583:
```suggestion
     !~~ Real r_2
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     ! CALL  check_range("clit", veg%clit, ranges%clit, ktau, met)
+
+     !! 2D
+
+     !!! logical,	public,	DIMENSION(:, ktau, met), POINTER	::	deciduous		! Review: No need of this since T/F	
+
+     !!! Integer
+     ! CALL  check_range("disturbance_interval", veg%disturbance_interval, ranges%disturbance_interval, ktau, met)
+
+     !!! Real
```
ccarouge, line 598:
```suggestion
     !~ 2D

     !~~ logical,	public,	DIMENSION(:, ktau, met), POINTER	::	deciduous		! Review: No need of this since T/F	

     !~~ Integer
     ! CALL  check_range("disturbance_interval", veg%disturbance_interval, ranges%disturbance_interval, ktau, met)

     !~~ Real
```

src/offline/cable_checks.F90
```diff
@@ -50,141 +52,561 @@ MODULE cable_checks_module
...
+     CALL  check_range("refl", veg%refl, ranges%refl, ktau, met)
+     CALL  check_range("taul", veg%taul, ranges%taul, ktau, met)
+     CALL  check_range("froot", veg%froot, ranges%froot, ktau, met)
+
+     !!! Real r_2
```
ccarouge, line 603:
```suggestion
     !~~ Real r_2
```

src/offline/cable_checks.F90
ccarouge, line 1: I don't understand why we have so many check_range procedures. As far as I can tell, we only use them with real arrays for the variables to check. So we only ever use the r1 versions.

ccarouge (changes requested):
I was trying to finish this review but I have a relatively important question about `cable_checks.F90` that I think we need to resolve before I continue with the review.

src/offline/cable_checks.F90
ccarouge: @abhaasgoyal That's the important question I want us to resolve before going further.

src/offline/cable_output.F90
```diff
@@ -40,15 +40,15 @@ MODULE cable_output_module
...
   USE cable_write_module
   USE netcdf
   USE cable_common_module, ONLY: filename, calcsoilalbedo, CurYear,IS_LEAPYEAR, cable_user,&
        gw_params
   IMPLICIT NONE
   PRIVATE
-  PUBLIC open_output_file, write_output, close_output_file, create_restart
-  INTEGER :: ncid_out ! output data netcdf file ID
+  PUBLIC open_output_file, write_output, close_output_file, create_restart, check_and_write
+  INTEGER, SAVE :: ncid_out, ncid_restart ! output/restart data netcdf file ID
```
ccarouge, line 51: Why add the SAVE keyword? All the entries in the MODULE specification will be saved by default.

src/offline/cable_output.F90
```diff
@@ -233,16 +233,39 @@ MODULE cable_output_module
   TYPE(output_temporary_type), SAVE :: out
   INTEGER :: ok   ! netcdf error status
 
+
+  ! Some Additional internal variables (TODO: Check their position)
+  LOGICAL :: writenow ! write to output file during this time step?
+  INTEGER, SAVE :: out_timestep ! counter for output time steps
```
ccarouge, line 239: Same comment about the SAVE keyword

src/offline/cable_output.F90
```diff
@@ -1324,142 +1346,137 @@ SUBROUTINE open_output_file(dels, soil, veg, bgc, rough)
...
-         'clay', REAL(soil%clay, 4), ranges%clay, patchout%clay, 'real')
-    IF(output%params .OR. output%sand) CALL write_ovar(ncid_out, opid%sand,    &
-         'sand', REAL(soil%sand, 4), ranges%sand, patchout%sand, 'real')
-    IF(output%params .OR. output%silt) CALL write_ovar(ncid_out, opid%silt,    &
-         'silt', REAL(soil%silt, 4), ranges%silt, patchout%silt, 'real')
-    IF(output%params .OR. output%css) CALL write_ovar(ncid_out, opid%css,      &
-         'css', REAL(soil%css, 4), ranges%css, patchout%css, 'real')
-    IF(output%params .OR. output%rhosoil) CALL write_ovar(ncid_out,            &
+    CALL   check_and_write(output%iveg, opid%iveg,    &
+         'iveg', REAL(veg%iveg, 4), ranges%iveg, patchout%iveg, 'integer', met)
```
ccarouge, line 1350: All these parameters, starting here and below, do not need to be checked here since they are constant in the run and they have been checked with constant_check. But I guess it is easier for future development to update everything to `check_and_write()`.

src/offline/cable_checks.F90
abhaasgoyal: Oh yes, that seems right (since we are converting values to `REAL(4)` in `cable_output`. However, initially I made this for the range checks done in function `constant_check_range (cable_checks.F90)` which did not convert the parameter to another type - there was a case in L585 (`rootbeta`) for example where `r_2` is being used. I can remove the instances for types where no calls are being made.

Another thing - I also see we are checking many of the same constant variables in `open_output_file (cable_output.F90)` as `constant_check_range`, but some variables are there which are only checked in `constant_check_range`. For now, I think it is best to leave as it is.

src/offline/cable_output.F90
```diff
@@ -1633,1266 +1651,526 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
-          ! Divide accumulated variable by number of accumulated time steps:
-          out%PlantCarbLeaf = out%PlantCarbLeaf / REAL(output%interval, 4)
-          ! Write value to file:
-          CALL write_ovar(out_timestep, ncid_out, ovid%PlantCarbLeaf, 'PlantCarbLeaf', out%PlantCarbLeaf,    &
-               ranges%TotLittCarb, patchout%PlantCarbLeaf, 'default', met)
-          ! Reset temporary output variable:
-          out%PlantCarbLeaf = 0.0
-       END IF
+  SUBROUTINE generate_out_write_acc_d1(output_var, varID, vname, out_var, acc_val, vrange, writepatch, met)
+     !! TODO: Remove met
```
ccarouge, line 2143: Why do you want to remove met? Don't you need it in `check_range`? Do you want to move to get it through a USE statement instead of an input arg?

src/offline/cable_output.F90
```diff
@@ -1633,1266 +1651,526 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
-               ranges%NEE, patchout%PlantTurnoverWoodCrowding, 'default', met)
-          ! Reset temporary output variable:
-          out%PlantTurnoverWoodCrowding = 0.0
-       END IF
+  END SUBROUTINE write_output
 
+  SUBROUTINE acc_out_var_d1(output_var, out_var, acc_val)
+     LOGICAL, INTENT(IN) :: output_var ! Whether to write the current variable
+     REAL(4), DIMENSION(:) :: out_var
+     REAL(4), DIMENSION(:) :: acc_val
```
ccarouge, line 2004: Could you update all these new routines to remove the DIMENSION keyword as below:

```suggestion
     REAL(4) :: out_var(:)
     REAL(4) :: acc_val(:)
```

src/offline/cable_output.F90
```diff
@@ -1633,1266 +1651,526 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
+          ! Write value to file:
+          CALL write_ovar(out_timestep, ncid_out, varID, vname, &
+               out_var, writepatch, dimswitch, met)
+          ! Reset temporary output variable:
+          out_var = 0.0
+     END IF
+  END SUBROUTINE check_and_write_d2
+
+  SUBROUTINE check_and_write_d1_p(output_par, parID, pname, out_par, prange, writepatch, dimswitch, met, restart)
+     ! TODO: Remove dimswitch
```
ccarouge, line 2083: Can you remove dimswitch instead of leaving the TODO?

ccarouge (changes requested):
I think I agree with the design of the changes in the rest of the code but I have a few comments.

We'll also need to look at the documentation.

src/offline/cable_checks.F90
ccarouge: > Another thing - I also see we are checking many of the same constant variables in open_output_file (cable_output.F90) as constant_check_range, but some variables are there which are only checked in constant_check_range. For now, I think it is best to leave as it is.

Yes leave as is. When the code is clearer, we can think of improvements here.

> I can remove the instances for types where no calls are being made.

Why not have a `REAL(4)` conversion for the parameters as well in `constant_check_range`? The range for these is also a REAL(4) so using a higher or lower precision variable in the comparison won't give us anything.

abhaasgoyal: @ccarouge Makes sense, I'll do this

ccarouge:
After reviewing this, I think we should have split this work into 3:
- check the parameters at the start (constant_check_range)
- redesign cable_output.F90 with introducing `generate_out_write_acc` and `acc_out_var`
- change the range checks for the output.

But I guess it's easier to see with the final version than at the start.

I'm just mentioning it as something to try and learn from for later work.

src/offline/cable_output.F90
abhaasgoyal: Yup, it makes sense but this also depends on when `open_output_file` is being called (do we want this behaviour). But yes, we can see this in the future.

src/offline/cable_output.F90
abhaasgoyal: I was trying to minimize the list of parameters such as for `met`, `dimswitch`, `restart`. There are some aspects to this

- Another file landuse is calling `check_and_write` 
- We don't want circular dependencies w.r.t. import statements (Like `USE cable_driver_module` wouldn't be good since 
- State management of local variables could be tricky, so a functional style would be good here 

So I'll create a type for `output_settings` and add it as a parameter (I was thinking of doing that hence the TODO)

src/offline/cable_output.F90
abhaasgoyal: I also got confused by `SAVE` in the `out` variable, should I remove the attribute there as well?

src/offline/cable_output.F90
abhaasgoyal: Did for new routines as well for some old routines, but needs to be changed in many places. We can look into this in the future.

src/offline/cable_output.F90
ccarouge: For a link about the SAVE statement: https://community.intel.com/t5/Intel-Fortran-Compiler/Placement-of-SAVE-statement-within-a-module/td-p/1154139

I'm not sure what SAVE statement you are talking about but I wouldn't use SAVE in any module specification area since the standards now say it isn't required.

src/offline/cable_output.F90
ccarouge: I don't mind yet for the old stuff, just trying to make sure new code we introduce follows the standards we will want to enforce.

src/offline/cable_output.F90
```diff
@@ -251,6 +262,11 @@ MODULE cable_output_module
       MODULE PROCEDURE :: check_and_write_d2_p
   END INTERFACE check_and_write
   
+  INTERFACE reset_on_write
+      MODULE PROCEDURE :: reset_on_write_d1
+      MODULE PROCEDURE :: reset_on_write_d2
+
+  END INTERFACE reset_on_write
```
ccarouge, line 269:
```suggestion
  INTERFACE reset_on_write
      MODULE PROCEDURE :: reset_on_write_d1
      MODULE PROCEDURE :: reset_on_write_d2
  END INTERFACE reset_on_write
  
```

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
+    REAL(4), INTENT(IN) :: out_var(:)
+    LOGICAL, INTENT(IN) :: writenow
+
+    REAL(4) :: res(SIZE(out_var))
+
+    IF (output_var .AND. writenow) THEN
+      res = 0.0
+    ELSE
+      res = out_var
+    END IF
```
ccarouge, line 2092: These are easier to read like this:

```suggestion
    res = out_var

    ! Reset the value if it has been written to file
    IF (output_var .AND. writenow) THEN
      res = 0.0
    END IF
```

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
+    REAL(4), INTENT(IN) :: out_var(:, :)
+    LOGICAL, INTENT(IN) :: writenow
+
+    REAL(4) :: res(SIZE(out_var, DIM=1), SIZE(out_var, DIM=2))
+
+    IF (output_var .AND. writenow) THEN
+      res = 0.0
+    ELSE
+      res = out_var
+    END IF
```
ccarouge, line 2107:
```suggestion
    res = out_var
    
    ! Reset the value if it has been written to file
    IF (output_var .AND. writenow) THEN
      res = 0.0
    END IF
```

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
 
+    IF (output_var) THEN
+      ! Accumulate out_var until interval timesteps
+      res = out_var + acc_val
+      IF (writenow) THEN
+        res = res/REAL(output%interval, 4)
+      END IF
+    ELSE
+      res = out_var
+    END IF
```
ccarouge, line 2058:
```suggestion
    res = out_var

    IF (output_var) THEN
    ! Accumulate out_var until interval timesteps
      res = out_var + acc_val
      IF (writenow) THEN
        res = res/REAL(output%interval, 4)
      END IF
    END IF
```

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
+
```
ccarouge, line 2078: Simpler to read.

```suggestion
    res = out_var

    IF (output_var) THEN
      ! Accumulate out_var until interval timesteps
      res = out_var + acc_val
      IF (writenow) THEN
        res = res/REAL(output%interval, 4)
      END IF
    END IF

```

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
-     IF (output_var) THEN
-          ! Add current timestep's value to total of temporary output variable:
-          out_var = out_var + acc_val
-          IF (writenow) THEN
-               ! Divide accumulated variable by number of accumulated time steps:
-               out_var = out_var / REAL(output%interval, 4)
-          END IF
-     END IF
-  END SUBROUTINE acc_out_var_d1
+  PURE FUNCTION acc_out_var_d1(output_var, out_var, acc_val, writenow) RESULT(res)
```
ccarouge, line 2042: Actually, this function could be written and defined as an ELEMENTAL function. In which case, it can be applied to an array of any size. See: https://fortranwiki.org/fortran/show/elemental

This removes the need to have 2 functions and an interface.

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
+      IF (writenow) THEN
+        res = res/REAL(output%interval, 4)
+      END IF
+    ELSE
+      res = out_var
+    END IF
+
+  END FUNCTION acc_out_var_d2
+
+  PURE FUNCTION reset_on_write_d1(output_var, out_var, writenow) RESULT(res)
```
ccarouge, line 2081: Same as for `acc_out_var`, this could be simplified with an ELEMENTAL function.

src/offline/cable_output.F90
```diff
@@ -1935,239 +1979,260 @@ SUBROUTINE write_output(dels, ktau, met, canopy, casaflux, casapool, casamet, ss
...
+    INTEGER, INTENT(IN) :: varID ! variable's netcdf ID
+    CHARACTER(LEN=*), INTENT(IN) :: vname ! name of variable
+    REAL(4), INTENT(IN) :: out_var(:, :)
+    REAL(4), INTENT(IN) :: acc_val(:, :)
+    REAL, INTENT(IN) :: vrange(2) ! max and min for variable
+    LOGICAL, INTENT(IN) :: writepatch ! write patch-specific info for this var?
+    TYPE(output_var_settings_type), INTENT(IN) :: out_settings  ! output specific settings
+
+    IF ((check%ranges .EQ. ON_TIMESTEP) .OR. (out_settings%writenow .AND. (check%ranges .EQ. ON_WRITE))) THEN
+      CALL check_range(vname, acc_val, vrange, out_timestep, out_settings%met)
```
ccarouge, line 2143: Since we are calling `check_range` from a function that has various forms depending on the array size, we could simplify `check_range()` to be an elemental function and call `range_check()` from here.

But I think it complicates `contant_check_range()` so it's probably not worth it at this stage. It might make more sense once we reorganise the data structures.

ccarouge (changes requested):
That's great. I think we getting there. I just had an idea with the elemental functions which I think is worth putting in. The rest is trivial.

src/offline/cable_output.F90
abhaasgoyal: How would it complicate `constant_check_range`?

We would also change a lot in `write_ovar` as well, which we could do in another issue

src/offline/cable_output.F90
abhaasgoyal: I see, initially I thought `elemental` required all arguments to be array, but the requirement is every argument should be scalar, and operations on any 2 arguments must conform to the same shape? (just want to clarify whether this is correct)

ccarouge: Yes, you are correct. It's a bit different for functions and subroutines.

I found this great source of information: https://www.ibm.com/docs/en/xffbg/121.141?topic=procedures-elemental

Here is the excerpt for your question:

> The actual arguments in a reference to an elemental procedure can be either of the following:
> - All scalar. For elemental functions, if the arguments are all scalar, the result is scalar.
> - One or more array-valued. The following rules apply if one or more of the arguments is array-valued:
>    - For elemental functions, the shape of the result is the same as the shape of the array actual argument with the greatest rank. If more than one argument appears then all actual arguments must be conformable.
>    - For elemental subroutines, all actual arguments associated with [INTENT(OUT)](https://www.ibm.com/docs/en/SS2MB5_14.1.0/com.ibm.xlf141.bg.doc/language_ref/intent.html#intent) and INTENT(INOUT) dummy arguments must be arrays of the same shape, and the remaining actual arguments must be conformable with them.

src/offline/cable_output.F90
ccarouge: In constant_check_range, we call check_range for each parameter. If we change to an ELEMENTAL function, then we need 2 lines per parameter in constant_check_range instead of one: 
- call to check_range
- call to range_abort

So I'm not sure it is worth doing it now.

abhaasgoyal: I think `check_range` can call `range_abort` if we make all arguments as scalar - `met_type` and `var_range` are the only ones. I also think we can do this in the future.

abhaasgoyal:
Note: Before merging, I'll squash the commits into one with message and appropriate description

ccarouge (approved):
All good!! Well done.

ccarouge:
@abhaasgoyal actually don't merge! I'm missing the info on what tests have been run after review.

abhaasgoyal:
Tested using custom-build in `benchcab=4.0.2` (comment out `remove_module_lines`)
```
realisations:
  - repo:
      local:
        path: /home/659/ag9761/ACCESS-NRI/CABLE
    build_script: ./build.bash
    patch:
      cable:
        check:
         ranges: 1
  - repo:
      git:
        branch: main
    build_script: /scratch/tm70/ag9761/bench_example/src/main/build.bash

modules: [
  intel-compiler/2021.1.1,
  netcdf/4.7.4,
  openmpi/4.1.0
]
```
The output comparisons differ in defining headers for `RootResp` and `StemResp` (as mentioned in the previous comment so ran the following command to filter out any changes in `runs/fluxsite/analysis/bitwise-comparisons`:
```sh
# Tail outputs file names along with the contents
tail -n +1 * | sed -n '/Failed to find variable RootResp/!p' | sed -n '/Failed to find variable StemResp/!p' | tee temp
```
 There seem to be no differences in some tests like `AU-ASM_2011-2017_OzFlux_Met`, `AU-Ctr_2010-2017_OzFlux_Met`, however some tests like `AU-GWW_2013-2017_OzFlux_Met` (which I hadn't tested before are not being outputted with the changes. The errors seem like the following:

```
IFFER : LENGTHS OF RECORDS : time (0) <> time (87648)
DIFFER : VARIABLE : time : SIZE : 0 <> 87648
DIFFER : VARIABLE : SWdown : SIZE : 0 <> 87648
DIFFER : VARIABLE : LWdown : SIZE : 0 <> 87648
DIFFER : VARIABLE : Tair : SIZE : 0 <> 87648
DIFFER : VARIABLE : Rainf : SIZE : 0 <> 87648
DIFFER : VARIABLE : Snowf : SIZE : 0 <> 87648
DIFFER : VARIABLE : Qair : SIZE : 0 <> 87648
DIFFER : VARIABLE : Wind : SIZE : 0 <> 87648
DIFFER : VARIABLE : PSurf : SIZE : 0 <> 87648
.... (variables in gen_acc_and_write)
```

I'll look more into this and let you know

src/offline/cable_write.F90
```diff
@@ -901,13 +883,6 @@ SUBROUTINE write_output_variable_r1(ktau, ncid, varID, vname, var_r1,        &
                 ! patches):
                 otmp3xyt(land_x(i), land_y(i), 1) = SUM(var_r1(landpt(i)%cstart:   &
                      landpt(i)%cend) * patch(landpt(i)%cstart:landpt(i)%cend)%frac)
```
abhaasgoyal, line 885: @ccarouge does processing like this affect check range bounds for variables like `HSoil`, where it's just not the index anymore

src/offline/cable_output.F90
abhaasgoyal: L794 should be corrected to `output%casa .AND. output%AutoResp` to account for the same output file and the desired behaviour of setting group values

src/offline/cable_write.F90
abhaasgoyal: nvm it doesn't

src/offline/cable_output.F90
```diff
@@ -3453,215 +2754,238 @@ SUBROUTINE create_restart(logn, dels, ktau, soil, veg, ssnow,
...
+!$                       ranges%swilt, patchout_var, out_settings)
     ! Soil dimensioned variables/parameters:
-!$    CALL write_ovar (ncid_restart, rpid%froot, 'froot', REAL(veg%froot, 4),    &
-!$                     ranges%froot, .TRUE., 'soil', .TRUE.)
-    CALL write_ovar (ncid_restart, tggID, 'tgg', REAL(ssnow%tgg, 4),           &
-         ranges%SoilTemp, .TRUE., 'soil', .TRUE.)
-    CALL write_ovar (ncid_restart, wbID, 'wb', ssnow%wb, ranges%SoilMoist,     &
-         .TRUE., 'soil', .TRUE.)
-    CALL write_ovar (ncid_restart, wbiceID, 'wbice', ssnow%wbice,              &
-         ranges%SoilMoist, .TRUE., 'soil', .TRUE.)
```
abhaasgoyal, line 3495: The variables `wb`, `wbice`, and `gammzz` are not being converted to `REAL(4)` so they have precision error in the output after the changes

abhaasgoyal:
Tested with the following configurations:

- Check ranges as `true`
```
realisations:
  - repo:
      local:
        path: /home/659/ag9761/ACCESS-NRI/CABLE
    build_script: ./build.bash
    patch:
      cable:
        check:
         ranges: 1
         exit: true
  - repo:
      git:
        branch: main
    build_script: /scratch/tm70/ag9761/bench_example/src/main/build.bash
    patch:
      cable:
        check:
         ranges: true

modules: [
  intel-compiler/2021.1.1,
  netcdf/4.7.4,
  openmpi/4.1.0
]
```
- Check ranges as `false`
```
realisations:
  - repo:
      local:
        path: /home/659/ag9761/ACCESS-NRI/CABLE
    build_script: ./build.bash
    patch:
      cable:
        check:
         ranges: 1
         exit: false
  - repo:
      git:
        branch: main
    build_script: /scratch/tm70/ag9761/bench_example/src/main/build.bash
    patch:
      cable:
        check:
         ranges: false

modules: [
  intel-compiler/2021.1.1,
  netcdf/4.7.4,
  openmpi/4.1.0
]
```
- Include restart file
```
realisations:
  - repo:
      local:
        path: /home/659/ag9761/ACCESS-NRI/CABLE
    build_script: ./build.bash
    patch:
      cable:
        check:
         ranges: 1
         exit: false
        filename:
         restart_out: './restart.nc'
        output:
         restart: true
  - repo:
      git:
        branch: main
    build_script: /scratch/tm70/ag9761/bench_example/src/main/build.bash
    patch:
      cable:
        check:
         ranges: false
        filename:
         restart_out: './restart.nc'
        output:
         restart: true
```
Then `nccmp -d -f` the respective `restart.nc`

Status
______

As of now, the bitwise comparisons of the output files pass in all configs.
Note: There is slight error in writing restart file for certain variables:

https://github.com/CABLE-LSM/CABLE/pull/287/files#diff-841e0c9e12f43660f431fb6d37758c20eed183be6ba0294b25c88712ca18daf3L3492

src/offline/cable_output.F90
ccarouge: Yes, that's fine.

ccarouge:
@abhaasgoyal Good to go! 🎉

PR #288, opened 2024-05-09
implemented generalised procedure for determining woody fraction.

juergenknauer:
## Description

As discussed in the email, I have restructured and streamlined the cable_LUC_EXPT.F90 routine and the way it handles the woody fraction. Specific changes:
- new soubroutine get_woody_fraction() that calculates the woody fraction of a grid cell (previously called CPC in bios and hard-coded constants in all other cases). This is now integrated into the luc_expt structure as luc_expt%woodfrac
- luc_expt%woodfrac replaces CPC and hard-coded fractions throughout the rest of the code. It is also applied to transitions. All this makes a couple of if/where statements redundant.
- I have NOT yet changed the other changes we discussed (including the ptoc transitions that were removed from the code but ought to be in there). So this version should give identical results to the preivous one. I am testing that right now.

## Question:

- in the previous version, CPC was a real, allocatable. It was allocated and deallocated in the LUC_EXPT_INIT routine. I have changed it to be a real :: cpc(mland) since I thought we know its lenght from the beginning. Would be good if you could check this explicitly.

Happy for any sort of feedback and suggestions of course.

### DIFF
--- a/offline/cable_LUC_EXPT.F90
+++ b/offline/cable_LUC_EXPT.F90
@@ -930,9 +930,6 @@ CONTAINS
        write(*,'(a)') 'warning: past end of LUH2 record'
     endif
 
-    ! Determine woody fraction (forest and shrub cover).
-    call get_woody_fraction(LUC_EXPT)
-
     ! Adjust transition areas based on primary wooded fraction
     LUC_EXPT%INPUT(ptos)%VAL   =  LUC_EXPT%INPUT(ptos)%VAL   * LUC_EXPT%woodfrac
     LUC_EXPT%INPUT(ptog)%VAL   =  LUC_EXPT%INPUT(ptog)%VAL   * LUC_EXPT%woodfrac
--- a/offline/cable_LUC_EXPT.F90
+++ b/offline/cable_LUC_EXPT.F90
@@ -942,6 +942,13 @@ CONTAINS
     LUC_EXPT%INPUT(smharv)%VAL =  LUC_EXPT%INPUT(smharv)%VAL * LUC_EXPT%woodfrac
     LUC_EXPT%INPUT(syharv)%VAL =  LUC_EXPT%INPUT(syharv)%VAL * LUC_EXPT%woodfrac
 
+    LUC_EXPT%INPUT(ptoc)%VAL = LUC_EXPT%INPUT(ptoc)%VAL * LUC_EXPT%woodfrac
+    LUC_EXPT%INPUT(ptoq)%VAL = LUC_EXPT%INPUT(ptoq)%VAL * LUC_EXPT%woodfrac
+    LUC_EXPT%INPUT(stoc)%VAL = LUC_EXPT%INPUT(stoc)%VAL * LUC_EXPT%woodfrac
+    LUC_EXPT%INPUT(stoq)%VAL = LUC_EXPT%INPUT(stoq)%VAL * LUC_EXPT%woodfrac
+    LUC_EXPT%INPUT(ctos)%VAL = LUC_EXPT%INPUT(ctos)%VAL * LUC_EXPT%woodfrac
+    LUC_EXPT%INPUT(qtos)%VAL = LUC_EXPT%INPUT(qtos)%VAL * LUC_EXPT%woodfrac
+
   END SUBROUTINE READ_LUH2
 
 
--- a/offline/cable_LUC_EXPT.F90
+++ b/offline/cable_LUC_EXPT.F90
@@ -299,13 +299,17 @@ CONTAINS
     LUC_EXPT%secdf    = max(1.0-LUC_EXPT%grass-LUC_EXPT%primaryf, 0.0)
     
     !! JK Debug
+    ! For some reason the transition from secdf does not work correctly
+    ! if there is secdf = 0.0 at first timestep (1580). This is a bandaid
+    ! solution that allows a 'correct' transition. Minimum fraction of 0.01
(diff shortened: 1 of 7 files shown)


offline/cable_LUC_EXPT.F90
```diff
@@ -929,38 +930,16 @@ SUBROUTINE READ_LUH2(LUC_EXPT)
        write(*,'(a)') 'warning: past end of LUH2 record'
     endif
 
+    ! Determine woody fraction (forest and shrub cover).
+    call get_woody_fraction(LUC_EXPT)
```
har917, line 934: Do we need to recalculate here - or can we just inherit %woodfrac from the call to get_woody_fraction from within LUC_EXPT_INIT?  

This going to end up rereading the Climate/MVG file every year so will slow things down.

If it does need to redone - then I suggest putting the read of MVG/Climate into LUC_EXPT_INIT and restrict get_woody_fraction() to go from %biome to %woody_fraction.

offline/cable_LUC_EXPT.F90
```diff
@@ -477,40 +370,16 @@ SUBROUTINE LUC_EXPT_INIT(LUC_EXPT)
...
-             LUC_EXPT%grass    = LUC_EXPT%primaryf * 0.2
-             LUC_EXPT%primaryf = LUC_EXPT%primaryf * 0.8
-          ELSEWHERE (LUC_EXPT%biome .eq. 5 .or. LUC_EXPT%biome .eq. 6 ) ! DBL
-             LUC_EXPT%grass    = LUC_EXPT%primaryf * 0.3
-             LUC_EXPT%primaryf = LUC_EXPT%primaryf * 0.7
-          END WHERE
-       END WHERE
-    ENDIF
-
+    ! set secondary vegetation area to be zero where land use transitions don't occur
```
har917, line 373: I suspect that I would reorganise this routine so that similar sections are co-located.  There's a lot of setting then resetting things as you go along which is somewhat confusing.

offline/cable_LUC_EXPT.F90
```diff
@@ -477,40 +370,16 @@ SUBROUTINE LUC_EXPT_INIT(LUC_EXPT)
...
+    ! set grass component of primary vegetation cover
+    WHERE (LUC_EXPT%prim_only .eqv. .TRUE.)
+      LUC_EXPT%secdf    = 0.0
+      LUC_EXPT%primaryf = 1.0
+      LUC_EXPT%grass    = 0.0
```
har917, line 378: I'll throw this comment in - not directly relevant but possibly worth including in the next revision.

I think we need to force %crop and %past to zero here - just to avoid potential inconsistency (yes everyone should have set up the states and transitions files appropriately but some people are just going to use the prim_only input file and think everything gets overwritten.

However - note that the Read Climate_file routine sets prim_only to true or non-woody potential vegetation - so no native grass to crop?

har917 (changes requested):
Overall I think this is nearly there - see comment about whether or not we need to re-evaluate %woodfrac from within READ_LUH2.  Either way I think we need a minor change before committing.

A couple of notes
-  this will not reproduce the current BIOS results - that is okay as this fix resolves a bug with BIOS.
- I'm happy with the switch to CPC(mland) from ALLOCATABLE.  This is appropriate, probably should have been the method originally, and is inline with our general trend to use localised memory.
- There's some weird stuff around what happens when %prim_only is true - either when set through the input file or when IGBP veg class is > 5.  This needs following up.

offline/cable_LUC_EXPT.F90
juergenknauer: Good point, I don't think it needs to be recalculated because LUC_EXPT_INIT is called before all occurrences of READ_LUH2 and because woodfrac does not change over time (maybe it should, but that's a different issue).

I deleted the two highlighted lines.

offline/cable_LUC_EXPT.F90
juergenknauer: It could most likely be improved. But at this point I am afraid of breaking things. Maybe something for later.

offline/cable_LUC_EXPT.F90
juergenknauer: yes, agree we should do that. Will include that in the next commit when i include the transitions to crop and pasture.

Will also double-check the Read_ClimateFile routine for consistency.

juergenknauer:
Thanks for the thorough review. Will follow up the remaining issue regarding the consistency of the LU states in the prim_only state.

juergenknauer:
I have done now a couple of more commits. Some of them relate to pre-processing and aren't super critical. But it would be good if you could have a look at the very last commit. That's the one where I included the within-grass transitions from and to croplands and pastures. 

I have also defined harvest index as parameter and reduced from 0.9 to 0.5 in crops.

I am doing a test run right now. It's running, but I don't know what difference it makes.

har917 (approved):
okay for this to go forward - my remaining concern is whether any of this intersects with the MPI code (in whether the new component LUC_EXPT%woodfrac needs to be distributed/shared across processors).  However a) I can't find instances of the LUC_EXPT% in the MPI sharing code and b) POPLUC is shared and all variables that depend on LUC_EXPT%woodfrac make their way into the POPLUC% TYPE - so I think this okay.

Additional comments in the LUH2 processing script would be helpful (especially around the combination of variables going into the different states and transitions)

juergenknauer:
Thanks for the review! I think it's time to merge it back to the main branch.

Issue #212, opened 2024-02-19
CMIP6 historical with POP on and off


ccarouge:
Run with CMIP6 historical configuration with POP turned on and off.

**Requirements:**
- compare results in ilamb. And with the TRENDY runs? Can we compare simulations on different grids in ilamb?
- how long should we run?

Whyborn:
Some work on the CABLE-POP_TRENDY branch is needed for this. The branch currently contains a number of input and configuration routines that are hard-coded to filenames and hidden configurations e.g. in `CRU_INIT` and `CRU_GET_FILENAME` subroutines in cable_cru_TRENDY.f90. Should generalise these routines and remove "hidden" configuration parameters, shifting them to the externally visible namelist files.

JhanSrbinovsky:
I had to clone the CABLE repo to get the CABLE-POP-TRENDY branch (again), when I noticed this branch/issue and the name sounded like what I was doing, although it isnt.

I started months back on "merging" the CABLE-POP branch into CABLE3 but it has been so long that I can't exactly remember which files come from where. What I am doing first is pushing the POP branch into the same format as the CABLE3 code. I set it up to build/run serial without POP. I dont care about merging the POP part, whatever is in Juergen's branch will totally over-ride whatever we have. So apart from some minor changes to get it to build/run the code is on the way to being in the same "format". I put it down in the middle of things and just to verify that some of the files that I have already moved are exactly Jurgen's, but it has been so long that my copy, probably on SCRATCH/ has disappeared.

ccarouge:
> it has been so long that my copy, probably on SCRATCH/ has disappeared.

Do you mean everything has disappeared or have you pushed your work to GitHub? If it's still on GitHub, it would be good to have a branch name and an issue opened for that work. Since that is not what this issue is about. I'll modify the title to make this clearer.

JhanSrbinovsky:
> > it has been so long that my copy, probably on SCRATCH/ has disappeared.
> 
> Do you mean everything has disappeared or have you pushed your work to GitHub? If it's still on GitHub, it would be good to have a branch name and an issue opened for that work. Since that is not what this issue is about. I'll modify the title to make this clearer.

Nah, all good. I meant just my local copy of Jurgen's branch

ccarouge:
@Whyborn I think this issue is too wide and in essence a duplicate of #207 . I'm changing it's description to only be about reading the met. files. The rest of the work should be done in separate issues.

PR #290, opened 2024-05-14
Refactor of the CABLE-POP Met input routines.

Whyborn:
# CABLE

## Description

This is a work in progress for the new Met forcing routines for the CABLE-POP_TRENDY branch. This branch will involve some fairly major changes to the ```cable_cru_trendy.F90``` module, to remove all hard-coded references to files and remove all internal switches hidden through the use of the ```Run```  namelist variable.

The new input routines allow the easy handling of datasets spanning multiple files. It is currently tested on the Met inputs, and should be able to be ported to other inputs without too much work.

The initial purpose of this branch is to bring the code to a state where we can run with ACCESS-ESM1.5 Met forcings with Land Use Change and Tree Demography science included, addressing #212 .

## Type of change

- [x] New feature

## Checks

- [x] Checked against small-scale TRENDY configuration, with ```extent=64.0,66.0,60.0,62.0``` in ```run_TRENDY.sh```.
- [x] Checked against global configuration.


offline/cable_input.F90
```diff
@@ -2887,5 +2921,402 @@ SUBROUTINE allocate_cable_vars(air,bgc,canopy,met,bal,                         &
...
+
+  CurrentFile(IndxStartDate:IndxStartDate) = "*"
+  CurrentFile(IndxStartDate+1:) = CurrentFile(IndxStartDate+11:)
+
+  ! Now we've processed the supplied string, pass it to the ls unix command.
+  ! We need to write the output to a temporary file and read it back in to
+  ! get the date ranges for each file.
+  CALL execute_command_line("ls -1 "//TRIM(CurrentFile)//" >&
+     __FileNameWithNoClashes__.txt", EXITSTAT = ExStat, CMDSTAT = CStat)
+
```
mcuntz, line 3078: Is there no other way to do this? Calling the OS is not optimal. It will depend which shell the user is in, for example if overwriting an existing file.
You also remove the possibility to compile on Windows (except with the Linux sub-system).

offline/cable_input.F90
```diff
@@ -2887,5 +2921,402 @@ SUBROUTINE allocate_cable_vars(air,bgc,canopy,met,bal,                         &
...
+  ApplyMask: DO LandCell = 1, SIZE(LandIDx)
+    ok = NF90_GET_VAR(STD%CurrentFileID,&
+      STD%CurrentVarID, DataArr(LandCell), START = (/LandIDx(LandCell),&
+      LandIDy(LandCell), TimeIndex/))
+      IF (ok /= NF90_NOERR) THEN
+        CALL handle_err(ok, "Failed reading "//STD%FileNames&
+          (STD%CurrentFileIndx)//" in read_metvals.")
+      END IF
+  END DO ApplyMask
+
```
mcuntz, line 3317: Is there the possibility to read the whole field at once, for example for a global run?
Reading point by point might be very costly if the input data is in netcdf4 and has a chunksize of the global field. Then you read SIZE(LandIDx) times a whole global field.

offline/cable_cru_TRENDY.F90
mcuntz, line 1: I cannot comment on the file. Too much changes.
But did you test it, for example in the environment trendy_v12? Does it give the same results?

mcuntz (changes requested):
I appreciate any generalisation for reading spatial data. I find the approach to call the OS not very portable, especially when people are using different shells. There should be a library in Fortran that deals with the OS such as a the glob module in Python.

This is a pull request so it means that it was tested. Did it gave the same results as the "normal" Trendy S3 run, for example?

offline/cable_input.F90
Whyborn: I was trying to find a better way to do this, but I wasn't able to find anything. I found a few stack overflow threads on the issue, and the consensus was "There's no easy way to do this in Fortran". The examples I found recommended using the method I ended up using.

offline/cable_input.F90
Whyborn: Do you mean the functionality of the old DirectRead option? I didn't notice a big performance hit when I removed it, but I can add it back in.

offline/cable_cru_TRENDY.F90
Whyborn: I tested it using TRENDY_v12 using the "S0" experiment. I am currently bringing the run scripts up to date for the other experiment options, reflecting the changes made to the Met recycling by adding the recycling logicals to the namelist.

Whyborn:
I am working on the TRENDY_v12 configuration that would go alongside these changes, reflecting the new IO routines and the changes to the namelist, see [this pull request](https://github.com/CABLE-LSM/TRENDY_V12/pull/6). This should be done over the next few days, and I will test against the original "S1", "S2", "S3" experiments for bitwise equivalence.

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1433 @@
...
+      CRU%NDepVID)
+    CALL handle_err(ok, "Reading NDep variable")
+  END IF
+END SUBROUTINE CRU_INIT
+
+
+SUBROUTINE CRU_GET_SUBDIURNAL_MET(CRU, MET, CurYear, ktau, kend)
+
+  ! Obtain one day of CRU-NCEP meteorology, subdiurnalise it using a weather
+  ! and return the result to the CABLE driver.
```
ccarouge, line 217:
```suggestion
  ! Obtain one day of CRU-NCEP meteorology, subdiurnalise it using a weather
  ! generator and return the result to the CABLE driver.
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1433 @@
...
+  end if ! End of If newday
+
+  ! Decision has been made, that first tstep of the day is at 0:01 am
+  call WGEN_SUBDIURNAL_MET(WG, CRU%mland, nint(met%hod(1)*3600./dt))
+
+  ! assign to cable variables
+
+  ! strangely, met% is not save over Wait all in MPI...!
+  !  met%pmb = CRU%MET(pres)%METVALS !CLN interpolation??
+  ! Assign weather-generated data, or daily values, as required to CABLE
```
ccarouge, line 354: To align with original file.

```suggestion
  !  met%pmb = CRU%MET(pres)%METVALS !CLN interpolation??
  
  ! Assign weather-generated data, or daily values, as required to CABLE
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1433 @@
...
+    IF (ok /= NF90_NOERR) THEN
+      CALL handle_err(ok, "Reading from NDep")
+    END IF
+  END DO ApplyLandMask
+
+END SUBROUTINE get_cru_ndep
+
+!!------------------------------------------------------------------------------!
+
+!SUBROUTINE cru_read_metvals(STD, MetData, Year, DayOfYear,&
```
ccarouge, line 1301: Why keep this commented out subroutine?

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1433 @@
...
+  smoy = 1
+  syear = CRU%CYEAR
+
+  ! Allocate memory for the average long-wave radiation down for some reason?
+  ALLOCATE(CRU%AVG_LWDN(mLand))
+END SUBROUTINE read_landmask
+
+!------------------------------------------------------------------------------!
+
+!SUBROUTINE prepare_spatiotemporal_dataset(FileTemplate, Dataset)
```
ccarouge, line 924: Why keep this commented out subroutine?

ccarouge (changes requested):
This is going to be a long process. So I've decided to leave several reviews focussing on different aspects. This one is on the documentation. Plus a few things I had put before deciding to split the review in bits.

First major comment. In cable_cru_TRENDY.F90 can you restore the order of the subroutines that existed before the modifications? I think it might make reviewing changes easier. Add the new (not modified, completely new) subroutines to the end if any.

I haven't always checked the comments against the original comments, it's ok if the reply to some suggestions is: Was in original version, leaving as is.

For the new documentation in general, it would be good if you could follow the [new guidelines](https://cable.readthedocs.io/en/latest/developer_guide/documentation_guidelines/science_doc/). Don't worry about getting the formatting exactly right since the pull request isn't going to build the doc for you. We can correct what we need when it gets into the main branch.

Whyborn:
A lot of the commentary is in place to help both myself and reviewers, and will be changed to meet documentation guidelines before being pushed to a production branch. I'll address all the documentation changes in a single commit, then address each of the functional changes separately and reference the commit in a reply to each of the comments.

Would you like me to move the routines in ```cable_input.F90``` to ```cable_cru_TRENDY.F90``` for the review process (and maybe for the final product)? I think it makes sense to have them in ```cable_input.F90```, but it might be better to have all the changes in one place.

ccarouge:
I don't know yet where the routines in cable_input.F90 should be. I haven't yet looked at the code logic in enough detail. 

Suggestions can also be applied directly from github in bulk. So you could add all the ones you agree to in one commit on github. Then add the other doc changes in a commit worked on locally and then the functional changes.

offline/cable_input.F90
```diff
@@ -2887,5 +2921,402 @@ SUBROUTINE allocate_cable_vars(air,bgc,canopy,met,bal,                         &
...
+  END DO FindStart
+
+  ! Iterate through for <enddate>
+  FindEnd: DO CharIndx = 1, LEN(CurrentFile) - 9
+    ! Check against <enddate>, if found set the index
+    IF (CurrentFile(CharIndx:CharIndx+8) == EndDate) THEN
+      IndxEndDate = CharIndx
+      EXIT FindEnd
+    END IF
+  END DO FindEnd
```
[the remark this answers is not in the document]
Whyborn: Addressed in [f762c68](https://github.com/CABLE-LSM/CABLE/commit/f762c68d9983f5a00009207a69cf8b0fa7f1c6c1)

offline/cable_input.F90
```diff
@@ -2887,5 +2921,402 @@ SUBROUTINE allocate_cable_vars(air,bgc,canopy,met,bal,                         &
...
+  CALL execute_command_line("ls -1 "//TRIM(CurrentFile)//" >&
+     __FileNameWithNoClashes__.txt", EXITSTAT = ExStat, CMDSTAT = CStat)
+
+  ! Now we want to read this temporary file back in and inspect the contents
+  ! to determine the date ranges for each file. We then write back to a
+  ! reserved name file, that now contains the original filenames as well as
+  ! the year ranges for each file as described in the opening preamble to
+  ! this subroutine.
+
+  OPEN(InputUnit, FILE = "__FileNameWithNoClashes__.txt", IOSTAT = ios)
```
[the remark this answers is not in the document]
Whyborn: Intended to, figured no reason to define a variable for it and forgot to remove definition. Addressed in https://github.com/CABLE-LSM/CABLE/commit/d51b18819a2c1d0e49168e16e43af67fa60fce63

Whyborn:
The routines now achieve bitwise equivalence for the S1, S2 and S3 experiments. The various other subexperiments that existed in the ```cable_cru_trendy.F90``` are not available directly, but can be achieved through setting of the ```cru.nml``` namelist. The branch of TRENDY_V12 at [5-Update-Configuration-for-Met-Refactor)](https://github.com/CABLE-LSM/TRENDY_V12/tree/5-Update-Configuration-for-Met-Refactor) is a version which points to the symlinked Met data, renamed to the desired format.

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+  ! Uses the NetCDF dataset at ```LandmaskFile``` to determine which cells are
+  ! computed in this run of CABLE.
+
+  ! Read the landmask file to determine which land cells we're looking at.
+  ! Define the input variables
+  CHARACTER(len=256), INTENT(IN)  :: LandmaskFile
+  TYPE(cru_type), INTENT(INOUT)   :: CRU
+
+  ! Assign an integer to store the file ID, and to record the error status
+  INTEGER       :: FileID, ErrStatus
```
ccarouge, line 711: VSCode tells me there is a `ErrStatus` variable in the parent scope. Want to use that one instead?

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+  ! Need a counter and iterators when iterating through the points in the land mask
+  INTEGER       :: MaskCounter, LatIndx, LonIndx, VarIndx
+
+  INTEGER, DIMENSION(:,:), ALLOCATABLE :: LandMask
+
+  ! Now we attempt to read the netCDF file
+  ErrStatus = NF90_OPEN(TRIM(LandmaskFile), nf90_nowrite, FileID)
+  IF (ErrStatus /= NF90_NOERR) THEN
+    CALL NC_ABORT(ErrStatus, "Error reading the landmask file.")
+  END IF
```
ccarouge, line 731: Why use `NC_ABORT()` and not `handle_err()`? I would settle on using only one. Was that a copy/paste code already there?

The comment is valid for the other `NC_ABORT()` calls in this file.

```suggestion
  CALL handle_err(ErrStatus, "Error reading the landmask file.")
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+SUBROUTINE prepare_temporal_dataset(FileName, TargetArray)
+  !*## Purpose
+  !
+  ! Read the CO2 dataset.
+  !
+  !## Method
+  !
+  ! Read the ```FileName``` text file which contains the yearly CO2 data in
+  ! ```YEAR VALUE``` pairs into ```TargetArray```.
+
```
ccarouge, line 847: If the name is generic, so should the description. Nothing in the name of the subroutine refers to CO2 specifically, so let's remove it from the description.


```suggestion
  ! Read a dataset varying in time but constant in space.
  !
  !## Method
  !
  ! Read the ```FileName``` text file which contains the time-varying data in
  ! ```TIME VALUE``` pairs into ```TargetArray```.

```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+  ! values.
+  IF (CRU%DirectRead) THEN
+
+    ApplyLandmaskDirect: DO GridCell = 1, CRU%mLand
+      ok = NF90_GET_VAR(CRU%NDepFID, CRU%NDepVID, CRU%NDepVals(GridCell),&
+      START = (/land_x(GridCell), land_y(GridCell), TimeIndex/))
+      IF (ok /= NF90_NOERR) THEN
+        CALL handle_err(ok, "Reading from NDep")
+      END IF
+    END DO ApplyLandMaskDirect
```
ccarouge, line 1032:
```suggestion
      CALL handle_err(ok, "Reading from NDep")
    END DO ApplyLandMaskDirect
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+  ELSE
+
+    ALLOCATE(TmpArray(CRU%xDimSize, CRU%yDimSize))
+
+    ok = NF90_GET_VAR(CRU%NDepFID, CRU%NDepVID, TmpArray,&
+      START = (/1, 1, TimeIndex/), COUNT = (/CRU%xDimSize, CRU%yDimSize, 1/))
+    IF (ok /= NF90_NOERR) THEN
+      CALL handle_err(ok, "Reading from NDep")
+    END IF
+
```
ccarouge, line 1042:
```suggestion
    CALL handle_err(ok, "Reading from NDep")

```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+      CRU%NDepVals(GridCell) = TmpArray(land_x(GridCell), land_y(GridCell))
+    END DO ApplyLandmaskIndirect
+  END IF
+
+
+END SUBROUTINE get_cru_ndep
+
+!!------------------------------------------------------------------------------!
+
+ELEMENTAL FUNCTION Esatf(TC)
```
ccarouge, line 1053: I'll open an issue to remove this function later. It already exists in the science code. I think there are several versions of it for various input type/dimensions. So no need to have this in here.

ccarouge (commented):
There are quite a few comments. The main one I think is about setting the `nMet` variable and using that to remove all the special cases for the diffuse fraction.

Happy to discuss anything that isn't clear

ccarouge (changes requested):
There are quite a few comments. The main one I think is about setting the `nMet` variable and using that to remove all the special cases for the diffuse fraction.

Happy to discuss anything that isn't clear

offline/cable_input.F90
```diff
@@ -2887,5 +2924,377 @@ SUBROUTINE allocate_cable_vars(air,bgc,canopy,met,bal,                         &
...
+    IF (ok == NF90_NOERR) THEN
+      EXIT FindVar
+    END IF
+  END DO FindVar
+
+  ! Handle the case where we don't find the variable
+  IF (ok /= NF90_NOERR) THEN
+    CALL handle_err(ok, "Error finding desired variable in "//Dataset%FileNames&
+      (Dataset%CurrentFileIndx))
+  END IF
```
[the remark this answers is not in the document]
Whyborn: This was a holdover from the existing code. There is some argument I think to perhaps changing the name of the function to something like `handle_iostatus` because the implication is that the function only handles errors. I think that's a problem for another time, I'll remove these clauses from `IF` blocks.

offline/cable_input.F90
```diff
@@ -2887,5 +2924,377 @@ SUBROUTINE allocate_cable_vars(air,bgc,canopy,met,bal,                         &
...
+    ! If the requested is:
+    !   - Before the time-range of our data, then use the first day from the
+    !     first file in the dataset
+    !   - After the time-range of our data, then use the last day from the last
+    !     file in the datasetc
+    IF (YEAR < STD%StartYear(1)) THEN
+      ! Before the first year
+      STD%CurrentFileIndx = 1
+      YearIndex = STD%StartYear(1)
+      TimeIndex = 1
```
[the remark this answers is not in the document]
Whyborn: This was to get bit-wise reproducibility with the existing routines. I couldn't tell you the why unfortunately. I suspect it might be to handle the `prevTmax` and `nextTmin` "consistently" at the start of the centennial run, but I agree it would make more sense to get the equivalent day from the nearest year. There's already special handling for these next/prev variables.

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1101 @@
...
+
+IMPLICIT NONE
+
+TYPE CRU_MET_TYPE
+  REAL, DIMENSION(:), ALLOCATABLE :: MetVals
+END TYPE CRU_MET_TYPE
+
+TYPE CRU_TYPE
+  INTEGER   :: mLand              ! Number of land cells in the run
+  INTEGER   :: nMet               ! Number of Meteorological variables
```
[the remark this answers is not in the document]
Whyborn: I think there are two instances where keeping the `IF (CRU%ReadDiffFrac)` is wiser in `CRU_GET_SUBDIURNAL_MET` where we manipulate the `met` data structure. There is a logical attached to the `cable_user` struct, `calc_fdiff`, which should control all this behaviour, so `ReadDiffFrac` should not exist at all. Thoughts on this?

offline/cable_input.F90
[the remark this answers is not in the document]
ccarouge: We certainly need to settle on one set of error-handling routines for netcdf and other file I/O. As I said, it is hard for me sometimes to figure out what is new and what is left over from previous code version.

offline/cable_cru_TRENDY.F90
```diff
@@ -436,65 +407,55 @@ SUBROUTINE cru_get_daily_met(CRU, LastDayOfYear)
...
+  DummyDay = CRU%CTStep + 1
+  DummyYear = CRU%cYear
+
+  ! Special handling at the last day of the year
+  DaysInYear = 365
+  IF (CRU%LeapYears) THEN
+    DaysInYear = 366
+  END IF
+
+  IF (CRU%CTStep == LastDayOfYear) THEN
```
ccarouge, line 446: Was it what I had in the suggestion? Don't we need a `is_leapyear(CRU%cYear)` as well in the IF condition? Because we only have 366 days during a leap year, other years will have 365 days even if `CRU%LeapYears` is TRUE. Also you forgot to change `LastDayOfYear` to `DaysInYear`.

```suggestion
  DaysInYear = 365
  IF (CRU%LeapYears .AND. is_leapyear(CRU%cYear)) THEN
    DaysInYear = 366
  END IF

  IF (CRU%CTStep == DaysInYear) THEN
```

ccarouge:
@Whyborn What tests have you run since the last changes to the code have been made? Can we ensure we still get bitwise reproducibility of previous simulations?

ccarouge:
@Whyborn can you resolve the "conversations" where changes have been made? This will allow me to know where we stand here.

Whyborn:
Conversations marked as resolved @ccarouge

offline/cable_cru_TRENDY.F90
ccarouge: Issue created #318

Whyborn: This will be addressed in a later PR.

Whyborn:
Results using the CABLE-POP_TRENDY branch have been bit-wise reproduced for the experiments `S0-3` on reduced domains using the adapted namelists at [TRENDY_V12](https://github.com/CABLE-LSM/TRENDY_V12/tree/5-UpdateConfiguration-for-Met-Refactor).

mcuntz:
Dear Lachlan @Whyborn ,

I wanted to test the new met routines on our machine before accepting the changes. For this I would need to take the changes from the `5-Update-Configuration-for-Met-Refactor` branch of the `TRENDY_v12` repository. This is a bit convoluted now with my and your parallel amendments. Could you help me with this, please?

1. I take the changes in `cru.nml`.
2. I add a `met_names.nml`.
3. I have to provide the forcing files with the new names. Do you already have a nice one-liner to rename all files in a directory to the filename? Perhaps even all files in a list of directories? (I can do the co2 and ndep by hand ;-)
4. Then there are a few other changes in run_cable.sh. Do I need the `*Recycle` variables? Do I need `CO2Method = "Yearly"` (same for ndep)?

Whyborn:
My intention was to merge in my changes to the configuration at `TRENDY_V12` after you had merged your changes. It should be relatively easy to do, since the changes are quite separate.

1. Will depend on where you symlink the met data to
2. `met_names.nml` should be able to be taken as is
3. I do have a python script which I used to symlink the `CRU-JRA_1x1` data at [symlink_files.py](https://gist.github.com/Whyborn/97c57babc7a531e0605c9d44802a7ce1)
4. The `Recycle` denotes whether the Met data should be recycled, for `CO2Method` and `NDepMethod` are `"Yearly"` to get it from the Met year and `YYYY` to get it from a particular year.

ccarouge:
@mcuntz:
> I have to provide the forcing files with the new names.

The forcing files in `cru.nml` can be named in any way you want as long as they follow a template (for each variable) with a start date (YYYYMMDD format) and an end date (YYYYMMDD format).

@Whyborn I think the description of `File` in the README should be beefed up to explain what the fields accept.

Whyborn:
I had a longer description in the bulk of the README on the Met file template, but I've moved it to its own heading and linked to that heading under the ```File``` list item so it's easier to find.

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1069 @@
...
+USE cable_input_module,     ONLY: SPATIO_TEMPORAL_DATASET, find_variable_ID,&
+                                prepare_spatiotemporal_dataset, read_metvals,&
+                                open_at_first_file
+USE cable_def_types_mod,    ONLY: MET_TYPE, r_2, mLand
+USE cable_weathergenerator, ONLY: WEATHER_GENERATOR_TYPE, WGEN_INIT,&
+                                WGEN_DAILY_CONSTANTS, WGEN_SUBDIURNAL_MET
+USE mo_utils,               ONLY: eq
+#IFDEF __MPI__
+USE mpi,                    ONLY: MPI_Abort
+#ENDIF
```
mcuntz, line 27: Must be lowercase `ifdef` and `endif`:
```
#ifdef __MPI__
 USE mpi,                    ONLY: MPI_Abort
#endif
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1069 @@
...
+
+  ! Set up and read the namelist
+  NAMELIST /crunml/ rainFile, lwdnFile, swdnFile, presFile, qairFile,&
+                    TmaxFile, TminFile, uwindFile, vwindFile, fdiffFile,&
+                    CO2File, NDepFile, landmaskFile,&
+                    rainRecycle, lwdnRecycle, swdnRecycle, presRecycle,&
+                    qairRecycle, TmaxRecycle, TminRecycle, uwindRecycle,&
+                    vwindRecycle, fdiffRecycle,&
+                    ReadDiffFrac, CO2Method, NDepMethod, MetRecyc, LeapYears,&
+                    RecycStartYear, DtHrs, DirectRead
```
mcuntz, line 550: Definition of namelist must be in definition section of file, i.e. move it to line 502.

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1069 @@
...
+      HeaderCounter = HeaderCounter + 1
+    ELSE
+      ! Otherwise, its a line of useful data
+      LineCounter = LineCounter + 1
+
+      ! Read the line of data as a (time, value) pair. We don't want to use the
+      ! value this time around because we haven't yet allocated the array to
+      ! store it.
+      READ(LineInFile, FMT = '(I4, F)') Time, DummyValue
+    END IF
```
mcuntz, line 878: The Fortran standard does not allow simply 'F' as format. I propose to use free format because Fortran reads the correct type and precision due to the variable definition:
```
READ(LineInFile, *) Time, DummyValue
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1069 @@
...
+
+  ! First step over the header lines
+  SkipHeader: DO iter = 1, HeaderCounter
+    READ(FileID, FMT = '(A)', IOSTAT = ios) LineInFile
+  END DO SkipHeader
+
+  ! Now the useful contents of the file into the array
+  ReadValues: DO iter = StartTime, EndTime
+    READ(FileID, FMT = '(A)', IOSTAT = ios) LineInFile
+    READ(LineInFile, FMT = '(I4, F)', IOSTAT = ios) Time, TargetArray(iter)
```
mcuntz, line 901: Same as above. I propose:
```
READ(LineInFile, *, IOSTAT=ios) Time, TargetArray(iter)
```

offline/cable_cru_TRENDY.F90
```diff
@@ -1,1318 +1,1069 @@
...
-
-end module CABLE_CRU
+    ! Make sure it's in the bounds of the array
+    CO2Year = MAX(LBOUND(CRU%CO2Vals, 1), CO2Year)
+    CO2Year = MIN(UBOUND(CRU%CO2Vals, 1), CO2Year)
+
+    CO2Air(:) = CRU%CO2Vals(CO2Year)
+  ELSE
+    ! The user has specified the year
+    READ(CRU%CO2Method, '(I)') CO2Year
```
mcuntz, line 943: Same as above
```
READ(CRU%CO2Method, *) CO2Year
```
or
```
READ(CRU%CO2Method, '(I4)') CO2Year
```

mcuntz (changes requested):
`cable_cru_TRENDY.F90` does not compile with gfortran due to incompatibilities with the Fortran standard. I wrote the correct statements in the comments.

Otherwise I adapted the TRENDY_v12 repository and ran a TRENDY S3 run with success, having identical output and restart files.

Whyborn:
Great to hear. I've made the adjustments to the code to adhere to the Fortran standards. Hopefully you can now compile with gfortran. If you can, I'll close the conversation on those points.

mcuntz:
Compilation worked with gfortran. You could also try that on Gadi.

ccarouge (approved):
This looks good to go.

I have done a test for a single site using this branch and the head of the CABLE-POP_TRENDY branch. We don't expect differences since the new subroutines are only called if using the cru met. forcing for the moment, assuming other changes on the CABLE-POP_TRENDY branch were neutral to single site simulations.

The single site tests show identical values for the met forcing data. It shows very small differences in output variables. I've traced back at least some of the differences to the changes to qsatfjh() to avoid temporary arrays. Considering the tiny size of the differences, the fact the changes should only affect the CRU met forcings and the evidence of some numerical noise from other changes, I think it's safe to say these changes are neutral for the single sites.

mcuntz:
It intrigued me that the change to the temporary arrays should give differences.

So I went back to the CABLE-POP_TRENDY version with temporary arrays and compared it at the flux site FR-Hes against this 212_... branch. It did not give the same results.

I have to investigate further but I won't be able to do it until next Friday (26/07).

mcuntz:
I had a meeting canceled so pursued the topic. I was a bit too enthusiastic removing some unneeded parentheses in `qsatfjh`  so the order of calculation changed, which changed the bitwise comparison. It's back with the parentheses ;-)

It's all good now. Would be great of @SeanBryan51 could check that the "temporary array creation" is gone with the Intel compiler.

So now there are changes to earlier runs but this comes from the LUC changes of Jürgen, I think. I tried checking out different commits but the met-changes are now intermingled with the other changes and I am lost in the commits.
But I tested here: 1. the removing of creation of temporary arrays worked at single site runs (landmask with 1 gridcell set to 1), and 2. the 212 branch gave the same results for an TRENDY S3 run.

Next time I will do an intermediate branch even for such "inconspicuous" as removing the temporary array creation, I promise ;-)

ccarouge:
Thanks @mcuntz for digging into this. It's good to know where it is coming from.

PR #324, opened 2024-06-26
Read standard parameter ancillary data in NetCDF

### LINKED ISSUE #281: Reformat BIOS ancillary information
In the BIOS setup, the ancillary information is read in 2 steps:

- read gridinfo file
- read new information in binary files that replaces some of the information in the gridinfo file. Adds limits to the parameter values in CABLE. These limits will need to be applied in a pre-processing step
 
The binary file contains also additional variables that are not currently in the gridinfo file.

 We need to decide how to deal with this additional information: 

1. reformat to netcdf, store in a separate file and keep the current reading procedure by overwriting the information in CABLE. This would result in having a generic gridinfo file and then additional files for localised information.
2. or reformat to netcdf and overwrite the gridinfo file with that information. This would result in creating gridinfo files specific to each configuration as needed with different formats of gridinfo files for different configurations.

## Solution 1

It keeps the gridinfo file format constant but then adds a precedence rule between files. It makes it less clear to users what is the final source of the input information. Adds cases that need to be handled by CABLE instead of putting them in the pre-processing.

## Solution 2

It collates the information into one file which clarifies the source of the information to the users. Although, currently CABLE has rules of precedence in its input files. gridinfo has the least priority, then the met files and finally the restart file. It would require creating different gridinfo files for different configurations and the format of the gridinfo file would not be constant between configurations.

**Problem:** If we collate all the BIOS information in the gridinfo file, do we risk using different information coming from the restart file for example? Is this case improbable and only probable in case of user error? In the current implementation, does BIOS overwrite the information after reading in only the gridinfo file or after reading in all the other input files?

## Implementation

We tend to favour _solution 2_ but we need to clarify if the identified problem makes it impossible to implement it this way.

Some start towards an implementation is here: `/g/data/rp23/experiments/2024-04-17_BIOS3-merge/BIOS3_forcing/gridinfo_inprogress/convert_gridinfo_Aust.sh`.

Input data for this script is here: `/g/data/rp23/experiments/2024-04-17_BIOS3-merge/BIOS3_forcing/australia_0.05/params/nc`.

abhaasgoyal:
# CABLE
## Description

Fixes #281 

Aside from `climate` parameters, BIOS ancillary data is being read from the gridinfo. They are supposed to be read via a separate BLAZE NetCDF file https://github.com/CABLE-LSM/CABLE/issues/392

The original 0.5 degree gridinfo file ([`gridinfo_CSIRO_CRU05x05_4tiles.nc`](https://github.com/CABLE-LSM/bios_testing/blob/b45e0736cfd7fd1de1014319eb551b924542a96c/cable.nml#L4C63-L4C96)) went through 2 stages to make it compatible with reading to ancillary inputs

TODO: Once these PRs are merged, point to these specific files
1. Subsampling at 0.05 degrees using script https://github.com/CABLE-LSM/gists/pull/3
2. Add BIOS coastal remapping script for gridinfo to have the same mask for all input parameters. https://github.com/CABLE-LSM/gists/pull/4

## Type of change

New feature

## Testing

Run `act9test` and `1000pts-configuration` (TODO: provide links) to `BIOS_ancillary_netcdf` and `CABLE_POP-TRENDY` with `derived_parameters` and `check_parameter_values` being called after `cable_load_bios_params` instead of within `load_parameters`.


offline/cable_read.F90
ccarouge, line 1: Are the changes in this file from you? None of the added code seems to be used.

ccarouge (approved):
Good to go.

Issue #280, opened 2024-05-02
Consistency of ice points between soil and vegetation


ccarouge:
We want to enforce consistency in ice points between soil and vegetation in CABLE. Ideally, input data should be consistent but it can't be relied on.

We want to do the following

### ice soil => ice veg

All ice soil points (isoilm=9) should have ice vegetation (iveg=17), i.e.:
```
IF (soil%isoilm == 9) THEN
   veg%iveg=17
END IF
```
New ice vegetation tiles should have their parameters (veg%) changed to use the parameters for ice. This should ideally be done close to when `veg%` parameters are set.

### ice veg => ice soil

All ice vegetation (iveg=17) should have ice soil (isoilm=9). 

New ice soil tiles should have their soil parameters changed to use the parameters for ice. For frozen soil, the only ones we want are soil%rhosoil=910 and soil%css=2100. These 2 parameters need to be added to some CABLE constants (they are the density of frozen water and specific heat of frozen water), they will not be read from namelists or inputs. They are physics constants anyway.

Again this should be done close to when `soil%` parameters are set.

### Don't use hard-wired values

In the explanation above I used hard-wired values (iveg=17 and isoilm=9), we want to replace these by some constants instead of the direct hard-wired index. Some of this already exists in the code: https://github.com/CABLE-LSM/CABLE/blob/a6e0dccd19ee16158d64891979033f92d98197c5/src/params/grid_constants_cbl.F90#L47-L48
Just need to add the frozen vegetation index to it and use that. Eventually, we will want these in namelist files but let's not do that as part of this issue.

### Warning

Add a nice explicit warning message that the tiles have been changed for consistency. Explaining what changes CABLE does.

ccarouge:
@rkutteh Above is an explanation of what we need to do to advance on the GW merge. Let me know if it does not make sense.

JhanSrbinovsky:
cable_surface_types reads in the surface type index in AM3 at least. From memory it is partially done in JAC as well.

In the UM we effectively hard-wire soil%isoilm = 2 EXCEPT where veg%iveg = ICE. then we use soil%isoilm = 9 so that we can pick up our ICE params for soil type = ice. All the UM soil params come in spatially anyway.

[...]

PR #326, opened 2024-06-27
Add a table of physical constants to the user guide

rkutteh:
# CABLE

## Description

Added a table of physical constants and a table of mathematical constants to the user guide. Additional tables of constants can be added following the same pattern.

Fixed missing/incorrect units in code and documentation.

**It might be a good time at this stage to decide on the numerical value to use for the ice density: 910 from cable_soilparm.nml OR 921 from cable_phys_constants_mod.F90 ?**  Interestingly another well known code uses 917.

----
📚 Documentation preview 📚: https://cable--326.org.readthedocs.build/en/326/


documentation/docs/user_guide/constants/index.md
```diff
@@ -0,0 +1,16 @@
...
+
+|   CABLE constants   	 |
+|------------------------|
+| Physical constants   	 |
+| Mathematical constants |
+| 			 |
+| 			 |
+| 			 |
+| 			 |
+|          	         |
```
ccarouge, line 16: Any reason to have these extra empty lines here?

ccarouge (changes requested):
Not approving to prevent accidental merge until we settle on the density of ice value.

ccarouge:
@har917 For the density of ice, should we take advantage of this PR to settle on one value? Here is what we have so far:

| Source | Value | Comment |
|--------|-------|-----------|
| cable_soilparm.nml | 910 | Usually not used in offline but used in JaC |
| cable_phys_constants_mod.F90 | 921 | Used for freeze/melt water balance calculations |
| UM value | 917 | Is used in ACCESS |

Looking at Wikipedia, 917 kg/m3 is the closest value to the density of ice. And since we are going to use this value in ACCESS, it seems a good value to settle on.

One issue we have with offline CABLE is that the soil properties are given in the gridinfo file and their correctness depends on the file creator's checking. But we can at least make sure the values we provide in the code and the example namelist are agreed on.

har917:
It would be a good idea I think - but I have no particular preference either way (as long as it's reasonable - there's a whole stack of  processes that can cause variations to that value, what we need is consistency through the code base).

To add to the confusion - [SLI](https://github.com/CABLE-LSM/CABLE/blob/NESP2pt9_BLAZE/core/biogeophys/cable_sli_numbers.F90) uses 920. (GW seems to inherit the `cable_phys_constants_mod` value which is a relief).

 On the ancillary issue - we are reading in soil type.  Can't we - as part of @rkutteh issue - say 'where a land point has been determined to be of ice soil type - then set all the soil parameters to be that of ice as per the constants module'?

ccarouge:
@har917 If you don't mind about the value, I would move to 917.

> Can't we - as part of @rkutteh issue - say 'where a land point has been determined to be of ice soil type - then set all the soil parameters to be that of ice as per the constants module'?

Yes we can add that to #280 

Can we deal with SLI later? It's already introducing its own set of numbers and there is more than the ice density that is different there.

har917:
> Can we deal with SLI later?

Absolutely yes - I only went there as I wanted to check on the other soil models (SLI and GW).  Ensuring consistency with SLI constants will come up earlier than I suspect you think - since it's part of BIOS,

documentation/docs/user_guide/constants/index.md
rkutteh: Yes, as hinted to in the description, they are intended to accommodate in the future the names of the additional constants in CABLE/src/params but feel free to delete if they look ugly, I can reintroduce later on.

rkutteh:
> @har917 If you don't mind about the value, I would move to 917.
> 
> > Can't we - as part of @rkutteh issue - say 'where a land point has been determined to be of ice soil type - then set all the soil parameters to be that of ice as per the constants module'?

That's what we have already put in there, as in:

             soil%rhosoil(k) = density_ice  
             soil%css(k) =     csice        

where density_ice   and  csice    are from cable_phys_constants_mod.F90

@ccarouge  @har917 are you happy then if I go ahead and set density_ice = 917 in cable_phys_constants_mod.F90 and soilin%rhosoil to 917 for 'Permanent ice' in cable_soilparm.nml  ?
 
We need to be careful also about adjusting the numerical values of parameters like cs_rho_ice   ! heat capacity * density ice

documentation/docs/user_guide/constants/index.md
ccarouge: I would prefer to add rows when we need them. 

```suggestion

```

ccarouge:
@rkutteh 

> That's what we have already put in there

No, it isn't. This only sets the values for "new" icy soils. The idea would be if the soil type is given as ice in one of the input file, then overwrite the associated parameter values read in the file with the ones in the physical constants. All of them, not just the new ones we create with the check on the veg type.

ccarouge:
@rkutteh Please go ahead and update the density and cs_rho_ice values.

rkutteh:
> @rkutteh
> 
> > That's what we have already put in there
> 
> No, it isn't. This only sets the values for "new" icy soils. The idea would be if the soil type is given as ice in one of the input file, then overwrite the associated parameter values read in the file with the ones in the physical constants. All of them, not just the new ones we create with the check on the veg type.

OK, that's easy...

ccarouge:
@rkutteh Can you update the values here about ice density and specific heat for ice according to #332. Then, ask for review again.

rkutteh:
@ccarouge The pull request #332 for setting the ice density and specific heat has been merged, so I am asking for another review for the present pull.

ccarouge (approved):
This is all good.

### Testing

Using benchcab, we see identical results for both the fluxsite tests and the spatial configuration with this branch and the main branch. This seems to indicate no to minimal impact on the CABLE results in general.

ccarouge:
The CI is failing because of [a known issue](https://github.com/ACCESS-NRI/build-ci/issues/170). Let's wait a few days to see if this gets resolved before merging.

PR #332, opened 2024-07-03
Set ice density to 917 kg/m3

rkutteh:
Set density_ice = 917 kg/m3 in cable_phys_constants_mod.F90 and soilin%rhosoil to 917 kg/m3 for 'Permanent ice' in cable_soilparm.nml, in connection to #326 

----
📚 Documentation preview 📚: https://cable--332.org.readthedocs.build/en/332/


ccarouge (approved):
Following the current unavailability of analysis in modelevaluation.org, this PR is reviewed without the scientific evaluation results.

@rkutteh Good to go.

Issue #335, opened 2024-07-04
facilitate output of potential evaporation directly from the offline code base


har917:
There are many applications where direct knowledge of potential evaporation (and/or its relation to simulated evaporation or evapotranspiration) would be of interest. However this is not facilitated by CABLE despite the `canopy%epot` variable carrying this information.  While an approximation of average potential evaporation could be estimated from other outputs, the nonlinearity in the equations implies that it would be better evaluated by CABLE's inbuilt time averaging.

Note that `ssnow%potev` in CABLE is more strictly the 'latent heat flux consistent with potential evaporation from the soil' not evaporation itself - i.e. units are currently W/m2 not mm of water/s - this may need to be handled within this issue.

har917:
Process
- follow the general structure to add an additional output to offline model under flag out%potevap (check there maybe an existing entry)
- allocate output variable to `canopy%potev / air%rlam / ssnow%cls` (which is the conversion of the 'latent heat associated with potential evaporation to potential evpaoration'

har917:
@ccarouge I may have been unlucky with exactly when you did a merge/update of MAIN - but benchcab seems to have failed for me.

I think the problem is that the cable.nml file that gets created is/was still back with 'check%range = .FALSE.' - with the net result that nothing runs (and the bitwise comparison fails)  - see
[cable.nml.txt](https://github.com/user-attachments/files/16243971/cable.nml.txt)

I also note that the site specific outputs seem to reference `ERROR ../../src/nccmp_state.c:156 No such file or directory` -e.g. 
[AU-ASM_2011-2017_OzFlux_Met_S0_R0_R1.txt](https://github.com/user-attachments/files/16243990/AU-ASM_2011-2017_OzFlux_Met_S0_R0_R1.txt) indicating that something else is missing.

Any suggestions? Perhaps I need to do a fresh git clone but I did do a benchcab clean all.

har917:
> I think the problem is that the cable.nml file that gets created is/was still back with 'check%range = .FALSE.' 

I've checked - git status indicates that my local copy of benchcab is in sync with origin/main except for the config.yaml file, and the nml file attached is that created by benchcab (for AU-ASM). The out.txt file is giving the NAMELIST input error

ccarouge:
[...]

PR #345, opened 2024-07-15
Update value for checks%range

### LINKED ISSUE #344: Fix namelist incompatibility and some incompatibilities for GNU compiler in the new checks%range
PR #287 has changed the type of the range%checks entry in the cable.nml namelist, but the corresponding change in the namelist with the source code was not implemented.

In addition, the code has introduced some IF statements on integer values that [will not compile with GNU compiler](https://gcc.gnu.org/onlinedocs/gfortran/Implicitly-convert-LOGICAL-and-INTEGER-values.html). These should be replaced with appropriate LOGICAL expressions.

Additionally, some non-standard line split symbols where introduced (\ instead of &).

ccarouge:
# CABLE

## Description

Fixes #344. Update the namelist provided with the code for the changes to check%range and other fixes for compatibility with GNU compiler.

## Type of change

- [X] Bug fix

## Checklist

I have compiled the code and run the Tumbarumba test case with the updated namelist. The test case finished:
```
 Finished.    36.78621      seconds needed for year
 Finished.    36.78769      seconds needed for        70128  hours
```

----
📚 Documentation preview 📚: https://cable--345.org.readthedocs.build/en/345/

### DIFF
--- a/documentation/docs/user_guide/inputs/meteorological_forcing.md
+++ b/documentation/docs/user_guide/inputs/meteorological_forcing.md
@@ -127,5 +127,6 @@ CABLE’s time step size is calculated from the first two values of the time var
 
 ## Checking ranges
 
-It is possible to check that the ranges of the meteorological variables are physically possible using `check%ranges=.TRUE.` in the `cable.nml` namelist variable.
+It is possible to check that the ranges of the meteorological variables are physically possible using `check%ranges=1` or`check%ranges=2` (see [explanation][nml_desc]) in the `cable.nml` namelist variable.
 
+[nml_desc]: cable_nml.md#list-of-namelist-variables
--- a/documentation/docs/user_guide/inputs/pft_params_nml.md
+++ b/documentation/docs/user_guide/inputs/pft_params_nml.md
@@ -156,7 +156,8 @@ The CABLE default values of the vegetation parameters in Table 2 for the vegetat
 
 
 The CABLE distribution provides the default vegetation parameter values from Table 3 in the namelist file pft_params.nml, including the vegetation types from Table 1 in the top part.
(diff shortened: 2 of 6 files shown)


ccarouge:
@abhaasgoyal we forgot to update this namelist file with the changes to check the range.

har917 (commented):
@ccarouge @abhaasgoyal   This change looks okay ... BUT ...

I'm not sure about previous change - we've switched from a TRUE/FALSE LOGICAL kind for `check%ranges` to an integer but there's still a few instances in the code of `IF(check%range)THEN`

I can see that working for `check%range=0,1` but not for `2` - at least with some compilers.

har917 (approved):
In testing #335 and #346 I tested with `check%ranges = 0, 1, 2`.  This would seem to be fine - at least for the compiler on gadi.

ccarouge:
@har917 The main branch does not currently compile with GNU compiler anyway for other reasons (the abort() function for example accepts a string argument with Intel but not GNU...). But I agree we should change the IF statements as we don't want to go further from compatibility with GNU than we currently are. It's just going to be tricky to see all of these without being able to compile. I'll add that change and update the issue.

abhaasgoyal:
For clarity, maybe we can have the check `check%ranges /= NO_CHECK` in the `IF` statement?

abhaasgoyal (approved):
LGTM

abhaasgoyal (changes requested):
Edit: Apologies since already approved, but there was a small change which if done would be good. There are some `check%ranges=.TRUE.` in the documentation, which can also be changed (in `pft_params_nml.md`, and `meteorological_forcings.md`.

abhaasgoyal (approved):
LGTM

Issue #347, opened 2024-07-15
Hard-wired vegetation types


rml599gh:
An issue to start collecting cases where vegetation numbers are hard-wired or vegetation ordering is assumed.

subroutine um2cable_lp in cable_um_init_subrs.F90 appears to have a special case for permanent ice (confusingly called perma frost in the comment line). If I'm reading it correctly, this assumes ice is the last tile. I suspect it would be better to remove this special case from the code and just prepare ancillary files which have the values in that we want to use - I think this is mostly being used for soil ancillary fields.

har917:
***lakes***
- lakes tile are typically hard-wired to 16 throughout the code base (#341)
- `gw_hydro` includes 3 instances where the condition `veg%iveg .GE. 16` (or equivalently `.GE. lakes_cable`.  This assumes the ordering will be lake, then ice.

***vegetation***
- `cable_parameters` includes a set of conditions linking the number of vegetation type (`mytype`) and tile index `veg%iveg` to determine the `veg%deciduous` flag

- `cable_phenology` includes conditions on `veg%iveg` to determine a number of behaviours
- `cable_phenology` assumes the tile order when determimning if summergreen crop (line 91)

- `cbl_hyd_redistr` uses `%iveg` to determine whether hydraulic redistribution can occur

***phenology***
This needs a thorough review - there is clearly an assumption that relates the ordering of the phenology inputs to the vegetation type ordering - particularly impacts `casa_inout` and `casa_readbiome` 

***soil***
- `cable_roughness` includes a link from `iveg = 17` (ice) to roughness parameters (and hardwires soil type to `isoilm = 9` a the same time)
- `gw_hydro` includes multiple cases of `isoilm .EQ.`, `isoilm .NE.` etc. many of which are hard-wired
- `cbl_snow_accum`, `cbl_snowMelt` and `cbl_snowDensity` have conditions dependent on `isoilm /= 9`
- `cbl_conductivity` and `cbl_oldConductivity` have conditions dependent on `isoilm .EQ. 9`
- `cable_carbon` uses `%isoilm` as an index to a ancillary (line 295)
- `%isoilm` gets used through the coupled models to trigger whether to override the grid ancillary values or ice values for parameter values.
- `isoilm` is used in `cable_canopy` to switch to a JULES/MOSES for a coupling variable `%gswx_T`

rml599gh:
[...]

PR #348, opened 2024-07-16
fixing broken lake tiles in offline #335 and addresses #347

### LINKED ISSUE #335: facilitate output of potential evaporation directly from the offline code base

### LINKED ISSUE #341: lake tiles in offline appear broken
Conceptually the lake tile in CABLE is modelled as a permanently saturated bare soil surface.  In the coupled model this is achieved in two steps:
- at the beginning of each time step (explicit section) the soil is set to saturation. any increment of water necessary to achieve this is diagnosed into `%wb_lake` (see [um_update_soilsnow](https://github.com/CABLE-LSM/ACCESS3-JULES/blob/main/src/control/cable/util/update/cbl_um_update_soilsnow.F90))
- in `soilsnow`/`surfbv` the increment `%wb_lake` is used to extract water from runoff, drainage and from the deepest soil layer in an attempt to balance water within the time step.  Any residual imbalance is retained and then used as a scaling on river outflow (in fully-coupled model).

In offline we have retained the section in `surfbv` but do not have an obvious equivalent section to the resetting of soil moisture to saturation and evaluation of `%wb_lake`.  Since `%wb_lake` is initialised to zero the whole code becomes effectively inactive.   The net result is that lake tile behaves like a bare soil tile with notable variations in soil moisture (as seen in recent analysis of the offline-as-ACCESS configuration)

Also: follow up on what is supposed to happen with wetlands - are these supposed to be saturated grasslands?

har917:
# CABLE

## Description

This PR aims to achieve two tasks
- removing the hard-wired lake index from the offline code (including GW module)
- maintaining lakes at saturation in offline model (soil-snow only)

Fixes #341
Addresses #347

## Type of change

- [x] Bug fix
- [x] Science advancement 

## Checklist

----
📚 Documentation preview 📚: https://cable--348.org.readthedocs.build/en/348/

### DIFF
--- a/src/science/soilsnow/cbl_surfbv.F90
+++ b/src/science/soilsnow/cbl_surfbv.F90
@@ -10,6 +10,8 @@ SUBROUTINE surfbv (dels, met, ssnow, soil, veg, canopy )
 
 USE smoisturev_mod,               ONLY: smoisturev
     USE cable_common_module
+    USE grid_constants_mod_cbl, ONLY : lakes_cable
+    
 IMPLICIT NONE
 
     REAL, INTENT(IN) :: dels ! integration time step (s)
@@ -102,7 +104,7 @@ IMPLICIT NONE
 
     !  Rescale drainage to remove water added to lakes (wb_lake)
     ssnow%sinfil = 0.0
-    WHERE( veg%iveg == 16 )
+    WHERE( veg%iveg == lakes_cable )
        ssnow%sinfil  = MIN( ssnow%rnof1, ssnow%wb_lake ) ! water that can be extracted from the rnof1
        ssnow%rnof1   = MAX( 0.0, ssnow%rnof1 - ssnow%sinfil )
        ssnow%wb_lake = MAX( 0.0, ssnow%wb_lake - ssnow%sinfil)
--- a/src/offline/cable.nml
+++ b/src/offline/cable.nml
@@ -1,6 +1,6 @@
 &cable
-   filename%met = 'TumbaFluxnet.1.3_lake.nc'
-   filename%out = 'out_cable_lake.nc'
+   filename%met = 'TumbaFluxnet.1.3_met.nc'
+   filename%out = 'out_cable.nc'
    filename%log = 'log_cable.txt'
    filename%restart_in  = ' ' 
    filename%restart_out = './restart_out.nc'
--- a/src/offline/cable.nml
+++ b/src/offline/cable.nml
@@ -1,6 +1,6 @@
 &cable
-   filename%met = 'TumbaFluxnet.1.3_met.nc'
-   filename%out = 'out_cable.nc'
+   filename%met = 'TumbaFluxnet.1.3_lake.nc'
+   filename%out = 'out_cable_lake.nc'
    filename%log = 'log_cable.txt'
    filename%restart_in  = ' ' 
    filename%restart_out = './restart_out.nc'
@@ -23,7 +23,7 @@
    output%params    = .TRUE.  ! input parameters used to produce run
    output%patch     = .TRUE.  ! write per patch  
    output%balances  = .TRUE.  ! energy and water balances
-   check%ranges     = .FALSE.  ! variable ranges, input and output
+   check%ranges     = 0       ! variable ranges, input and output
    check%energy_bal = .TRUE.  ! energy balance
    check%mass_bal   = .TRUE.  ! water/mass balance
    verbose = .TRUE. ! write details of every grid cell init and params to log?
--- a/src/offline/cbl_model_driver_offline.F90
+++ b/src/offline/cbl_model_driver_offline.F90
@@ -53,6 +53,7 @@ USE cable_phys_constants_mod, ONLY : CCAPP   => CAPP
 USE cable_phys_constants_mod, ONLY : CEMLEAF => EMLEAF
 USE cable_phys_constants_mod, ONLY : CEMSOIL => EMSOIL
 USE cable_phys_constants_mod, ONLY : CSBOLTZ => SBOLTZ
+USE cable_phys_constants_mod, ONLY : density_liq
     !mrd561
     USE cable_gw_hydro_module, ONLY : sli_hydrology,&
          soil_snow_gw
@@ -104,6 +105,20 @@ REAL :: xk(mp,nrb)
 
 !iFor testing
 cable_user%soil_struc="default"
+
+!At start of each time step ensure that lakes surface soil layer is fully saturated.
+!Diagnose any water needed to maintain this - this will be removed from 
+!runoff, drainage and/or deepest soil layer in surfbv
+!For offline case retain the water imbalance between timesteps - permits
+!balance to be maintained in the longer term. This differs to the coupled model
+!where %wb_lake is zero'd each time step (and river outflow is rescaled)
+WHERE( veg%iveg == lakes_cable .AND. ssnow%wb(:,1) < soil%ssat ) 
+  ssnow%wbtot1(:)  = REAL( ssnow%wb(:,1) ) * density_liq * soil%zse(1)
+  ssnow%wb(:,1) = soil%ssat
+  ssnow%wbtot2  = REAL( ssnow%wb(:,1) ) * density_liq * soil%zse(1)
+ENDWHERE
+ssnow%wb_lake = ssnow%wb_lake + MAX( ssnow%wbtot2 - ssnow%wbtot1, 0.)
+
 
 CALL ruff_resist( veg, rough, ssnow, canopy, veg%vlai, veg%hc, canopy%vlaiw )
 
--- a/src/offline/cbl_model_driver_offline.F90
+++ b/src/offline/cbl_model_driver_offline.F90
@@ -106,15 +106,15 @@ REAL :: xk(mp,nrb)
 !iFor testing
 cable_user%soil_struc="default"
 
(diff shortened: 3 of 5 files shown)


har917:
commit 4aaccb5054aa5459c7c2d5ec97676f5fa4aa9241 produces bitwise equivalence on a case where the tumbarumba test case has been converted to a lake with very small canopy height (hc=0.001)
index_fix_Qh
index_fix_Qle
index_fix_Tscrn

har917:
@ccarouge @JhanSrbinovsky @rml599gh @bibivking - in implementing this fix to the offline code I've noticed that the coupled model sets the lake tile water balance at field capacity (`%sfc`) and not at saturation (`%ssat`).  Can anyone advise which would be correct? (I think it should be saturation but I'm not a hydrologist)

har917:
commit b326199e9dcbeabb54c84625949f537662b7b537 enforces saturation at the start of the time step.

The original time series of soil temperature and moisture were
original_soilM
original_soilT

This is revised to 
fixed_saturation_soilM
fixed_saturation_soilT

A comparison of original vs revised soil layer 1 moisture is
fixed_saturation_soilM1
noting that this is generated from output at end of time step so soil layer 1 can drop below `%ssat=0.42` in this output.

har917:
The revision also impacts the surface energy balance
fixed_saturation_Qh
fixed_saturation_Qle
and diagnostic screen level temperature and humidity
fixed_saturation_Tscrn
fixed_saturation_Qscrn

The causal relationship for some of these differences is difficult to assign - there is the immediate impact of maintaining saturation (more latent, less sensible),  an immediate impact through the lower surface temperatures (less latent, less sensible and less outgoing longwave), and a delayed response because of the temporal dynamics of soil moisture and temperature (in particular the timing of the transition from <> freezing changes)

src/science/soilsnow/cbl_surfbv.F90
```diff
@@ -10,6 +10,8 @@ SUBROUTINE surfbv (dels, met, ssnow, soil, veg, canopy )
 
 USE smoisturev_mod,               ONLY: smoisturev
     USE cable_common_module
+    USE grid_constants_mod_cbl, ONLY : lakes_cable
```
JhanSrbinovsky, line 13: My only comment with this is that we've just pivoted in AM3 to using a namelist input to define lakes_cable. Something which will need to be ported to offline anyway - so the is a fair solution until that happens

har917:
I am a bit concerned about the impact on the deepest soil layer moisture seen above - this is because of the lines in `surfbv` that remove water from that layer to fulfill the water requirements of the surface layer.  For this particular case this removal accumulates through time faster than the water can be replenished by downwards transport through the soil column and deep soil layer moisture reaches `(%sfc+%swilt)/2` for large parts of the year.

Note that this removal does not respect the `frozen_limit` nor any frozen fraction of soil moisture implying that the code may not conserve energy (and that the code is inefficient in that lines 112-115 is redundant given line 116-119)

Obvious tests would be
- remove 116-119 (and use the higher limit)
- remove 112-119 (and not allow water to be extracted from the deepest soil layer)
and/or
- modify these lines to acknowledge that only liquid water can be extracted and that `frozen_limit` must be abided by.

har917:
Looking to separate concerns - will focus this issue on implementing the ***same*** physics as in coupled mode.  We will leave the investigation as to whether to use `%sfc` or `%ssat` in formulations until later - noting that this may need additional flexibility to choice of formulation of `%wetfac`.

har917:
Following the coupled model approach and reverting to using `%sfc` in the code results in
fixed_field_soilM
fixed_field_soilT!
fixed_field_soilM1

The revised fix still extracts substantive water (over time) from the lowest layer but the gradient in soil moisture is reduce (which should have less impact on the dynamics).

har917:
The corresponding (using `%sfc` not `%ssat`) impacts on the variables are
fixed_saturation_Qh
fixed_saturation_Qle
fixed_field_Tscrn
fixed_saturation_Qscrn

There is a much reduced impact on screen level temperature  (which is closer to expectation) using this approach.

har917:
Benchcab tests ran successfully - noting that this is a largely meaningless assessment given that none of the sites should encounter the revised code.  These required the `check%ranges` patch - 
[config.yaml.txt](https://github.com/user-attachments/files/16370826/config.yaml.txt)

[benchmark_cable_qsub.sh.o121637899.txt](https://github.com/user-attachments/files/16370802/benchmark_cable_qsub.sh.o121637899.txt)

me.org is still down so sending to review.

har917:
In the current `gw_hydro` code there are some conditions on `%iveg .GE. 16` which implies lakes or ice.  The current implementation removes the hard-wired indexes (replacing `16` with `lakes_cable`) but retains the `.GE.` condition (i.e. the assumed tile ordering).

It should be possible to implement a minor variation to the changeset here to use `(veg%iveg == lakes_cable) .OR. (veg%iveg == ice_cable)` as the condition and resolve this.

JhanSrbinovsky:
sounds like a solid fix

ccarouge:
@har917 Sorry I'm getting to this a bit late. I'm a bit puzzled with the code in `surfbv`. You say some of it is redundant. But that's not how I read it.

I read this code for lake in this way:

```
       xxx = MAX(0.0_r_2, (ssnow%wb(:,ms) - REAL(soil%sfc(:),r_2))*soil%zse(ms)*1000.0)
       ssnow%sinfil  = MIN( REAL(xxx), ssnow%wb_lake )
       ssnow%wb(:,ms) = ssnow%wb(:,ms) - REAL(ssnow%sinfil / (soil%zse(ms)*1000.0), r_2)
```
will potentially remove some water from the 6th layer and then

```
       xxx = MAX(0.0_r_2, (ssnow%wb(:,ms) - 0.5*(soil%sfc + soil%swilt))*soil%zse(ms)*1000.0)
       ssnow%sinfil  = MIN( REAL(xxx), ssnow%wb_lake )
       ssnow%wb(:,ms) = ssnow%wb(:,ms) - ssnow%sinfil / (soil%zse(ms)*1000.0)
```
will potentially remove more water.

So I don't see these as being redundant. Maybe in reality, there will never be enough water to draw twice and one of this will be 0 but there is nothing in the code stopping it to happen.

Is the logic here trying to get `wb` down to `sfc` or `0.5*(sfc+swilt)` depending on which is smaller and some restrictions according to `wb_lake` value?

**Unit issue?**
I'm also not sure about units here. If we take the first case and wb > sfc, then we have:
```
xxx = wb - sfc*zse*1000
sinfil = wb - sfc*zse*1000.    (assumption on the min here)
wb = wb - sinfil/(zse*1000) = wb - (wb -sfc*zse*1000)/(zse*1000) = wb - wb/(zse*1000)+sfc
```
This "wb-wb/(zse*1000)" is very weird., I'm not sure the division by `zse*1000` should be in there when we update `wb`. 

I'm not sure this should be part of this fix or come as a different issue or if I'm completely wrong, so looking for your input first.

har917:
@ccarouge A couple of quick replies:
- I don't see why we need the first 
```
       xxx = MAX(0.0_r_2, (ssnow%wb(:,ms) - REAL(soil%sfc(:),r_2))*soil%zse(ms)*1000.0)
       ssnow%sinfil  = MIN( REAL(xxx), ssnow%wb_lake )
       ssnow%wb(:,ms) = ssnow%wb(:,ms) - REAL(ssnow%sinfil / (soil%zse(ms)*1000.0), r_2)
```
if we then go and do the second extraction.  Since `swilt<=sfc` always, `0.5*(sfc+swilt)<=sfc` and the net effect with/without these first 4 lines will be the same.  The code removes as much water from the 6th soil layer as required to satisfy `%wb_lake` without forcing `%wb` below `0.5*(sfc+swilt)` either way; these four lines just force it to do it in 2 steps not 1.

- the factor 1000 here is a `density_liq`.  We need to conserve water while noting that `sinfil` and `wb_lake` are in kg/m2 whereas `wb` is in m3 water/m3 soil (and `sinfil` in particular has other dependencies).  You're correct that the conversions are meaningless if you're at the limit on `xxx`, but they're not if you're not at the limit.

- there is a problem here in that none of this respects any frozen water in the soil (i.e. lake) or the frozen limit - but that's a separate issue and highly unlikely to be encountered.

ccarouge (approved):
All good.

har917:
@ccarouge one final comment on units:  you said you were concerned about units in
```
xxx = wb - sfc*zse*1000
sinfil = wb - sfc*zse*1000.    (assumption on the min here)
wb = wb - sinfil/(zse*1000) = wb - (wb -sfc*zse*1000)/(zse*1000) = wb - wb/(zse*1000)+sfc
```

You missed a bracket - it should be
```
xxx = (wb - sfc)*zse*1000
sinfil = (wb - sfc)*zse*1000.    (assumption on the min here)
wb = wb - sinfil/(zse*1000) = wb - ((wb - sfc)*zse*1000)/(zse*1000) = sfc
```
so everything works out

PR #349, opened 2024-07-16
consistency_ice_veg_soil

### LINKED ISSUE #280: Consistency of ice points between soil and vegetation

### LINKED ISSUE #338: consistency_ice_veg_soil
Add a subroutine to CABLE to ensure the consistency of ice points between soil and vegetation as described in #280

rkutteh:
# CABLE

Add a subroutine to CABLE to ensure the consistency of ice points between soil and vegetation, related to #338.
This branch will be used for the benchcab testing.

Fixes #280
Fixes #338

----
📚 Documentation preview 📚: https://cable--349.org.readthedocs.build/en/349/


rkutteh:
@ccarouge @JhanSrbinovsky @har917 While attempting to run benchcab on my consistency_ice_veg_soil branch and main I am getting the same errors from both codes:. 

The error is coming from cable.nml , specifically someone set check%ranges     = 0    instead of a logical


**from *.out:**

Cannot process command line definition of namelist fie using cable.nml as the namelsit


**from *.err:**

forrtl: severe (17): syntax error in NAMELIST input, unit 10, file /g/data/w97/rk4417/projects-svn/cable-dev/iceconsistency_benchcab/runs/payu-laboratory/work/crujra_access_R0_S0/cable.nml, line 26, position 21
Image              PC                Routine            Line        Source
cable-mpi          0000000000A9920B  for__io_return        Unknown  Unknown
cable-mpi          0000000000AC6D6E  for_read_seq_nml      Unknown  Unknown
cable-mpi          000000000045D58D  Unknown               Unknown  Unknown
cable-mpi          000000000040E071  Unknown               Unknown  Unknown
cable-mpi          000000000040DB62  Unknown               Unknown  Unknown
libc-2.28.so       000014D995AC6D85  __libc_start_main     Unknown  Unknown
cable-mpi          000000000040DA6E  Unknown               Unknown  Unknown

har917:
@rkutteh @ccarouge @abhaasgoyal MAIN is undergoing a transition from `check%ranges TRUE/FALSE` to `check%ranges 0,1,2`.   I suspect that because you are working on an older dev branch your branch will need TRUE/FALSE whereas the MAIN needs 0

The 'easy fix' will be to include
```
- repo: 
      git: 
        branch: <path-to-development-branch>
    patch:
      cable:
        check:
          ranges: .false.
```
in your benchcab.yaml file.

The better solution would be to update your dev branch with MAIN - but that possibly too much of a task for your current work. @ccarouge any other options?

rkutteh:
@har917 main itself fails to run, not just may branch

ccarouge:
@rkutteh Sorry, we broke benchcab with an update to CABLE. This triggered some thoughts on how best to fix but also on how to avoid that in the future. This resulted in a delayed solution. The solution is coming very soon.

For you, you'll need to add the following to your benchcab config.yaml:
```
- repo: 
      git: 
        branch: main
    patch:
      cable:
        check:
          ranges: 0
```
Then, both realisations should have the correct format for their namelists.

src/offline/cable_parameters.F90
```diff
@@ -2064,7 +2064,195 @@ SUBROUTINE check_parameter_values(soil, veg, ssnow)
...
+          k = landpt(i)%cstart + j - 1    ! absolute position of active patch 
+          
+          IF(soil%isoilm(k) == ICE_SoilType) THEN  ! check to see if ice soil patch
+
+             WRITE(*,*) 'SUBROUTINE load_parameters:'
+             WRITE(*,*) 'At land point number ', i
+             WRITE(*,*) 'And patch number ', k
+             WRITE(*,*) 'isoilm is ICE_SoilType'
+             WRITE(*,*) 'Set rhosoil = density_ice from CABLE physical constants'
+             WRITE(*,*) 'Set css     = csice from CABLE physical constants'
```
ccarouge, line 2113: The write statements need to be directed to `logn` logical unit so that they are written in the CABLE log file and not the standard output. 
`logn` is in `cable_IO_vars_module` which is in the USE statements on this `cable_param_module` module so it shouldn't need to be passed in by argument.

This is a comment valid for all the WRITE() statements not just the ones here.

ccarouge (changes requested):
I have merged in main into this branch in order to get spatial outputs with benchcab.

Looking at the outputs I couldn't quite say what was happening and saw the information from the consistency checks is missing from the log file so I have one change requested.

Once this is in, we can run benchcab and have a quick check on how the results are affected.

src/offline/cable_parameters.F90
rkutteh: @ccarouge Happy to make the change but just want to point out before I do so that I modelled these write statements after the ones in SUBROUTINE check_parameter_values. Overall there are 41 occurrences of WRITE(*,*) in cable_parameters.F90. I am guessing then that you are planning eventually to change all of them ?

rkutteh: Ok, I made the requested changes.

ccarouge: We will work on existing stuff later. For the moment, let's make sure the new stuff gives the info in the correct place. Thanks for making the change.

ccarouge:
Running the spatial configuration in benchcab, we get:

- vegetation type differences:
iveg_diff

- soil type differences:
isoil_diff

The differences definitely show where we expect them.
@har917 any last-minute thoughts on this? I think this should probably go behind a namelist option.

har917:
@ccarouge @rkutteh I'm a bit surprised that there's a grid cell in Iceland with dominant ice-iveg which has necessitated the swtich to ice soil ... but I don;t know enough about this spatial configuration to know whether to worry about that.

Otherwise it looks plausible .. certainly the changes are all in regions where we would expect them.

The question around whether to put under a namelist really comes down to how the community wants to handle this kind of 'noticeable bug-fix' - see the #354/#355 discussion.  I don;t think we've settled on a position on this topic yet.

ccarouge:
> The question around whether to put under a namelist really comes down to how the community wants to handle this kind of 'noticeable bug-fix' - see the https://github.com/CABLE-LSM/CABLE/issues/354/https://github.com/CABLE-LSM/CABLE/issues/355 discussion. I don;t think we've settled on a position on this topic yet.

Interesting. I wouldn't put this as a bug fix. The code is working perfectly well before and after the change, we haven't changed how the model calculates the outputs or reads in the inputs. We add a "clean up" of the input data. I would consider this a new feature, as it allows people to forgo that QA step in their setup (it's irrelevant that it wasn't routinely done) and let CABLE do it for them. 

Obviously, not having this on can lead to inconsistent inputs resulting in erroneous outputs, but it isn't because CABLE has a bug, it is because proper QA on the inputs hasn't been done.

ccarouge:
@rkutteh I'll feel better with having this with a namelist option because as said earlier this is a new functionality and not a bug fix for me. I'll try and get this namelist option in today.

ccarouge:
Checked when using the namelist option set to true, we got the same results as without the namelist option. And results when set to false are the same as before the change.

ccarouge (approved):
It's good for me

ccarouge:
@rkutteh Can you check what I've done for the namelist option? And what I have added to the docs before we close this off. I can't set you as a reviewer because you've opened the pull request.

rkutteh:
> @rkutteh Can you check what I've done for the namelist option? And what I have added to the docs before we close this off. I can't set you as a reviewer because you've opened the pull request.

Yes, will do

rkutteh:
> @rkutteh Can you check what I've done for the namelist option? And what I have added to the docs before we close this off. I can't set you as a reviewer because you've opened the pull request.

Everything looks good. You may want to add the new ice consistency option to cable.nml with its default setting, as a way of informing the user, but that's up to you. Please let me know when you are happy for me to merge (I realize you already approved).

PR #352, opened 2024-07-25
Fix: uninitialised variables in cable_canopy.F90

### LINKED ISSUE #351: Uninitialised variables in canopy routines (sum_rad_gradis, DvLitt)
`sum_rad_gradis` is calculated as `SUM(rad%gradis,2)`. It is first calculated in the `dryLeaf` subroutine (cbl_dryLeaf.F90, line 201) , then it is passed into `wetLeaf` (in cable_canopy.F90, line 405) and used there. Finally it is re-calculated and used in `cable_canopy` after `wetLeaf` (cable_canopy.F90, line 421).
However, it is not listed in the call to `dryLeaf` so appears to be unreliably passed back to canopy for use in `wetLeaf`. Offline testing seemed to show that this rarely crashed the model, even when compiled with Debug. Sometimes a floating point overflow error was generated. When it did run, repeated runs were not always reproducible.
One fix would be to recalculate `sum_rad_gradis` between the calls to `dryLeaf` and `wetLeaf`. Another option is to pass `rad%gradis` into `wetLeaf` and recalculate the sum there. This is what was done in CABLE2.4 era code.

ccarouge:
# CABLE

## Description

Fixes #351. 

Move the definition of sum_rad_gradis and sum_rad_rniso out of the NITER loop in `define_canopy()` since these variables do not change in the loop. Having a quick comparison between the outputs before and after this commit [the effect is negligeable](https://github.com/CABLE-LSM/CABLE/pull/352#issuecomment-2249656007).

Use `sum_rad_gradis` and `sum_rad_rniso` as calculated in define_canopy() everywhere in define_canopy(), dryLeaf() and wetLeaf(), instead of recalculating them sometimes.

Move the initialisation of `canopy%DvLitt` and `canopy%kthLitt` to the top of `define_canopy()` since these variables are constant throughout the run. We also want these variables to be defined in all cases to avoid errors in function calls, but only used with `cable_user%litter` turned on. 

## Type of change

- [X] Bug fix

## Testing

benchcab/me.org analysis: https://modelevaluation.org/analyses/anywhere/BPLvifw3DyARokYdg/s6k22L3WajmiS9uGv/CJGXP5GQWhGf3nH28/all

Tested in benchcab the changes for DvLitt and kthLitt on their own. benchcab showed bitwise identical results when moving the initialisation compared to the previous implementation with the litter option on.

----
📚 Documentation preview 📚: https://cable--352.org.readthedocs.build/en/352/

### DIFF
--- a/src/science/canopy/cable_canopy.F90
+++ b/src/science/canopy/cable_canopy.F90
@@ -189,8 +189,9 @@ logical :: sunlit_veg_mask(mp)
     ! Initialise canopy%DvLitt and canopy%kthLitt. This value is only used if
     ! cable_user%litter is .TRUE.
     ! Reference:
-    ! Mathews (2006), A process-based model of offine fuel moisture,
+    ! Matthews (2006), A process-based model of fine fuel moisture,
     !                 International Journal of Wildland Fire 15,155-168
+    !                 https://doi.org/10.1071/WF05063
     ! assuming here u=1.0 ms-1, bulk litter density 63.5 kgm-3
     canopy%kthLitt = 0.3_r_2 ! ~ 0.2992125984251969 = 0.2+0.14*0.045*1000.0/63.5
     canopy%DvLitt = 3.1415841138194147e-05_r_2 ! = 2.17e-5*exp(1.0*2.6)*exp(-0.5*(2.08+(1.0*2.38)))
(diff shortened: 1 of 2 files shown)


ccarouge:
Since we are introducing 2 changes that will break reproducibility I have decided to document some of the differences separately.
In the first commit we only move the calculation of `sum_rad_gradis` before calling `wetLeaf`. benchcab shows differences in the results compared to the main branch but these are minimal. Here are the timeseries for Qle and Qh for the start of the runs for a random flux site:
Qh
Qle

rml599gh:
Also noting that my tests showed that without the sum_rad_gradis bug fixed, site runs were not consistently reproducible.

rml599gh (approved):
Reference at line 192. Matthews not Mathews. Title is 'of fine fuel' not 'of offine fuel'. Probably useful to include the doi (https://doi.org/10.1071/WF05063)

Issue #354, opened 2024-07-29
bug in spitter call (NESP2pt9 and TRENDY_POP branches)


har917:
At some point in the history of CABLE development the allocation of downwelling shortwave radiation into direct and diffuse components has gained bug for offline studies - this likely occurred at the time when the code  considered the fractionation  between visible and near-infrared radiation.

The direct-diffuse radiation (unless provided via other means) is evaluated through the use of the spitter function.  Originally this assigned the fractions based on the ratio of the timestep's shortwave radiation to a clear sky value (at that location and time of year).  When the partitioning between VIS:NIR occurred spitter was not adjusted and currently the call takes in the current VIS (or NIR) shortwave and compares to the clear sky value for *total* shortwave radiation (not the clear sky value for the VIS (or NIR) shortwave).  The ratio evaluated is then smaller than expected - and the diffuse fraction diagnosed is greater than it would be.

If only for testing purposes - we need to retrofit the model to allow for the original intent of spitter to be used.

This implementation will need to be different in the current MAIN (CABLE3) and in the CABLE-POP-TRENDY/NESP2pt9/BLAZE branches because of the code structure.

har917:
@ccarouge There are a couple of up front question we need answer as part of this issue
- as a bug fix - should this go in on a switch?
- how do we manage the code structure difference between MAIN and the older branches?

har917:
For CABLE-POP branch the suggestion is to work in the cable_radiation.f90, subroutine `init_radiation`.

In the short-term we can a logical switch - at around line 64 add
`LOGICAL :: use_spitter_bugfix          !to become a namelist option`
 
The default value of the new switch would be (apply the bug fix - which is a change) via line 67 add
`use_spitter_bugfix = .TRUE.`
 
Finally at the calls to spitter (line 132) is then replaced
```
! Define beam fraction, fbeam:
IF (use_spitter_bugfix) THEN   
      !beam fraction defined on total shortwave, equal in VIS and NIR
      rad%fbeam(:,1) = spitter(met%doy, met%coszen, met%fsd(:,1)+met%fsd(:,2))
      rad%fbeam(:,2) = spitter(met%doy, met%coszen, met%fsd(:,1)+met%fsd(:,2))
ELSE 
    !CABLE2.x scheme
     rad%fbeam(:,1) = spitter(met%doy, met%coszen, met%fsd(:,1))
[...]

PR #357, opened 2024-07-30
fix bug in spitter call (MAIN)

### LINKED ISSUE #355: bug in spitter call (MAIN branch)
As per #354 - we need to retrofit the model (MAIN branch) to allow for the original intent of spitter to be used in offline studies.

This will need to be different in MAIN compared to older offline branches such as CABLE-POP-TRENDY/BLAZE/NESP2pt9

har917:
# CABLE

## Description

Following analysis in CABLE4 and discussion in #354 we should use total SW down in the call to `spitter` and apply the resultant `Fbeam` equally to the visible and near-infrared bands.  The original intent (cable1.4) of spitter is to evaluate the fraction of SW down that is diffuse in nature - via the beam fraction `Fbeam`.  This was evaluated once based on the total shortwave down.  From CABLE2 onwards spitter was called twice - once for the visible and once for the near-infrared components.  However the function itself was not adjusted - the result is that beam fraction is universally smaller than the original intent with unknown impacts on the model trajectory.

This change set reverts the evaluation of `Fbeam` to operate on the total SW down and then applies `Fbeam` equally to the two radiation bands.

This has been implemented as a bug fix, not on a switch.  This change may require recalibration of the CABLE model (notably stomatal conductance parameters, leaf nitrogen exponential and any other parameters that impact GPP). 

Fixes #355

## Type of change
- [x] Bug fix

## Checklist

----
📚 Documentation preview 📚: https://cable--357.org.readthedocs.build/en/357/

### DIFF
--- a/src/science/radiation/cbl_init_radiation.F90
+++ b/src/science/radiation/cbl_init_radiation.F90
@@ -378,8 +378,9 @@ REAL, INTENT(IN) :: SW_down(mp,nrb)           ! Downward SW [met%fsd]
 
 
 ! Define beam fraction, fbeam:
-RadFbeam(:,1) = spitter(mp, cpi, metDoy, coszen, SW_down(:,1))
-RadfBeam(:,2) = spitter(mp, cpi, metDoy, coszen, SW_down(:,2))
+! #355 beam fraction defined using total SW_down; applies to VIS&NIR equally
+RadFbeam(:,1) = spitter(mp, cpi, metDoy, coszen, SW_down(:,1)+SW_down(:,2))
+RadFbeam(:,2) = RadFbeam(:,1)
 
 ! coszen is set during met data read in.
 WHERE (coszen < Ccoszen_tols_huge )


har917:
As expected benchcab comparing MAIN and this branch detected differences in many (but interestingly not all) cases - this is expected.  Closer inspection of these files indicates that this occurs in all output variables, in all output timestamps (except for some sites where run starts overnight) indicating exact replication up to the point of first non-zero SW_down which then triggers a change.  Detailed output indicates that variability in output over time/correlation between outputs remains qualitatively similar as expected given a bug fix.  

benchcab log at 
[benchmark_cable_qsub.sh.o122014778.txt](https://github.com/user-attachments/files/16423994/benchmark_cable_qsub.sh.o122014778.txt)

Proceeding to me.org for scientific assessment.

har917:
modelevaluation.org results [at](https://modelevaluation.org/analyses/anywhere/FHtqKb4KnJAMQDe5q/s6k22L3WajmiS9uGv/4MJGnH9zGcm2e6Sp6/all)

General impression is that not much has changed - there is
- a general shift from latent to sensible heat
-  slight overall improvement on LE
-  slight degradation on H [configuration dependent])
-  a general decrease in magnitude of NEE (degradation in the mean, overall improvement against observations)
- no particular difference in impact by PFT type

355-spitter-me-aggregate

har917:
@ccarouge This looks safe enough to apply as is - however should this go in on a switch?

ccarouge (approved):
This is good to go.

We will have to do a bugfix release for this, it will come in a few weeks I suspect once we agree on what information we should include in.

PR #369, opened 2024-08-15
Cable pop trendy cropfrac

juergenknauer:
# CABLE

## Description

Several bug fixes and code changes related to fixing the extent of croplands and their dynamics in CABLE.

## Type of change

- [ v] Bug fix
- [ v] New or updated documentation

## Checklist

- [v ] The new content is accessible and located in the appropriate section.
- [v ] I have checked that links are valid and point to the intended content.
- [v ] I have checked my code/text and corrected any misspellings

### DIFF
--- a/core/biogeochem/POPLUC.F90
+++ b/core/biogeochem/POPLUC.F90
@@ -1772,7 +1772,6 @@ CONTAINS
     ALLOCATE(POPLUC%crop(arraysize))
     ALLOCATE(POPLUC%past(arraysize))
     ALLOCATE(POPLUC%ptos(arraysize))
-    ALLOCATE(POPLUC%ptog(arraysize))
     ALLOCATE(POPLUC%stog(arraysize))
     ALLOCATE(POPLUC%gtop(arraysize))
     ALLOCATE(POPLUC%gtos(arraysize))
--- a/core/biogeochem/POPLUC.F90
+++ b/core/biogeochem/POPLUC.F90
@@ -84,7 +84,8 @@ MODULE POPLUC_Types
      REAL(dp), DIMENSION(:,:),POINTER :: FTransferGross => null()
      REAL(dp), DIMENSION(:),POINTER :: pharv => null(), smharv => null(), syharv => null()
      ! ag prod pool (grazing + crop harvest) and loss to atm, loss of C from biosphere due to crop/pasture harvest
-     REAL(dp), DIMENSION(:),POINTER :: AgProd => null(), AgProdLoss => null(), FAg => null()
+     REAL(dp), DIMENSION(:),POINTER :: AgProd => null(), AgProdPast => null(), AgProdCrop => null(), &
+          AgProdLoss => null(), AgProdLossPast => null(), AgProdLossCrop => null(), FAg => null()
      REAL(dp), DIMENSION(:,:),POINTER :: HarvProd => null(), ClearProd => null() ! wood harvest and clearance pools
      REAL(dp), DIMENSION(:,:),POINTER :: fracHarvProd => null(), fracClearProd => null()
      REAL(dp), DIMENSION(:,:),POINTER :: HarvProdLoss => null(), ClearProdLoss => null()
@@ -201,7 +202,11 @@ CONTAINS
     popluc%kClear                  = 0.0_dp
     popluc%cRelClear               = 0.0_dp
     popluc%AgProd                  = 0.0_dp
+    popluc%AgProdPast              = 0.0_dp
+    popluc%AgProdCrop              = 0.0_dp
     popluc%AgProdLoss              = 0.0_dp
+    popluc%AgProdLossPast          = 0.0_dp
+    popluc%AgProdLossCrop          = 0.0_dp
     popluc%FAg                     = 0.0_dp
     popluc%FluxPHarvResidtoLitter  = 0.0_dp
     popluc%FluxSHarvResidtoLitter  = 0.0_dp
@@ -280,7 +285,11 @@ CONTAINS
     write(*,*) 'kClear ', popluc%kClear
     write(*,*) 'cRelClear ', popluc%cRelClear
     write(*,*) 'AgProd ', popluc%AgProd
+    write(*,*) 'AgProdPast ', popluc%AgProdPast
+    write(*,*) 'AgProdCrop ', popluc%AgProdCrop
     write(*,*) 'AgProdLoss ', popluc%AgProdLoss
+    write(*,*) 'AgProdLossPast ', popluc%AgProdLossPast
+    write(*,*) 'AgProdLossCrop ', popluc%AgProdLossCrop
     write(*,*) 'FAg ', popluc%FAg
     write(*,*) 'FluxPHarvResidtoLitter ', popluc%FluxPHarvResidtoLitter
     write(*,*) 'FluxSHarvResidtoLitter ', popluc%FluxSHarvResidtoLitter
@@ -1505,17 +1514,27 @@ CONTAINS
 
        POPLUC%HarvProdLoss(g,:)  = kHarvProd * POPLUC%HarvProd(g,:)
        POPLUC%ClearProdLoss(g,:) = kClearProd * POPLUC%ClearProd(g,:)
-       POPLUC%AgProdLoss(g)      = kAgProd * POPLUC%AgProd(g)
+       POPLUC%AgProdLossPast(g)  = kAgProd * POPLUC%AgProdPast(g)
+       POPLUC%AgProdLossCrop(g)  = kAgProd * POPLUC%AgProdCrop(g) 
+       POPLUC%AgProdLoss(g)      = POPLUC%AgProdPast(g) + POPLUC%AgProdCrop(g)  
 
        if (POPLUC%grass(g).gt.0.0_dp .and. l.eq.j+2) then
+          ! Those quantities are calculated at grid cell level (hence multiplied by frac) 
           POPLUC%FAg(g)        = casaflux%Charvest(l)*patch(l)%frac
-          POPLUC%AgProd(g)     = POPLUC%AgProd(g) + casaflux%Charvest(l)*patch(l)%frac - POPLUC%AgProdLoss(g)
-          casaflux%charvest(l) = 0.0_dp
-          casaflux%nharvest(l) = 0.0_dp
-          casaflux%fharvest(l) = min(POPLUC%past(g)/POPLUC%grass(g),1.0_dp) * HIpast + &
-               min(POPLUC%crop(g)/POPLUC%grass(g),1.0_dp) * HIcrop !  fraction grass AGB to be removed next year
-          !write(*,*)'harvest', g,   patch(l)%frac,  casaflux%fharvest(l), POPLUC%crop(g),  POPLUC%past(g)
-          !casaflux%fharvest(l) = 0; ! test vh!
+          POPLUC%AgProdPast(g) = POPLUC%AgProdPast(g) + casaflux%CharvestPast(l) &
+                                 * patch(l)%frac - POPLUC%AgProdLossPast(g)
+          POPLUC%AgProdCrop(g) = POPLUC%AgProdCrop(g) + casaflux%CharvestCrop(l) & 
+                                 * patch(l)%frac - POPLUC%AgProdLossCrop(g)
+          POPLUC%AgProd(g)     = POPLUC%AgProdPast(g) + POPLUC%AgProdCrop(g)
+          casaflux%charvest(l)     = 0.0_dp
+          casaflux%charvestpast(l) = 0.0_dp
+          casaflux%charvestcrop(l) = 0.0_dp
+          ! Note: we are not keeping track of nharvest and pharvest for crops and pastures separately!
+          casaflux%nharvest(l)     = 0.0_dp
+          ! fraction grass AGB to be removed next year through grazing and harvest
+          casaflux%fharvestPast(l) = min(POPLUC%past(g)/POPLUC%grass(g),1.0_dp) * HIpast
+          casaflux%fharvestCrop(l) = min(POPLUC%crop(g)/POPLUC%grass(g),1.0_dp) * HIcrop
+          casaflux%fharvest(l)     = casaflux%fharvestPast(l) + casaflux%fharvestCrop(l)
           casaflux%fcrop(l) = min(POPLUC%crop(g)/POPLUC%grass(g), 1.0_dp)
        endif
        DO j=1, 3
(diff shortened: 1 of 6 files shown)


har917:
@juergenknauer @ccarouge Since this branch is what has been used for a major piece of output (and the branch is likely to advance in the future) - I think it's worth adding a tag at this point (actually after the merge happens).  How do we go about this?

core/biogeochem/POPLUC.F90
```diff
@@ -665,9 +687,6 @@ SUBROUTINE POPLUCStep(POPLUC,year)
          if (POPLUC%ptog(g) .gt. 0.0_dp) &
               CALL execute_luc_event('PRIMF','C3ANN',POPLUC%ptog(g),g,POPLUC)
 
-         ! if (POPLUC%stog(g) .gt.0.0_dp) &
-         !      CALL execute_luc_event('SECDF','C3ANN',POPLUC%stog(g),g,POPLUC)
-
          if (POPLUC%gtop(g) .gt.0.0_dp) &
```
har917, line 690: Not the topic of this particular PR - but should we remove `%gtop` from the code base entirely?

core/biogeochem/POPLUC.F90
```diff
@@ -1439,18 +1458,20 @@ SUBROUTINE POP_LUC_CASA_transfer(POPLUC, POP, LUC_EXPT, &
              irp = ilu + j -1
              ! Update tile area
              if (ilu == p) then
-                POPLUC%primf(g)  = max(patch(irp)%frac + dA_r(ilu) + dA_d(ilu), 0.0_dp)
+                POPLUC%primf(g) = min(1.0, max(patch(irp)%frac + dA_r(ilu) + dA_d(ilu), 0.0_dp))
```
har917, line 1461: My slight concern here is that if we need to constrain things between 0 and 1 here then we (likely) need a compensatory  adjustment to one of `dA_r(ilu)` and/or `dA_d(ilu)` which then feeds onto the LUC states.

core/biogeochem/POPLUC.F90
```diff
@@ -1493,17 +1514,27 @@ SUBROUTINE POP_LUC_CASA_transfer(POPLUC, POP, LUC_EXPT, &
...
-          casaflux%fharvest(l) = min(POPLUC%past(g)/POPLUC%grass(g),1.0_dp) * HIpast + &
-               min(POPLUC%crop(g)/POPLUC%grass(g),1.0_dp) * HIcrop !  fraction grass AGB to be removed next year
-          !write(*,*)'harvest', g,   patch(l)%frac,  casaflux%fharvest(l), POPLUC%crop(g),  POPLUC%past(g)
-          !casaflux%fharvest(l) = 0; ! test vh!
+          POPLUC%AgProdPast(g) = POPLUC%AgProdPast(g) + casaflux%CharvestPast(l) &
+                                 * patch(l)%frac - POPLUC%AgProdLossPast(g)
+          POPLUC%AgProdCrop(g) = POPLUC%AgProdCrop(g) + casaflux%CharvestCrop(l) & 
+                                 * patch(l)%frac - POPLUC%AgProdLossCrop(g)
+          POPLUC%AgProd(g)     = POPLUC%AgProdPast(g) + POPLUC%AgProdCrop(g)
+          casaflux%charvest(l)     = 0.0_dp
```
har917, line 1529: Note to self - the use of `l` here and restriction on hard-wires the POPLUC grass tile to be the last of the patch/tiles in each grid cell

offline/cable_LUC_EXPT.F90
```diff
@@ -337,7 +337,7 @@ SUBROUTINE LUC_EXPT_INIT(LUC_EXPT)
                      start=(/land_x(k),land_y(k),1/), &
                      count=(/1,1,tdimsize/) )
                 CALL HANDLE_ERR(STATUS, "Reading direct from "//LUC_EXPT%TransFile(i) )
-                !IF (sum(tmpvec).gt.1e-3 .OR. LUC_EXPT%primaryf(k).lt.0.99) LUC_EXPT%prim_only(k) = .FALSE.
+                !IF ((sum(tmpvec) .gt. 1e-3) .OR. (LUC_EXPT%primary_veg(k) .lt. 0.99)) LUC_EXPT%prim_only(k) = .FALSE.
```
har917, line 340: Delete these two commented out lines entirely?

offline/cable_LUC_EXPT.F90
```diff
@@ -166,7 +188,14 @@ SUBROUTINE LUC_EXPT_INIT(LUC_EXPT)
     LUC_EXPT%VAR_NAME(16) = 'crop'
     LUC_EXPT%VAR_NAME(17) = 'past'
 
-    LUC_EXPT%nfile = 17
```
har917, line 169: could also delete the blank line above

offline/cable_LUC_EXPT.F90
```diff
@@ -346,42 +346,67 @@ SUBROUTINE LUC_EXPT_INIT(LUC_EXPT)
...
+    LUC_EXPT%primaryf = LUC_EXPT%primary_veg * LUC_EXPT%woodfrac
+    LUC_EXPT%grass    = LUC_EXPT%grass + LUC_EXPT%primary_veg * (1.0 - LUC_EXPT%woodfrac)
+    LUC_EXPT%secdf    = max(1.0-LUC_EXPT%grass-LUC_EXPT%primaryf, 0.0)
+
+    !! JK Debug
+    ! For some reason the transition from secdf does not work correctly
+    ! if there is secdf = 0.0 at first timestep (1580). This is a bandaid
+    ! solution that allows a 'correct' transition. Minimum fraction of 0.01
+    ! was required for it to work correctly.
+    where ((LUC_EXPT%secdf < 0.01) .and. (LUC_EXPT%prim_only .eqv. .FALSE.))
```
har917, line 390: How does this work on grid cells set to be `%prim_only = .TRUE.`?

offline/cable_LUC_EXPT.F90
```diff
@@ -745,11 +784,6 @@ subroutine READ_ClimateFile(LUC_EXPT)
     LUC_EXPT%ivegp = climate%iveg
     LUC_EXPT%biome = climate%biome
 
-    ! non-woody potential vegetation not considered to undergo LU change
```
har917, line 748: Good to see you're happy with removing this condition!

offline/cable_LUC_EXPT.F90
```diff
@@ -935,20 +969,20 @@ SUBROUTINE READ_LUH2(LUC_EXPT)
     endif
 
     ! Adjust transition areas based on primary wooded fraction
+    ! Note that only the four transitions ptos, ptog, gtos, and stog are
```
har917, line 972: Thanks for including a comment

har917: Actually - on further thought - do `ptoc`, `ptoq`, `stoc`, `stoq`, `ctos` and `qtos` need a wood fraction or not?  These transitions are not within the grass type.

They may actually need to be `ptoc = ptoc * ptog` etc. here. - but this connects into exactly how the updated LUH2 processing works.

(for the moment it's probably okay - the code will work in serial mode which is what you've used for TRENDY v13)

har917 (approved):
A few overarching comments:
- This will break the MPI code - the new POPLUC% fields need to be broadcast across processors.
- I'm hopeful that the C13 code will be okay  - but do not know enough to say yay or nay.
- I'm keen that we remove %prim_only from the code base and deal with these cases via the inputs - however deserts/lakes may cause a problem.

My only immediately substantive comment is that I think we need to apply the wood fraction to more of the gross transitions - see inline comment - basically anything that goes into/out of the primary and secondary vegetation needs this.  However this may depend on how the LUH2 has been processed.   

Since the primary purpose here is to lock-in the code base used for TRENDY v13 into the CABLE-POP_TRENDY branch happy to approve.

juergenknauer:
> @juergenknauer @ccarouge Since this branch is what has been used for a major piece of output (and the branch is likely to advance in the future) - I think it's worth adding a tag at this point (actually after the merge happens). How do we go about this?

That's a good idea. Tagging the version I have used for TRENDY would be highly useful.

offline/cable_LUC_EXPT.F90
juergenknauer: It should be fine like this. They are included in the four main classes. For example, ptoc and ptoq are already accounted for through ptog. stoq is in included in stog. ctos is included in gtos. etc. 

After TRENDY I will also commit the LUH2 postprocesssing script that clarifies this in more detail.

offline/cable_LUC_EXPT.F90
juergenknauer: That was very misleading actually. Good riddance!

offline/cable_LUC_EXPT.F90
juergenknauer: then it is treated as any other prim_only grid cell.

offline/cable_LUC_EXPT.F90
juergenknauer: can do for the next commit.

offline/cable_LUC_EXPT.F90
juergenknauer: Can delete for the next commit.

core/biogeochem/POPLUC.F90
juergenknauer: yes, this assumption is made in a couple of places in the code. One would need to be really careful to disentangle that.

core/biogeochem/POPLUC.F90
juergenknauer: Good point. I probably haven't thought this through. I will keep it in mind for the next commit. For now I will double-check the patchfracs in the results and ensure they all add up to 1 (which I am pretty sure they do).

core/biogeochem/POPLUC.F90
juergenknauer: Fine with me. I think the logic that nothing can be converted to p is reasonable.

juergenknauer:
Thanks @har917 for your quick review. I did not have the capacity to test this with MPI. Hopefully one can get it working with MPI relatively easily. I have asked Matthias to review the C13 code separately and make changes if needed. 

Happy to remove prim_only. For TRENDY I did not remove it but followed the current logic in the code. There might actually be a good reason why it is in there and I am not sure the code would work with having it removed completely. 

Regarding the woodfrac. I tried to apply it to more transitions but it made things worse. Now I am pretty certain that the current implementation works best. I can provide a few examples at some point to clarify.

offline/cable_LUC_EXPT.F90
har917: unfortunately no.

For example (albeit a really odd situation) 
- let grass=crop=pasture=0 at the start of the time step
- let there be (only) a ptog which is a combination of ptoc and ptoq

if ptoq is used to evaluate the %pasture fraction (line 1471 in POPLUC.F90)  without the wood_fraction include, then you'd get pasture = grass (since you hit the hard limit) and crop = 0  - that's not the intent.

you can create other quirks (e.g. combinations of stog and stoc can lead to crop=0 error).

juergenknauer: Not sure I follow. Let's say grass = 0 at the beginning. And you have a conversion of ptog = 0.05, ptoq = 0.025, and ptoc = 0.025 and no other transitions. Shouldn't that work?

The way the LUH2 processing script works is that ptog is the sum of all transitions from primary to all grass types (so ptog = ptoq + ptoc in this case). So it should be constrained in that way. But there might edge cases!

har917: But say wood_frac is 0.4 - then ptog becomes 0.02 but ptoq and ptoc stay at 0.025.

The sequence is then grass->0.02, pasture -> 0.02 (because pasture<=grass at line 1471) and crop -> 0.0 (because crop < grass-pasture at line 1473)

This is an edge case - and one that is highly unlikely since if wood_frac<1 then I can't see how you get grass=0 - but you could conceive cases with a small amount of primn_grass and a single year where an unmodified ptoc and/or ptoq would be enough to hit a limit (e.g. rapid deforestation in the Amazon where CABLE-POP thinks there's little/no grass but LUH2 did)

IMPORTANTLY - all this is for the future (so not a blocker to this PR)

har917: After further thought - this is less of an issue (possibly even a complete non-issue) that I first thought.  The trick here is to remember that (since LUH2 is internal consistent) you can't have the case above, you have to have some LUH2 state available to transition away from.

If we have ptog = A >0, ptoq = A1 >0, ptoc = A2 >0 (A=A1+A2) with no other transitions and 0<WF<1 then that implies that
- LUH2 primary at start is at least A
- POPLUC PF at start  PF0>=WF A and POPLUC PG at start PG0>= A (1-WF)
- time stepping then givesPF -> PF0-WF A>=0, PG -> PG0-(1-WF)*A >=0
- GR = PG0 + CR0 + PA0 -> PG0 + CR0 + PA0 - (1-WF)*A + A1 + A2 = GR0 + WF*A >0
- with CR->CR0+A1 (always < GR, since PG0-(1-WF)*A>0, PA0>0 & A2>0)
- and PA-> PA0+A2 (always < GR-CR)

I haven't followed all the other cases through but I think it's okay

***EXCEPT*** for the awkward case of transitions from primary to rangeland and then to other types (see next)

juergenknauer: OK. Good that it seems to be Ok for most cases. Primaryn to rangeland and then to let's say cropland should be OK for most cases. Because now we account for rtoc conversions that is not multiplied by woodfrac. For example in North Australia we currently have 40% trees, 60% grass. If we have 5% rtoc this will turn 5% of that grass into crops. If we have 65% rtoc we turn 60% grass into crops, but not more. That would be one of the edge cases where our unknown (and in that case incorrect) woodfrac does not allow us to get the right LU areas.

ccarouge:
@juergenknauer Do you know the commit hash of the version you have used for TRENDY v13? So we can go ahead with the tag. Sorry I missed this comment last week.

juergenknauer:
Hi @ccarouge , it's most likely the most recent commit, but I will give you the final confirmation this afternoon. 

Another question: I previous years I always added the svn revision number to the header of the netcdf output files. So that users could always link it to the right version of the code. Can we do something similar with git?

juergenknauer:
Hi @ccarouge , this is the commit we can tag for TRENDY v13: e190b23 
This is the latest commit on the CABLE-POP_TRENDY branch.

thanks for tagging!

PR #377, opened 2024-08-29
Add routine to search for dimension IDs by name.

Whyborn:
# CABLE

## Description

The retrieving of dimension IDs from input NetCDF files is inconsistent across the code. Add routines to the ```cable_common``` module to standardise the retrieving of said IDs by searching through a list of possible names for the dimension.

Fixes #(issue)

## Type of change

- [x] Feature

## Checklist

### DIFF
--- a/core/biogeophys/cable_common.F90
+++ b/core/biogeophys/cable_common.F90
@@ -24,7 +24,7 @@
 
 MODULE cable_common_module
 
-  USE NetCDF,   only: NF90_INQ_DIMID, NF90_NOERR, NF90_EDIMMETA
+  USE NetCDF,   only: NF90_INQ_DIMID, NF90_NOERR, NF90_BADDIM
   IMPLICIT NONE
 
   !---allows reference to "gl"obal timestep in run (from atm_step)
@@ -709,7 +709,7 @@ CONTAINS
 
     ! Iterator and status checker- use NF90_EDIMMETA so we throw error when an
     ! empty list of names is passed.
-    INTEGER :: i, ok = NF90_EDIMMETA
+    INTEGER :: i, ok = NF90_EBADDIM
 
     ! Run through the possible names until we find the correct one
     CheckNames: DO i = 1, SIZE(DimNames)
--- a/core/biogeophys/cable_common.F90
+++ b/core/biogeophys/cable_common.F90
@@ -24,7 +24,7 @@
 
 MODULE cable_common_module
 
-  USE NetCDF,   only: NF90_INQ_DIMID, NF90_NOERR
+  USE NetCDF,   only: NF90_INQ_DIMID, NF90_NOERR, NF90_EDIMMETA
   IMPLICIT NONE
 
   !---allows reference to "gl"obal timestep in run (from atm_step)
@@ -707,9 +707,9 @@ CONTAINS
     INTEGER :: FileID, get_dimid
     CHARACTER(LEN=*), DIMENSION(:)  :: DimNames
 
-    ! Iterator and status checker- use NF90_EINVAL so we throw error when an
+    ! Iterator and status checker- use NF90_EDIMMETA so we throw error when an
     ! empty list of names is passed.
-    INTEGER :: i, ok = NF90_EINVAL
+    INTEGER :: i, ok = NF90_EDIMMETA
 
     ! Run through the possible names until we find the correct one
     CheckNames: DO i = 1, SIZE(DimNames)
--- a/core/biogeophys/cable_common.F90
+++ b/core/biogeophys/cable_common.F90
@@ -33,12 +33,12 @@ MODULE cable_common_module
   INTEGER, SAVE :: CurYear  ! current year of multiannual run
 
   ! Dimension name lists for searching NetCDF dimensions IDs
-  CHARACTER(LEN=16), DIMENSION(5) :: LatNames = &
-    (/"latitude", "lat", "lats", "y", "Latitude"/)
-  CHARACTER(LEN=16), DIMENSION(5) :: LonNames = &
-    (/"longitude", "lon", "lons", "x", "Longitude"/)
-  CHARACTER(LEN=16), DIMENSION(3) :: TimeNames = &
-    (/"time", "t", "Time"/)
+  CHARACTER(LEN=16), DIMENSION(5), PARAMETER :: LatNames = &
+    ["latitude", "lat", "lats", "y", "Latitude"]
+  CHARACTER(LEN=16), DIMENSION(5), PARAMETER :: LonNames = &
+    ["longitude", "lon", "lons", "x", "Longitude"]
+  CHARACTER(LEN=16), DIMENSION(3), PARAMETER :: TimeNames = &
+    ["time", "t", "Time"]
 
   ! user switches turned on/off by the user thru namelists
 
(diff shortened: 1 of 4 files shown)


core/biogeophys/cable_common.F90
```diff
@@ -684,6 +685,105 @@ subroutine get_unit(iunit)
 
   end subroutine get_unit
 
+  FUNCTION latitude_dimid(FileID)
+    !*## Purpose
+    !
+    ! Find the id of the latitude dimension
+    !
```
abhaasgoyal, line 692: Wanted to suggest an alternative design that would reduce repeats. You could read the dimensions in a functional manner across different variables. For example, the pseudocode would look like:

```fortran
SUBROUTINE read_dimid(FileID, possible_names)
    INTEGER :: FileID
    CHARACTER(LEN=*), INTENT(IN) :: possible_names
    ! Define other vars
    ! Run through the possible names until we find the correct one
    ! We take  possible_names[0] as default param name
    handle_error(ok, f"Failed to find any {possible_names(0)} from the list")
```
Then add [arrays of strings](https://fortran-lang.org/learn/quickstart/arrays_strings/#array-of-strings) as parameters. For example

```
  CHARACTER(LEN=10), DIMENSION(5), PARAMETER :: LatNames
  LatNames = ["y", "latitude", "lat", "lats", "Latitude"]
```

offline/cable_parameters.F90
```diff
@@ -219,16 +219,14 @@ SUBROUTINE read_gridinfo(nlon, nlat, npatch)
     ok = NF90_OPEN(trim(filename%type), 0, ncid)
     IF (ok /= NF90_NOERR) CALL nc_abort(ok, 'Error opening grid info file.')
 
-    ok = NF90_INQ_DIMID(ncid, 'longitude', xID)
-    IF (ok /= NF90_NOERR) ok = NF90_INQ_DIMID(ncid, 'x', xID)
-    IF (ok /= NF90_NOERR) CALL nc_abort(ok, 'Error inquiring x dimension.')
+    xID = longitude_dimid(ncid)
```
abhaasgoyal, line 222: Use the suggestion function `read_dimid` with `ncid` and possible names defined as parameter in the same module.

core/biogeophys/cable_common.F90
```diff
@@ -684,6 +685,105 @@ subroutine get_unit(iunit)
...
+    !## Method
+    !
+    ! Run through a list of possible names for the latitude dimension, until one
+    ! is found in the NetCDF dimensions.
+
+    INTEGER :: FileID, latitude_dimid
+    INTEGER :: i, ok
```
abhaasgoyal, line 699: It's a small thing, however if the length of `LatNames` is 0, then `ok` is undefined. Could initialize `ok` for higher guarantees

abhaasgoyal (changes requested):
Functionality looks good, however had some suggestions for design changes

core/biogeophys/cable_common.F90
Whyborn: I think I know what you mean- have the ```cable_common``` define ```LatNames, LonNames, TimeNames```, which can be imported by other modules as necessary and passed to the ```read_dimid``` function. I guess we don't lose any clarity either, because the passed variable name makes it clear which dimension we're retrieving.

core/biogeophys/cable_common.F90
Whyborn: Ah yep, I guess makes sense as a safety thing, if when coupled with the prior change, someone else tries to hijack it for their own purposes and pass an invalid array.

Whyborn:
Made some adjustments- now a single routine, with the possible names defined in the ```cable_common``` module. Used the ```NF90_BADDIM``` as the default for ```ok```, since I'm fairly sure that's what the routine throws when the dimension can't be found.

core/biogeophys/cable_common.F90
```diff
@@ -24,13 +24,22 @@
...
   IMPLICIT NONE
 
   !---allows reference to "gl"obal timestep in run (from atm_step)
   !---total number of timesteps, and processing node
   INTEGER, SAVE :: ktau_gl, kend_gl, knode_gl, kwidth_gl
   INTEGER, SAVE :: CurYear  ! current year of multiannual run
 
+  ! Dimension name lists for searching NetCDF dimensions IDs
+  CHARACTER(LEN=16), DIMENSION(5) :: LatNames = &
+    (/"latitude", "lat", "lats", "y", "Latitude"/)
```
abhaasgoyal, line 37: You can use square brackets (Using `(/ x /)` is part of the old standard in Fortran)

core/biogeophys/cable_common.F90
```diff
@@ -684,6 +693,136 @@ subroutine get_unit(iunit)
...
+      END IF
+    END DO CheckNames
+
+    CALL handle_err(ok, "Failed to find any "//TRIM(DimNames(1))//" dimensions&
+      from the list.")
+
+  END FUNCTION get_dimid
+
+  !FUNCTION latitude_dimid(FileID)
+    !!*## Purpose
```
abhaasgoyal, line 728: Not sure why we have commented these out (we can delete these right)?

core/biogeophys/cable_common.F90
abhaasgoyal: Also, it'd be great if you could define it as a parameter / constant (so that it is faster in runtime).

abhaasgoyal (commented):
I noticed more small changes before merging

core/biogeophys/cable_common.F90
abhaasgoyal: Hmm, maybe we could have these defined in a derived type called `PossibleNamesType` / similar, but we want these to be compile-time constants for efficiency as well. If you think it wouldn't make much of a difference (since it would probably be optimized in `Release` builds), then what do you think about going for derived type?

core/biogeophys/cable_common.F90
abhaasgoyal: See comment about [derived types](https://github.com/CABLE-LSM/CABLE/pull/377#discussion_r1760568177) earlier, which would contradict introducing `PARAMETER`. Happy for you to implement any

Whyborn: Is there a good reason to use square brackets over the old standard? Or has Fortran updated their standards?

core/biogeophys/cable_common.F90
Whyborn: Oh yep, they will be removed.

core/biogeophys/cable_common.F90
abhaasgoyal: So I noticed that the current fortran documentation has all `[]` braces. Also, although it may be subjective, the author of the book Modern Fortran preferred to use square braces as well. That makes using `[]` optional, but I used it in `cable_checks`,  so it'd be nice to have a common coding standard.

image

core/biogeophys/cable_common.F90
Whyborn: I don't think something like that is necessary, I don't see any benefit to that over just having an array for each set of names. It lets us use ```PARAMETER``` as well.

core/biogeophys/cable_common.F90
Whyborn: That's a pretty good reason, I'll make that change.

Whyborn:
I've changed to square brackets and made the lists of names ```PARAMETER```s.

abhaasgoyal (approved):
LGTM (can merge with squash)

PR #378, opened 2024-08-30
overwrite this with the version from CABLE-POP-TRENDY

### LINKED ISSUE #251: CABLE-POP_TRENDY refactor: core/biogeophys/cable_climate.F90

JhanSrbinovsky:
# CABLE

## Description

This keeps getting in the way of merging AND I dont think anyone else uses this SUBROUTINE? Hence requesting review by Ian

Fixes #251

## Checklist

- [X ] The new content is accessible and located in the appropriate section.
- [ X] I have checked that links are valid and point to the intended content.
- [ X] I have checked my code/text and corrected any misspellings 
                 - ALTHOUGH this is as it appears in CABLE-POP-TRENDY. 
                 - Tidying and ensuring compliance at this stage would make the move redundant

----
📚 Documentation preview 📚: https://cable--378.org.readthedocs.build/en/378/

### DIFF
--- a/src/science/misc/cable_climate.F90
+++ b/src/science/misc/cable_climate.F90
@@ -15,271 +15,475 @@
 ! Called from: cable_driver
 !
 ! History: Vanessa Haverd Jan 2015
-
+! Aug 2017: additional leaf-level met variables stored for use in optimising ratio
+! of Jmax to Vcmax
+#ifndef UM_CBL
 ! ==============================================================================
+module cable_climate_mod
 
-# ifndef UM_CBL
-MODULE cable_climate_mod
+  use cable_def_types_mod,  only: met_type, climate_type, canopy_type,soil_snow_type, mp, &
+       r_2, radiation_type, veg_parameter_type
+  use TypeDef,              only: i4b, dp
+  use cable_IO_vars_module, only: patch
+  use CABLE_COMMON_MODULE,  only: CurYear, filename, cable_user, HANDLE_ERR
 
-  USE cable_def_types_mod, ONLY: met_type, climate_type, canopy_type, mp, &
-       r_2, alloc_cbm_var, air_type, radiation_type
-  USE TypeDef,              ONLY: i4b, dp
-  USE cable_IO_vars_module, ONLY: patch
-  USE casa_ncdf_module, ONLY: HANDLE_ERR
-  USE CABLE_COMMON_MODULE, ONLY: CurYear, filename, cable_user
+contains
 
-CONTAINS
-  ! ==============================================================================
+  ! ------------------------------------------------------------------
 
+  subroutine cable_climate(ktau, kstart, ktauday, idoy, LOY, &
+       met, climate, canopy, veg, ssnow, rad, &
+       dels, np)
 
-  SUBROUTINE cable_climate(ktau,kstart,kend,ktauday,idoy,LOY,met,climate, canopy, &
-       air, rad, dels, np)
+    implicit none
 
+    integer,                  intent(IN)    :: ktau      ! integration step number
+    integer,                  intent(IN)    :: kstart    ! starting value of ktau
+    integer,                  intent(IN)    :: ktauday
+    integer,                  intent(IN)    :: idoy, LOY ! day of year (1-365) , Length oy
+    type(met_type),           intent(IN)    :: met       ! met input variables
+    type(climate_type),       intent(INOUT) :: climate   ! climate variables
+    type(canopy_type),        intent(IN)    :: canopy    ! vegetation variables
+    type(veg_parameter_type), intent(IN)    :: veg       ! vegetation parameters
+    type(soil_snow_type),     intent(IN)    :: ssnow
+    type(radiation_type),     intent(IN)    :: rad       ! radiation variables
+    real,                     intent(IN)    :: dels      ! integration time setp (s)
+    integer,                  intent(IN)    :: np
 
-    IMPLICIT NONE
+    integer :: d, y
+    integer, parameter :: COLDEST_DAY_NHEMISPHERE = 355
+    integer, parameter :: COLDEST_DAY_SHEMISPHERE = 172
+    real, parameter :: CoeffPT = 1.26
+    real, dimension(mp) :: mtemp_last, phiEq, ppc, EpsA, RhoA
+    integer :: startyear
+    integer :: MonthDays(12)
+    integer :: nmonth, tmp, nsd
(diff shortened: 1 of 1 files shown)


src/science/misc/cable_climate.F90
```diff
@@ -15,271 +15,475 @@
...
-    IF(MOD(ktau,ktauday)==1) THEN
-       climate%dtemp = met%tk - 273.15
-       climate%dmoist = canopy%fwsoil
-    ELSE
-       climate%dtemp = climate%dtemp + met%tk - 273.15
-       climate%dmoist = climate%dmoist + canopy%fwsoil
-    ENDIF
+    if (mod(ktau, ktauday) == 1) then
+       climate%dtemp     = met%tk - 273.15
+       climate%dmoist    = sum(real(ssnow%wb(:,:))*veg%froot(:,:), 2)
```
har917, line 101: I think we need to retain a comment here around the change in science from `%dmoist = %fwsoil` to `%dmoist  = SUM(%wb * %froot, 2)`.  These quantities are related but different.

src/science/misc/cable_climate.F90
```diff
@@ -15,271 +15,475 @@
...
 
+       ! Update GDD counters and chill day count
+       climate%gdd0  = climate%gdd0  + max(0.0, climate%dtemp-0.0)
+       climate%agdd0 = climate%agdd0 + max(0.0, climate%dtemp-0.0)
 
-       ENDIF
+       climate%gdd5  = climate%gdd5  + max(0.0, climate%dtemp-5.0)
+       climate%agdd5 = climate%agdd5 + max(0.0, climate%dtemp-5.0)
 
-       !jhan: see CABLE Ticket#149 and above CABLE_LSM: comment
```
har917, line 146: May need to retain this `#ifndef` condition - I don't know why it was necessary and likely superceded by the  `#ifdef UM_CBL`

src/science/misc/cable_climate.F90
```diff
@@ -330,223 +536,230 @@ SUBROUTINE biome1_pft(climate, np)
...
        ! (17) Ice/polar desert
-       IF(climate%biome(k)==999) THEN
+       if (climate%biome(k)==999) then
           climate%biome(k) = 17
           climate%iveg(k) = 17  !
-       ENDIF
+       endif
 
-       !jhan: see CABLE Ticket#149 and above CABLE_LSM: comment
-# ifndef ESM15
```
har917, line 543: see earlier comment about this `#ifndef`

src/science/misc/cable_climate.F90
JhanSrbinovsky: such as?

OK before we get too involved, I am. totally unfamiliar with this module - who uses cable_climate?

src/science/misc/cable_climate.F90
```diff
@@ -559,544 +772,171 @@ SUBROUTINE biome1_pft(climate, np)
...
+
+    !if (.not.cable_user%climate_fromzero) then
+    !   CALL READ_CLIMATE_RESTART_NC (climate, ktauday)
     !endif
 
-  END SUBROUTINE climate_init
-
-  ! ==============================================================================
-
-  SUBROUTINE WRITE_CLIMATE_RESTART_NC ( climate, ktauday )
```
har917, line 634: Where have the `write_climate_restart_nc` and `read_climate_restart_nc` subroutines/capability gone?  Ah-ha possibly into `write_netcdf_cbm_var()` which is in `cable_def_types_mod`.

So this PR may need to deal with that MODULE as well.

src/science/misc/cable_climate.F90
har917: As far as I know only CABLE-POP simulations actively use this routine - so that's the TRENDY and BIOS applications - however the `climate%` is woven throughout the offline code.

I'm not actually sure how CM2 and AM3 are handling this to be honest - the TYPE is there (if incomplete) and there is an allocation/initialisation.  However there is no call to the `cable_climate()` subroutine yet there is a use (in AM3) in `casa_rplant` - I would have thought that would have caused some 'interesting' results or a crash.

har917: To answer the implicit question - all we need to do is include a comment such as
`! Has been updated from %dmoist  = canopy%fwsoil `  with, if possible, a svn revision number.

har917: > there is a use (in AM3) in casa_rplant and `cbl_dryLeaf`

Specifically `casa_rplant()` and `cbl_dryLeaf` use `climate%qtemp_max_last_year` - this is initialised to 0.0.  HOwever both uses are inside a IF `cable_user%CALL_climate` condition so will only be active if triggered by the user.

src/science/misc/cable_climate.F90
```diff
@@ -15,271 +15,475 @@
 ! Called from: cable_driver
 !
 ! History: Vanessa Haverd Jan 2015
-
+! Aug 2017: additional leaf-level met variables stored for use in optimising ratio
+! of Jmax to Vcmax
+#ifndef UM_CBL
```
har917, line 20: mmm - not sure I really like this in the code base at least not without the corresponding #ifdef case

har917 (commented):
I think this is okay to go in with some key notes:
1. Care will be needed around the `#ifdef`s
2. I suspect that this will need partner updates to cable_def_types_mod() to work. 
3. this is one of the routines that will need to be rationalised as part of CABLE-4 - i.e. this PR should be viewed only as a temporary fix.

@ccarouge in light of the above - are you happy that @JhanSrbinovsky goes ahead and merges into MAIN (assuming that any technical quirks can be overcome)?

JhanSrbinovsky:
the ifndef um_cbl was because it is NEVER used in any UM application and there were corresponding things to bring in at build time which would have complicated things for a subroutine that isnt even called. That said - the version from CABLE-POP is even worse. They've added things to climate% TYPE which is I suppose expected. They've even added elements to met% TYPE.  

Online, climate% TYPE is declared etc, they are just never used in calculation of anything (well they shouldn't be anyway, I hope this is the case).


There will be more work needed to even get this to build properly! I would be tempted to just abandon this altogether, except that it needs to go in if indeed there is to be a CABLE4 code base. 

At the very least there will have to be things "taken out" (or reviewed if they are even necessary) that apply to reading and writing netcdf files.

In summary, 

- work needs to be done for this cable_climate to even be substituted to offline CABLE, including for applications where it is not called.  
- further thinking into what role this would play in an online CABLE4 is required. 

Close this PR until these issues are resolved

PR #410, opened 2024-09-27
Fix gfortran compilation and some warnings

micaeljtoliveira:
# CABLE

## Description

This PR fixes compilation with gfortran. I also fixes several compiler warnings (mostly about unused variables) in the MPI master/driver modules.

## Type of change

- [X] Bug fix

## Checklist

----
📚 Documentation preview 📚: https://cable--410.org.readthedocs.build/en/410/

### DIFF
--- a/src/offline/cable_mpidrv.F90
+++ b/src/offline/cable_mpidrv.F90
@@ -24,8 +24,7 @@ PROGRAM mpi_driver
   USE cable_mpicommon
   USE cable_mpimaster
   USE cable_mpiworker
-  USE cable_namelist_util, ONLY: get_namelist_file_name,&
-       CABLE_NAMELIST,arg_not_namelist
+  USE cable_namelist_util, ONLY: get_namelist_file_name
 
   IMPLICIT NONE
 
--- a/src/offline/cable_mpimaster.F90
+++ b/src/offline/cable_mpimaster.F90
@@ -158,19 +158,19 @@ CONTAINS
     USE cable_IO_vars_module, ONLY: logn,gswpfile,ncciy,leaps,globalMetfile, &
          verbose, fixedCO2,output,check,patchout,    &
          patch_type,landpt,soilparmnew,&
-         defaultLAI, sdoy, smoy, syear, timeunits, exists, output, &
-         latitude,longitude, calendar, set_group_output_values
+         timeunits, exists, output, &
+         calendar, set_group_output_values
     USE cable_common_module,  ONLY: ktau_gl, kend_gl, knode_gl, cable_user,     &
-         cable_runtime, fileName, myhome,            &
+         cable_runtime, fileName,            &
          redistrb, wiltParam, satuParam, CurYear,    &
          IS_LEAPYEAR, calcsoilalbedo,                &
          kwidth_gl, gw_params
-  USE casa_ncdf_module, ONLY: is_casa_time
-! physical constants
-USE cable_phys_constants_mod, ONLY : CTFRZ   => TFRZ
-USE cable_phys_constants_mod, ONLY : CEMLEAF => EMLEAF
-USE cable_phys_constants_mod, ONLY : CEMSOIL => EMSOIL
-USE cable_phys_constants_mod, ONLY : CSBOLTZ => SBOLTZ
+    USE casa_ncdf_module, ONLY: is_casa_time
+    ! physical constants
+    USE cable_phys_constants_mod, ONLY : CTFRZ   => TFRZ
+    USE cable_phys_constants_mod, ONLY : CEMLEAF => EMLEAF
+    USE cable_phys_constants_mod, ONLY : CEMSOIL => EMSOIL
+    USE cable_phys_constants_mod, ONLY : CSBOLTZ => SBOLTZ
     USE cable_input_module,   ONLY: open_met_file,load_parameters,              &
          get_met_data,close_met_file
     USE cable_output_module,  ONLY: create_restart,open_output_file,            &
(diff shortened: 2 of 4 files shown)


SeanBryan51:
@micaeljtoliveira I was able to build your `gcc_compilation_and_warnings` branch on Gadi with

```
./build.bash -C gnu --mpi -DCMAKE_Fortran_FLAGS="-ffree-line-length-none"
```

However the compiled executable crashes with segmentation fault (for both MPI and serial executables). This is probably outside the scope of this pull request but I will document this here for now.

For the serial case I ran the [example serial configuration in the CABLE repository](https://cable.readthedocs.io/en/latest/user_guide/uber_quick_guide/#in-slightly-more-detail):

```
cd src/offline/
../../bin/cable cable.nml
```

which outputs:

```
 THE NAME LIST IS cable.nml
 Use spatially-specific soil properties;          360         150
           0
           0
 When choosing spatially-specific soil properties,
 snow-free albedo is also overwritten by this data set.
 Reading CABLE_PFTPARM namelist...
 Reading CABLE_SOILPARM namelist...
 Total number of patches (countPatch):            1
 iveg           1           2
 patchfrac           1   1.00000000
  Could neither find restart file ./
  nor ./
  Pre-loaded default initialisations are used.
  Loaded some parameters from met input file: TumbaFluxnet.1.3_met.nc the rest are default values - check log file
 Reading CABLE_PFTPARM namelist...
 Reading CABLE_SOILPARM namelist...

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7fc420b985af in ???
#1  0x47e4da in ???
#2  0x40d6f8 in ???
#3  0x40321c in ???
#4  0x7fc420b847e4 in ???
#5  0x40325d in ???
#6  0xffffffffffffffff in ???
Segmentation fault
```

I was able to trace the segmentation fault to the [`acc_out_var`](https://github.com/CABLE-LSM/CABLE/blob/8f62897b6ff6d283565d5df374271425dbce93c7/src/offline/cable_output.F90#L2041-L2056) using the DDT debugger (requires setting [`CABLE_GNU_Fortran_FLAGS_DEBUG`](https://github.com/CABLE-LSM/CABLE/blob/8f62897b6ff6d283565d5df374271425dbce93c7/CMakeLists.txt#L30) to `-g -O0` and building in the debug configuration).

For the MPI case I ran the [crujra_accessN96_1h](https://github.com/CABLE-LSM/crujra_accessN96_1h) config which crashes similarly to the serial case.

Building your branch with the Intel compiler and running the executable seems to work fine so it looks like this is specific to the gfortran build.

SeanBryan51 (commented):
The debug configuration for `gfortran` has some pretty pedantic flags enabled and currently doesn't build on Gadi - is it worth ironing these out?

SeanBryan51:
@mcuntz I recall I used your compiler flags when adding the GNU compiler. The debug flags currently contain `-O` (equivalent to `-O1`) which enables some optimisations and prevents me from using the DDT debugger. Is there a reason why you use `-O` instead of `-O0` in the debug flags for gfortran?

mcuntz:
@SeanBryan51 -O is there simply for speed even in debug mode. I basically debug with print statements and do not use much the debuggers. No worry to change it.

However, I think that the compilation should work with _the pretty pedantic flags_. If it works with another compiler means that that compiler assumes something and it is not always what you'd assume if you had coded _properly_. So I would be 100% for ironing out the places where it is not compiling.

micaeljtoliveira:
@SeanBryan51 I'm able to reproduce your error. I've compiled cable with the `-fbacktrace` option (plus a few others), so I get a more useful error message:
```
At line 2908 of file /home/micael/workspaces/cable/git/src/offline/cable_input.F90
Fortran runtime warning: An array temporary was created
  Loaded some parameters from met input file: TumbaFluxnet.1.3_met.nc the rest are default values - check log file
 Reading CABLE_PFTPARM namelist...
 Reading CABLE_SOILPARM namelist...
At line 2194 of file /home/micael/workspaces/cable/git/src/offline/cable_output.F90
Fortran runtime error: Array bound mismatch for dimension 2 of array 'acc_val' (6/1)

Error termination. Backtrace:
#0  0x4745bc in generate_out_write_acc_d2
        at /home/micael/workspaces/cable/git/src/offline/cable_output.F90:2194
#1  0x477303 in __cable_output_module_MOD_write_output
        at /home/micael/workspaces/cable/git/src/offline/cable_output.F90:1769
#2  0x7ab84b in cable_offline_driver
        at /home/micael/workspaces/cable/git/src/offline/cable_driver.F90:942
#3  0x40379c in main
        at /home/micael/workspaces/cable/git/src/offline/cable_driver.F90:62
```

micaeljtoliveira:
Okay, I think I understand the problem, but I'm not so sure how you to fix it.

In the following call in `cable_output.F90`:

```fortran
CALL generate_out_write_acc(output%SoilMoist, ovid%SoilMoist, 'SoilMoist', out%SoilMoist, REAL(ssnow%wb, 4), ranges%SoilMoist, patchout%SoilMoistIce, out_settings)
```

`output%SoilMoist` is set to false and the `out%SoilMoist` and ` ssnow%wb` arguments are passed down to the `elemental` function `generate_out_write_acc_d2`. At that point, the compiler checks the bounds of those arguments, as it needs to be sure they are compatible, as the function needs to be called element-wise. But `output%SoilMoist` is actually never allocated (because `output%SoilMoist` is false) and the check fails.

There are basically two solutions:
1. Do not check array bounds are run-time.
2. Exit `generate_out_write_acc` when argument `output_var` is false **before** calling `acc_out_var`. 

I would recommend against option 1., as the array bounds check is actually very useful when debugging.

micaeljtoliveira:
Actually, even with the checks disabled, it fails (as I should have realized from the segfault message obtained by @SeanBryan51).

This means one really cannot call an elemental function over an array that has not been allocated, which makes total sense. I would thus argue that in this case gfortran is actually doing the right thing and it's the Intel compiler that is misbehaving.

micaeljtoliveira:
There's a few more problematic cases in the same module were several non-allocated arrays are being accessed. For example, here
```
CALL generate_out_write_acc(output%casa, ovid%PlantTurnover, 'PlantTurnover', out%PlantTurnover, &
                                REAL((SUM(casaflux%Cplant_turnover, 2))/86400.0/1.201E-5, 4), ranges%NEE, patchout%PlantTurnover, out_settings)
```
it might happen that `casaflux%Cplant_turnover` is not allocated, causing a segfault.

I think the only sensible solution is to fence all calls to `generate_out_write_acc` with an if statement. Note that this issue was introduced in #287. @abhaasgoyal FYI.

SeanBryan51:
@micaeljtoliveira thanks for looking into it.

> I think the only sensible solution is to fence all calls to generate_out_write_acc with an if statement.

This seems appropriate to me. I'm happy to let you decide if you want to fix this in this pull request.

micaeljtoliveira:
> This seems appropriate to me. I'm happy to let you decide if you want to fix this in this pull request.

I think I rather do it in a separate PR.

SeanBryan51:
@micaeljtoliveira FYI the build-ci action has been fixed (https://github.com/CABLE-LSM/CABLE/pull/409) - if you rebase off main you should get a green light on the CI checks

SeanBryan51 (approved):
Approving - other issues related to the gfortran build can be addressed at a later point

PR #422, opened 2024-10-16
Removed ESM15 version with hard-wired xfpleach parameter, #278

### LINKED ISSUE #278: ESM15 reconciliation - casa_delsoil
Lines 1339-1345 of casa_cnp.F90 (subroutine casa_delsoil) have alternate definitions of casaflux%Pleach depending on the ESM15 ifdef.

```
#           ifdef ESM15  
             casaflux%Pleach(nland)  =  (1.0e-4) &
                                              * max(0.0,casapool%Psoillab(nland))
#           else
             casaflux%Pleach(nland)  =  casaflux%fPleach(nland) &
                  * MAX(0.0,casapool%Psoillab(nland))
#            endif
```

ESM15 effectively hard-wires casaflux%fPleach to be 1.0e-4. For non-ESM15 cases, fPleach is set in read_biome with the value (dependent on soil order) coming from the pftlookup file. As long as the pftlookup file used for ESM1.5 has the value of 1.0e-4 for xfpleach, the model behaviour should be the same and the IFDEF could be removed.

A short test run should be done to confirm this.

rml599gh:
# CABLE

## Description

In casa_delsoil the evaluation of casaflux%Pleach had two versions, one specific to ESM1.5 (selected via an ifdef) and the alternat for all other cases. Following the code through, the ESM1.5 version had a hard-wired value (1.e-4) for casaflux%fPleach. casaflux%fPleach is filled from xfPleach which is read in from the pftlookup file. The ESM1.5 specific code can be safely removed with no change to the output if xfPleach is set to 1.e-4 in the pftlookup file. Note that there are versions of the pftlookup with xfPleach = 5.e-4. I am not aware of when/who set it to this value and whether it is considered to be better/more realistic.

Fixes #278

## Type of change

- [X] Code clean-up

## Checklist

----
📚 Documentation preview 📚: https://cable--422.org.readthedocs.build/en/422/

### DIFF
--- a/src/science/casa-cnp/casa_cnp.F90
+++ b/src/science/casa-cnp/casa_cnp.F90
@@ -1341,13 +1341,11 @@ END SUBROUTINE casa_delplant
                   +casaflux%Psimm(nland)
              ! net mineralization
 
-#           ifdef ESM15  
-             casaflux%Pleach(nland)  =  (1.0e-4) &
-                                              * max(0.0,casapool%Psoillab(nland))
-#           else
+             !rml 14/10/24 #278 remove ESM15 specific version as can be 
+             !accommodated by setting appropriate parameter in pftlookup
              casaflux%Pleach(nland)  =  casaflux%fPleach(nland) &
                   * MAX(0.0,casapool%Psoillab(nland))
-#            endif
+
              DO k=1,msoil
                 DO j=1,mlitter
                    casaflux%FluxPtosoil(nland,k) =  casaflux%FluxPtosoil(nland,k)  &
--- a/documentation/docs/user_guide/inputs/pftlookup_csv.md
+++ b/documentation/docs/user_guide/inputs/pftlookup_csv.md
@@ -107,7 +107,7 @@ on soil, vegetation carbon and nutrients dynamics.
 | fpptoL(frt)    | `casabiome%ftransPPtoL(nv,froot)`             | Flux factor of root phosphorus to litter pools \( (-) \) |
 | xkmlabp        | `xkmlabp(iso)`                                | Phosphorus absorption \( (gP \cdot m^{-2}) \)  |
 | xpsorbmax      | `xpsorbmax(iso)`                              | Maximum phosphorus absorption \( (gP \cdot m^{-2}) \) |
-| xfpleach       | `xfPleach(iso)`                               | Phosphorus leaching \( (-) \) |
+| xfpleach       | `xfPleach(iso)`                               | Phosphorus leaching \( (-) \) (hard-wired in the code for CABLE2.4/ACCESS-ESM1.5 at 1.e-4 independent of pft) |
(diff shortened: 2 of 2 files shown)


rml599gh:
This change is only relevant for ESM1.5/ESM1.6.
I have tested the change using 3 one month tests.
1. Current code with ifdef ESM15
2. Code with ESM15 specific code removed and standard pftlookup including xfPleach=5.e-4
3. Code with ESM15 specific code removed and pftlookup changed so xfPleach=1.e-4

Run 3 should give the same results as Run 1. Run 2 should be different.
casaflux%Pleach is used in the calculation of dpSoillabdt which is then used for update Psoillab and Psoilsorb, the labile and sorbed phosphorus pools (stash code 880 and 881).
The figure shows the difference in the global mean labile and sorbed phosphorus pool at the end of January between the original (case 1) and cases 2 (black, green) and 3 (red, blue). The difference is shown as a proportion of the original global mean. The y axis is pft and the global means have been calculated without tile fraction weighting. As expected, case 3 shows no difference from case 1 (the red line is under the blue line). The change in parameter value makes a very small difference in the pool size though this would presumably accumulate over time and tend to a slightly different steady state to previously.

I am comfortable that the ifdef ESM15 and associated code can be safely removed.

image

src/science/casa-cnp/casa_cnp.F90
```diff
@@ -1341,13 +1341,11 @@ SUBROUTINE casa_delsoil(veg,casapool,casaflux,casamet,casabiome)
                   +casaflux%Psimm(nland)
              ! net mineralization
 
-#           ifdef ESM15  
-             casaflux%Pleach(nland)  =  (1.0e-4) &
-                                              * max(0.0,casapool%Psoillab(nland))
-#           else
+             !rml 14/10/24 #278 remove ESM15 specific version as can be 
+             !accommodated by setting appropriate parameter in pftlookup
```
abhaasgoyal, line 1345: Instead of providing a comment here, is there any documentation on pftlookup versioning (similar to [bench_example](https://github.com/CABLE-LSM/bench_example/tree/main/namelists) repo?

abhaasgoyal (commented):
Looks good to me (although had a query on whether the comment could be put in a better place)

src/science/casa-cnp/casa_cnp.F90
rml599gh: I'm not aware of any documentation on pftlookup versioning. It might be good to add a comment to the appropriate row of the table in this page of the userguide, https://cable.readthedocs.io/en/latest/user_guide/inputs/pftlookup_csv/ to note that xfPleach was hard-wired in CABLE2.4 for ACCESS-ESM1.5. I haven't made changes to the documentation yet so will need to work out how to do that.

rml599gh: I've now made the change to the documentation that I suggested. I'd prefer to also keep the comment in the code as that allows anyone to refer back to the issue.

rml599gh:
@abhaasgoyal - are you OK to approve this now with the documentation addition noted above?

abhaasgoyal (approved):

PR #427, opened 2024-10-18
Correct array constructor for strings in CABLE-POP_TRENDY

mcuntz:
# CABLE

## Description

The model did not compile with gfortran v14.2.

Possible dimension names in input files are defined as a character arrays in cable_common_module. They have different character lengths. The current notation is not Fortran standard compliant. I replaced
```
CHARACTER(LEN=16), DIMENSION(5), PARAMETER :: LatNames = &
    ["latitude", "lat", "lats", "y", "Latitude"]

```
with
```
CHARACTER(LEN=*), DIMENSION(*), PARAMETER :: LatNames = &
    [CHARACTER(LEN=8) :: "latitude", "lat", "lats", "y", "Latitude"]

```
Same for `LonNames` and `TimeNames`.
I found this very cool notation on [stackoverflow](https://stackoverflow.com/questions/21552430/gfortran-does-not-allow-character-arrays-with-varying-component-lengths).

Another error was also in cable_common, where the continuation character was within the string:
```
CALL handle_err(ok, "Failed to find any "//TRIM(DimNames(1))//" dimensions&
    from the list.")

```
The compiler does not know if `&` is part of the string or the continuation character. I pulled it out of the string.

I also extended VarNames in spatio_temporal_dataset from 16 to length 32 because they are used currently only in cable_cru_Trendy and the names used there had a maximum of 32 characters.

I finally compiled with debug flags, made both arguments of Fortran intrinsics (min) the same kind, removed all unused variables and unused module variables mainly in cable_input and casa_inout.

I would suggest adding '-e03' to the Intel debug flags for better standard compliance. I have not tested that and have not changed CMakeLists.txt.

## Type of change

- [x] Bug fix

## Checklist

### DIFF
--- a/core/biogeochem/POPLUC.F90
+++ b/core/biogeochem/POPLUC.F90
@@ -1459,14 +1459,14 @@ CONTAINS
              irp = ilu + j -1
              ! Update tile area
              if (ilu == p) then
-                POPLUC%primf(g) = min(1.0, max(patch(irp)%frac + dA_r(ilu) + dA_d(ilu), 0.0_dp))
+                POPLUC%primf(g) = min(1.0_dp, max(patch(irp)%frac + dA_r(ilu) + dA_d(ilu), 0.0_dp))
              elseif (ilu == s) then
-                POPLUC%secdf(g) = min(1.0, max(patch(irp)%frac + dA_r(ilu) + dA_d(ilu), 0.0_dp))
+                POPLUC%secdf(g) = min(1.0_dp, max(patch(irp)%frac + dA_r(ilu) + dA_d(ilu), 0.0_dp))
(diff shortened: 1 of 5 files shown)


Whyborn:
Thanks Mathias, I'll pass on the -e03 recommendation to the team.

Whyborn (approved):

Whyborn:
Hi @mcuntz, I have merged this branch into the POP_TRENDY branch. Normally we get the author of the PR to do the merge, in case any last minute changes need to be made. Given the PR looked stable and would interact with some cleaning up of the Met routines that I'm doing, I executed the merge this time.

mcuntz:
Thanks @Whyborn. I somehow did not realise that I should do it, but sounds logical.

PR #434, opened 2024-10-25
Remove ESM15 special case related to lnonwood condition on dNplantdt(wood)

### LINKED ISSUE #423: ESM1.5 reconciliation: lnonwood condition
In subroutine  `casa_delplant` in casa_cnp.F90, the calculation of `casapool%dNplantdt(npt,wood)` is conditional on `casamet%lnonwood(npt)=0` unless it is compiled with `#define ESM15 YES`. 
Presumably the condition was added sometime after CABLE2.4. 
It was in CABLE2.5 but as an IF (lnonwood) else dNplantdt = 0. In the current code, dNplantdt is set to zero for all cases and then the lnonwood condition overwrites dNplantdt for woody pfts. 
The dNplantdt calculation includes `casaflux%kplant` which is `deltcasa/woodage`. woodage is read from the pftlookup file. If it was set to zero for non-woody pfts, this would give a divide by zero, so perhaps the condition was added to address this. If so, there is an equivalent calculation for `casapool%dpPplantdt(npt,wood)` which does not have a similar condition applied.

I don't think there should be any problem with ESM1.5/ESM1.6 including the condition, allowing #ifndef ESM15 to be removed. There should be no difference to the simulation. However a quick test did produce differences after 1 month so further testing and/or an alternative solution is required.

rml599gh:
# CABLE

## Description

Non-ESM15 code versions had a condition on the calculation of dNplantdt(:.wood) to only do the calculation for woody pfts, having first set it to zero for all pfts. An ifndef was used to ignore the condition when ESM15 was defined YES.
The was no apparent reason why ESM15 could not use the condition so the special case was removed.
For clarity the condition was re-written as an IF/THEN.
dPplantdt(:,wood) is calculated in the same way as dNplantdt but had no condition. For consistency, a similar condition has been added so that nitrogen and phosphorus are treated the same way.

Fixes #423

## Type of change

- [X ] Bug fix / code clean up

## Checklist

- [X ] I have checked my code/text and corrected any misspellings

----
📚 Documentation preview 📚: https://cable--434.org.readthedocs.build/en/434/

### DIFF
--- a/src/science/casa-cnp/casa_cnp.F90
+++ b/src/science/casa-cnp/casa_cnp.F90
@@ -1,4 +1,3 @@
-!#define ESM15 YES
 !==============================================================================
 ! This source code is part of the
 ! Australian Community Atmosphere Biosphere Land Exchange (CABLE) model.
@@ -1027,14 +1026,15 @@ DO npt=1,mp
                                       * casabiome%ftransNPtoL(veg%iveg(npt),leaf)
       ENDIF
 
-      casapool%dNplantdt(npt,wood) = 0.0
-#ifndef ESM15
-      ! offline/trunk uses this condition
-      IF (casamet%lnonwood(npt)==0)                                            & 
-#endif
-      casapool%dNplantdt(npt,wood) = - casaflux%kplant(npt,wood)             &
+      !R. Law 25/10/24 removed ESM15 case as no need to exclude the condition
+      !that is in offline/trunk. Also re-write as IF / THEN
+      IF (casamet%lnonwood(npt)==0) THEN                                          
+        casapool%dNplantdt(npt,wood) = - casaflux%kplant(npt,wood)             &
                                      * casapool%Nplant(npt,wood)               &
                                      * casabiome%ftransNPtoL(veg%iveg(npt),wood)
+      ELSE
+        casapool%dNplantdt(npt,wood) = 0.0
+      ENDIF
 
       casapool%dNplantdt(npt,froot) = - casaflux%kplant(npt,froot)             &
                                     * casapool%Nplant(npt,froot)               &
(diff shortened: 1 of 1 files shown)


rml599gh:
Testing consisted of 1 month ESM1.5 runs under piControl conditions.

Initial tests were difficult to interpret as the simulations drifted apart after 2-3 days. Turning off the LAI and vcmax feedbacks gave much clearer results as changes in the carbon cycle did not feedback to the climate.
Results from three tests are noted here.
423b: Code from main i.e. no change
423c: Removal of ESM15 case
423d: Change to IF/THEN for dNplantdt. Addition of condition for dPplantdt

This table shows global maximum (3 columns) and global average for Nitrogen wood pool at end of January for 423b, 423c, 423d.
image

As expected, woody pfts show no change across cases. Non-woody pfts are zero (not sure about bare ground but no change across tests).

However non-woody pfts are not exactly zero.
image
image
There are differences between 423b and 423c, showing the impact of including the condition. There is no difference between 423c and 423d since this is just a change in how the nitrogen code is written.

Similar results are seen for the Phosphorus wood pool. No difference for woody pfts between the tests.
For non-woody:
image
image
423b and 423c are the same. 423d is different due to the addition of the lnonwood condition for dPplantdt.

src/science/casa-cnp/casa_cnp.F90
```diff
@@ -1027,14 +1026,15 @@ SUBROUTINE casa_delplant(veg,casabiome,casapool,casaflux,casamet,            &
...
       ENDIF
 
-      casapool%dNplantdt(npt,wood) = 0.0
-#ifndef ESM15
-      ! offline/trunk uses this condition
-      IF (casamet%lnonwood(npt)==0)                                            & 
-#endif
-      casapool%dNplantdt(npt,wood) = - casaflux%kplant(npt,wood)             &
+      !R. Law 25/10/24 removed ESM15 case as no need to exclude the condition
+      !that is in offline/trunk. Also re-write as IF / THEN
```
SeanBryan51, line 1030: Just a general comment: I would write this in the commit message rather than the actual source code as a commit message should explain why you made the change. I try to reserve code comments for explaining why the code does what it does to help future developers looking at the code (e.g. why we check for `casamet%lnonwood(npt)==0` when setting `casapool%dNplantdt(npt,wood)`).

SeanBryan51 (approved):
I can't really comment on the differences you are getting in the model output but from a technical view it looks good to me. Just on the test cases you mentioned:

> 423b: Code from main i.e. no change
> 423c: Removal of ESM15 case
> 423d: Change to IF/THEN for dNplantdt. Addition of condition for dPplantdt

It would be great if each of these changes (423c and 423d) are separate commits so the differences between each case is visible to the reviewer.

src/science/casa-cnp/casa_cnp.F90
rml599gh: Thanks @SeanBryan51 for the feedback - I will keep this in mind for future code comments/commit messages. One of the challenges in this case was that there was no history as to when and why the condition had been added.

PR #438, opened 2024-10-29
431 add routine for more informative namelist errors

Whyborn:
# CABLE

## Description

Add a routine to handle namelist errors, that returns a much more informative error message than the baseline namelist error message. Utilises the new ```IOMSG``` specifier and the routine detailed [here](https://degenerateconic.com/namelist-error-checking.html).

Fixes #(431)

Edit: Turns out that this namelist checking routine relies on undefined behaviour. The problems I was seeing was the result of a "regression" in Intel Fortran 2023, see [this thread](https://community.intel.com/t5/Intel-Fortran-Compiler/Regression-with-quot-backspace-quot-in-2023-0/m-p/1448175). The comment from Steve Lionel in the accepted answer is important:

> The standard says, "If an error condition occurs during execution of an input/output statement, the position of the file becomes indeterminate."

This means that the so-called regression was not really a regression, just a change in already undefined behaviour. This use of ```BACKSPACE``` to return the incriminating line relies on undefined behaviour that may change at any time. Based on this, I decided against including this part of the routine, and just throwing back the ```IOMSG``` returned by the ```READ``` call.

## Type of change

- Quality of life change

## Checklist

### DIFF
--- a/core/biogeophys/cable_common.F90
+++ b/core/biogeophys/cable_common.F90
@@ -694,22 +694,19 @@ CONTAINS
 
   end subroutine get_unit
 
-  SUBROUTINE handle_namelist_iostat(ios, iomessage)
+  SUBROUTINE handle_iostat(ios, iomessage)
     !*## Purpose
     !
-    ! Take the return status of a namelist read and return an informative error
+    ! Take the return status of a read call and return an informative error
     ! message on failure status.
     !
     !## Method
     !
-    ! Use a combination of the new IOMSG specifier introduced in Fortran 2003
-    ! and the namelist checking routine [here]
-    ! (https://degenerateconic.com/namelist-error-checking.html) to throw an
-    ! informative error message when the reading of the namelist fails.
+    ! Use the IOSTAT and the new IOMSG specifier introduced in Fortran 2003
+    ! to throw an informative error message when the read errors.
 
     INTEGER, INTENT(IN) :: ios
     CHARACTER(LEN=*), INTENT(IN) :: ioMessage
-    CHARACTER(LEN=200) :: BadLine
 
     IF (ios /= 0) THEN
       ! Read of namelist failed
@@ -731,7 +728,7 @@ CONTAINS
       STOP ios
     END IF
 
-  END SUBROUTINE handle_namelist_iostat
+  END SUBROUTINE handle_iostat
 
   FUNCTION get_dimid(FileID, DimNames)
     !*## Purpose
--- a/core/biogeophys/cable_common.F90
+++ b/core/biogeophys/cable_common.F90
@@ -694,7 +694,7 @@ CONTAINS
 
   end subroutine get_unit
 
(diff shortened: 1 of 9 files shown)


SeanBryan51 (approved):
Looks good to me. But it's weird how you are getting different behaviour with the different compilers though - I think it would be worth looking into this before merging it in.

Whyborn:
What's even stranger is that I copied the routine into a little standalone routine (see [here])[https://gist.github.com/Whyborn/f01a768bf7ce59c82140710ee9629ab8] and ran it on Gadi, and it worked as desired with a single backspace. The namelist I used with this gist is

_test.nml_
```
&testnml
    var1 = 1
    abadvar = thiswillfail
    var2 = 2
/
```
So it doesn't seem like a compiler difference. A bit stumped? Why does the version in CABLE need 2 ```BACKSPACE``` calls to get the incriminating line?

SeanBryan51:
That's weird. Could it be a compiler flag issue?

Whyborn:
I feel like it must be related to that, but I tried with most of the flags that come with the ```DEBUG``` build I've been using and couldn't replicate the behaviour. I'll put it on the back-burner for now.

Whyborn:
I've added made the changes to all the namelist read calls I found in the CABLE code (ignored stuff in ```old/``` and ```UM/```). We could extend similar changes to all instances of ```READ``` in the code, but I think that might be too far out of scope.

core/biogeophys/cable_common.F90
```diff
@@ -693,12 +694,48 @@ subroutine get_unit(iunit)
...
+      ! relies on undefined behaviour, as the position in a file becomes
+      ! indeterminate if an error condition occurs during input/output (see
+      ! [this
+      ! thread](https://community.intel.com/t5/Intel-Fortran-Compiler/
+      ! Regression-with-quot-backspace-quot-in-2023-0/m-p/1448175)). While
+      ! it may work as desired for most compilers, it is not guaranteed, and
+      ! may change without warning at any time as evidenced by said thread.
+      !BACKSPACE(nmlUnit)
+      !READ(nmlUnit, FMT='(A)') BadLine
+      !WRITE(ERROR_UNIT, FMT='(A)') "Invalid line in namelist: "//TRIM(BadLine)
```
SeanBryan51, line 727:
```suggestion
```

I think it is safe to remove this comment as there will be no intention to uncomment the code in the future. This info would be best place either in the commit logs or the pull request description (which you have done already).

SeanBryan51 (approved):
Looks good. Well done getting to the bottom of that undefined behaviour! I've requested a change to remove the commented code but other than that, all is good 👍

core/biogeophys/cable_common.F90
Whyborn: Yep that was going to be my question, whether to leave this in there for posterity in case someone comes by in future and thinks "I know this other method, we should add that". As long as it's documented somewhere and findable I guess.

Issue #358, opened 2024-07-30
Rewrite of the MPI implementation for CABLE


ccarouge:
Although the current MPI implementation for CABLE is working in the `main` branch, it has been problematic for some time. In particular, developers find it difficult to know when and how to update it. The fact we have 3 drivers (serial, master and worker) makes it hard to keep everything in sync. Additionally, the MPI implementation is currently broken in the CABLE-POP_TRENDY branch.

We now have the opportunity to bring some expertise to completely redesign the implementation so it is easier for users, developers and maintainers of CABLE.

Following a first analysis of the code, the new implementation should:
- use one unique driver for serial and parallel modes
- use parallel I/O.
- modify the layout of the grid cells per processor to only have connected grid cells. This would simplify a future implementation of lateral flow.

har917:
@ccarouge (I think) We need to include an exercise to determine which variables need to be parallelised and which can be left as local.

I'm not sure whether this is part of this task or not, and (if yes) when in the sequence it should happen - and it will be made more complicated by ongoing efforts to refactorize the code and keeping things in sync with ACCESS/JAC.  However I think it would make the effort to recreate the MPI-version simpler (and the code more efficient).  For example - I'm pretty convinced that a lot of the POP-TYPE does not need to be parallelised.

ccarouge:
@har917 There are 2 parts to what you are saying:

1.  The CABLE code sends information between processors more frequently than needed.  
2. The code has way more global variables than needed.

This work will tackle the first aspect since that's inherent to the design of the MPI implementation. We will not touch the second part which is linked to the software design (and bad interpretations of the design by various people) and not the MPI implementation. The second aspect will be tackled when we work on the design of the code itself past CMIP.

micaeljtoliveira:
After further discussions, it was decided to start the refactoring by merging the three drivers (serial, mpi-master and mpi-driver). This will allow to identify the differences between them (eg., which features are not supported by all).

SeanBryan51:
We can now more or less see the differences in functionality between the MPI drivers and the serial driver (aside from MPI communication specifics).

[...]

PR #440, opened 2024-10-29
Add initialisation routine for offline drivers

SeanBryan51:
Whilst looking into merging the MPI master and worker drivers (#358), @micaeljtoliveira and I have identified that the model initialisation logic is duplicated across all drivers (serial and MPI master/worker). This change adds a preliminary initialisation routine to remove code duplication and improve code clarity across each of the drivers. The remaining initialisation code in the offline drivers will be moved to this routine in later pull requests to clean up the drivers further.

## Checklist

----
📚 Documentation preview 📚: https://cable--440.org.readthedocs.build/en/440/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,7 @@ target_link_libraries(cable_common_objlib PRIVATE PkgConfig::NETCDF)
 add_executable(
     cable
     src/offline/cable_driver.F90
+    src/offline/cable_driver_init.F90
     "$<TARGET_OBJECTS:cable_common_objlib>"
 )
 target_link_libraries(cable PRIVATE PkgConfig::NETCDF)
@@ -163,6 +164,7 @@ install(TARGETS cable RUNTIME)
 if(CABLE_MPI)
     add_executable(
         cable-mpi
+        src/offline/cable_driver_init.F90
         src/offline/cable_mpicommon.F90
         src/offline/cable_mpidrv.F90
         src/offline/cable_mpimaster.F90
@@ -170,6 +172,7 @@ if(CABLE_MPI)
         src/science/pop/pop_mpi.F90
         "$<TARGET_OBJECTS:cable_common_objlib>"
     )
+    target_compile_definitions(cable-mpi PRIVATE __MPI__)
     target_link_libraries(cable-mpi PRIVATE PkgConfig::NETCDF MPI::MPI_Fortran)
     install(TARGETS cable-mpi RUNTIME)
 endif()
--- a/src/offline/cable_driver_init.F90
+++ b/src/offline/cable_driver_init.F90
@@ -1,5 +1,6 @@
 MODULE cable_driver_init_mod
   !! Module for CABLE offline driver initialisation.
+  USE cable_namelist_util, ONLY : get_namelist_file_name
   IMPLICIT NONE
   PRIVATE
 
@@ -9,6 +10,11 @@ CONTAINS
 
   SUBROUTINE cable_driver_init()
     !! Model initialisation routine for the CABLE offline driver.
+
+    !check to see if first argument passed to cable is
+    !the name of the namelist file
+    !if not use cable.nml
+    CALL get_namelist_file_name()
 
 #ifdef __MPI__
     ! MPI specific initialisation
--- a/src/offline/cable_driver_init.F90
+++ b/src/offline/cable_driver_init.F90
@@ -1,0 +1,21 @@
+MODULE cable_driver_init_mod
+  !! Module for CABLE offline driver initialisation.
+  IMPLICIT NONE
+  PRIVATE
+
+  PUBLIC :: cable_driver_init
+
+CONTAINS
+
+  SUBROUTINE cable_driver_init()
+    !! Model initialisation routine for the CABLE offline driver.
+
+#ifdef __MPI__
+    ! MPI specific initialisation
+#else
+    ! Serial specific initialisation
+#endif
+
+  END SUBROUTINE cable_driver_init
+
+END MODULE cable_driver_init_mod
--- a/src/offline/cable_driver_init.F90
+++ b/src/offline/cable_driver_init.F90
@@ -1,12 +1,80 @@
 MODULE cable_driver_init_mod
   !! Module for CABLE offline driver initialisation.
-  USE cable_namelist_util, ONLY : get_namelist_file_name
+  USE cable_common_module, ONLY : &
+    filename,                     &
+    calcsoilalbedo,               &
+    redistrb,                     &
+    wiltParam,                    &
+    satuParam,                    &
+    cable_user,                   &
+    gw_params
+  USE cable_IO_vars_module, ONLY : &
+    soilparmnew,                   &
+    output,                        &
+    patchout,                      &
+    check,                         &
+    verbose,                       &
+    leaps,                         &
+    logn,                          &
+    fixedCO2,                      &
+    ncciy,                         &
+    gswpfile,                      &
+    globalMetfile
+  USE casadimension, ONLY : icycle
+  USE casavariable, ONLY : casafile
+  USE cable_namelist_util, ONLY : &
+    get_namelist_file_name,       &
+    CABLE_NAMELIST
 #ifdef __MPI__
   USE mpi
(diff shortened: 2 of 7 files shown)


src/offline/cable_driver_init.F90
```diff
@@ -0,0 +1,117 @@
...
+  LOGICAL, SAVE, PUBLIC :: spinup        = .FALSE. ! model spinup to soil state equilibrium?
+  LOGICAL, SAVE, PUBLIC :: spincasa      = .FALSE. ! TRUE: CASA-CNP Will spin mloop times, FALSE: no spin up
+  LOGICAL, SAVE, PUBLIC :: l_casacnp     = .FALSE. ! using CASA-CNP with CABLE
+  LOGICAL, SAVE, PUBLIC :: l_landuse     = .FALSE. ! using CASA-CNP with CABLE
+  LOGICAL, SAVE, PUBLIC :: l_laiFeedbk   = .FALSE. ! using prognostic LAI
+  LOGICAL, SAVE, PUBLIC :: l_vcmaxFeedbk = .FALSE. ! using prognostic Vcmax
+
+  REAL, SAVE, PUBLIC :: delsoilM ! allowed variation in soil moisture for spin up
+  REAL, SAVE, PUBLIC :: delsoilT ! allowed variation in soil temperature for spin up
+  REAL, SAVE, PUBLIC :: delgwM = 1e-4
```
micaeljtoliveira, line 45: This module could be a good place to stop using module variables for passing data between function, which is considered bad practice. Instead, data should be passed as function arguments.

src/offline/cable_driver_init.F90
```diff
@@ -0,0 +1,117 @@
...
+    !if not use cable.nml
+    CALL get_namelist_file_name()
+
+#ifndef __MPI__
+    WRITE(*,*) "THE NAME LIST IS ", CABLE_NAMELIST
+#endif
+    ! Open, read and close the namelist file.
+    OPEN(10, FILE=CABLE_NAMELIST, STATUS="OLD", ACTION="READ")
+    READ(10, NML=CABLE)
+    CLOSE(10)
```
micaeljtoliveira, line 100: Maybe a good place to start introducing the `newunit` specifier?

```suggestion
    OPEN(newunit=unit, FILE=CABLE_NAMELIST, STATUS="OLD", ACTION="READ")
    READ(unit, NML=CABLE)
    CLOSE(unit)
```
with `unit` being an integer to be declared above. 

This is safer and easier to read (here, `10` is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)))

src/offline/cable_driver_init.F90
```diff
@@ -0,0 +1,117 @@
...
+
+#ifdef __MPI__
+    CALL MPI_Init(ierr)
+    CALL MPI_Comm_dup(MPI_COMM_WORLD, comm, ierr)
+    CALL MPI_Comm_size(comm, np, ierr)
+
+    IF (np < 2) THEN
+      WRITE (*,*) 'This program needs at least 2 processes to run!'
```
micaeljtoliveira, line 108: Another good practice that you might consider using here:
```suggestion
      WRITE (error_unit,*) 'This program needs at least 2 processes to run!'
```
This requires adding:
```fortran
USE iso_fortran_env, ONLY : error_unit
```
along the other use statements above.

src/offline/cable_mpicommon.F90
```diff
@@ -27,6 +27,13 @@ MODULE cable_mpicommon
 
   PUBLIC
 
+  ! MPI commicator and rank are declared as global variables (see
+  ! cable_driver_init_mod for their initialisation). This is done so that the comm
+  ! and rank variables are only accessible in cable_driver_init_mod when MPI
+  ! compilation is enabled.
+  ! TODO(Sean): revise which module to declare these variables in
+  INTEGER :: comm, rank
```
micaeljtoliveira, line 35: I think you can leave this here for now. Later on we should introduce an object to wrap this kind of MPI information.

micaeljtoliveira (approved):
@SeanBryan51 Overall looks good!

Most of my comments are quite general and, at this point, they are all optional.

src/offline/cable_driver_init.F90
SeanBryan51: FYI I initially considered defining these module variables as function arguments to `cable_driver_init` but I realised the variables were not defined at the top level scope of `cable_mpidrv.F90` which prevents me from passing in these variables as arguments. We could have the master and worker drivers call `cable_driver_init` separately instead of in `cable_mpidrv.F90` so that the variables were in scope but we would need to take out the MPI specific initialisation out of `cable_driver_init` as it would be called after the split between master and worker. Given the other namelist variables are module variables, I figured that would be the easiest workaround.

src/offline/cable_driver_init.F90
SeanBryan51: I will leave this for now as @Whyborn is adding this in as part of https://github.com/CABLE-LSM/CABLE/pull/438

src/offline/cable_driver_init.F90
micaeljtoliveira: Thanks for the explanation. Let's leave this as it is for now. We can revisit this when all drivers are merged.

src/offline/cable_driver_init.F90
micaeljtoliveira: Lets continue this discussion in the issue so that it doesn't block this PR.

CMakeLists.txt
```diff
@@ -162,13 +163,15 @@ install(TARGETS cable RUNTIME)
 if(CABLE_MPI)
     add_executable(
         cable-mpi
+        src/offline/cable_driver_init.F90
```
micaeljtoliveira, line 166: I don't think this works as intended. Basically there are two rules to generate the same object file and associated module file, one for serial and one for MPI. This means that when they are recompiled the second time, they object file from the first compilation is overwritten.

Compilation when using Makefiles works, but more by accident than by design. On the other hand, Ninja simply refuses to compile the files and rightly so.

I think the correct solution here is to have all the serial and MPI specific objects and module files being generated in different folders.

micaeljtoliveira (changes requested):
I think the issue with the CMake mentioned above needs to be fixed before merging this.

CMakeLists.txt
SeanBryan51: For reference [this issue](https://gitlab.kitware.com/cmake/cmake/-/issues/22898) looks like it describes what is happening in our case. It provides a workaround using the [Fortran_MODULE_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/Fortran_MODULE_DIRECTORY.html) target property. Micael has proposed an alternative way to structure the CMakeLists.txt so we avoid this issue entirely.

micaeljtoliveira: I would suggest to merge this PR as is and I'll then create a new PR with further changes that will have the side-effect of fixing this issues.

micaeljtoliveira: New branch with proposed changes here: https://github.com/CABLE-LSM/CABLE/tree/introduce_mpi_group Relevant commit: b9e6056e636fe077ad2ddabd53a5f3f1d3661a2d

This still needs to some to be properly documented.

micaeljtoliveira (approved):

ccarouge (approved):

SeanBryan51:
Testing with `benchcab fluxsite -v` shows this branch bitwise reproduces model outputs from the main branch:

PBS log file:
```
2024-11-07 12:29:39,954 - INFO - benchcab.benchcab.py:369 - Running comparison tasks...
2024-11-07 12:29:39,977 - INFO - benchcab.benchcab.py:370 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
2024-11-07 12:32:37,432 - INFO - benchcab.benchcab.py:381 - 0 failed, 168 passed
```
Benchcab version: `4.1.0`

`config.yaml`:
```
realisations:
  - repo:
      git:
        branch: main
  - repo:
      git:
        branch: 425-offline-driver-initialisation-refactor

modules: [
  intel-compiler/2021.1.1,
  netcdf/4.7.4,
  openmpi/4.1.0
]
```

PR #453, opened 2024-10-30
enaable          rootbeta parametrization of froot

### LINKED ISSUE #451: enable root distribution flexibility in ESM1.5
There is a cable_user%access13roots flag that defaults to FALSE in utils/cable_runtime_opts_mod.F90.
It is currently only used in coupled/ACCESS-CM2/init/cable_um_init_subrs.F90, although in some of my tests I was also making use of it in offline code (a version of offline/cable_parameters.F90). It is not currently used in any of the ESM1.5 code but that would be one option for cleaning this up - noting that there are already related issues in the CABLE repository about making the root distribution choice more flexible.

See issue 23 in the UM7 repo.

JhanSrbinovsky:
# CABLE

## Description

enable parametrisation bases on rootbeta as default behaviour

Fixes #451 

## Type of change

## Checklist

----
📚 Documentation preview 📚: https://cable--453.org.readthedocs.build/en/453/

### DIFF
--- a/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_um_init_subrs.F90
+++ b/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_um_init_subrs.F90
@@ -406,14 +406,18 @@ veg%clitt(h)    = vegin%clitt(veg%iveg(h))
          END DO ! over each veg patch in land point
  
 IF (cable_user%access13roots) THEN
-  ! default prescribed values used in CMIP6 ACCESS-ESM1.5
-  veg%froot(:,1) = 0.05
-  veg%froot(:,2) = 0.20
-  veg%froot(:,3) = 0.20
-  veg%froot(:,4) = 0.20
-  veg%froot(:,5) = 0.20
-  veg%froot(:,6) = 0.15
+
+  ! prescribed values from vegin%froot so can be set to ESM1.5 values
+  ! which were the same for all pfts (0.05, 0.2, 0.2, 0.2, 0.2, 0.15)
+  ! or alternate prescribed root fractions could be used with pft dependence
+  DO is = 1, ms
+    DO h=1, mp
+      veg%froot(h,is) = vegin%froot(is,veg%iveg(h))
+    ENDDO
+  ENDDO
+
 ELSE
+
   ! parametrized values used in CMIP6 ACCESS-CM2
   ! calculate vegin%froot from using rootbeta and soil depth
   ! (Jackson et al. 1996, Oceologica, 108:389-411)
--- a/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_um_init_subrs.F90
+++ b/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_um_init_subrs.F90
@@ -329,8 +329,6 @@ END SUBROUTINE initialize_veg
    
 SUBROUTINE init_veg_pars_fr_vegin( soil_zse ) 
 
-USE cable_common_module, ONLY: cable_user
-USE cable_params_mod,    ONLY: vegin
 USE cable_um_tech_mod,   ONLY: veg, soil 
 USE cable_def_types_mod, ONLY: mp, ms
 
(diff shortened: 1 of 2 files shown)


rml599gh (commented):
This code change looks fine. Have you done any testing of the change?

Perhaps also worth cleaning up:
There are froot values set in cable_pft_params.F90 but these are never used and are incorrect if they were to be used - the froot values should sum to 1 over soil levels which is the first index for vegin%froot(soil-level, veg-type) but the 2nd index for veg%froot(mp, soil-level). Suggest delete the lines unless we wanted to have the capability to set froot values by pft.
I noticed rootbeta is allocated in allocate_veg_params_cbl.F90 while froot is allocated both in that file and in cable_define_types.F90. Is this correct?

JhanSrbinovsky:
I'll have to setup an ESM1.5 build/run w the current main CABLE and then a 2nd with this branch.

Analysis will be difficult. I'm almost certain that with both (main&here) set to TRUE/FALSE there will be no difference. However, testing the impact of TRUE vs FALSE, or comparing the 2 methods will show differences. A 20+ year run will then be needed BUT I am assuming that your original comment implies that you have already done this and so are wanting to move to the root_beta method. In which case, what am I testing - that the IF() condition works?

rml599gh:
I think it is worth testing that using access13roots=TRUE is the same as the current 'main'. For the change from froot to rootbeta - probably the best we can do is compare my test runs with yours. The carbon fluxes do seem to be quite sensitive to the change, including smaller seasonality - which was already underestimated in ESM1.5, so this may discourage us from going to the rootbeta method.

JhanSrbinovsky:
That's what I mean, we are effectively testing whether the IF condition actually works, or am I missing something?

oh BTW - I doubt the cable_define_types ALLOCATION is ever called. I just never deleted it.

JhanSrbinovsky:
I just looked at the allocate() issue. 

in the alloc_veg_params module the CALL is to allocate_veg_parameter_type. The cable_define_types thing CALL allocate_cbm_var. So the other is just dangling it seems.

JhanSrbinovsky:
how is this going @rml599gh ?

rml599gh (commented):
I tried to take a copy of the file that is changed: cable_um_init_subrs.F90 to run some tests myself and it doesn't compile because 'totdepth' and 'is' are not explicitly defined. Is this because the rootbeta block of code has been moved out of the init_veg_from_vegin subroutine into init_veg_pars_fr_vegin? I suggest we might want to do this the other way and move the ESM1.5 froot block into init_veg_from_vegin subroutine. This might set us up better for coding up the option to prescribe froot by pft.

JhanSrbinovsky:
odd - it seems to have built here - anyway - about to head into a meeting. I'll look at it after that.

JhanSrbinovsky:
haha - of course it has - checks doesnt build coupled models

JhanSrbinovsky:
builds in ESM now

rml599gh:
@JhanSrbinovsky - I've also had a go at giving extra flexibility as to how froot is set.
```
 IF (cable_user%access13roots) THEN
      ! prescribed values from vegin%froot so can be set to ESM1.5 values
      ! which were the same for all pfts (0.05, 0.2, 0.2, 0.2, 0.2, 0.15)
      ! or alternate prescribed root fractions could be used with pft dependence
      DO is = 1, ms
        DO h=1, mp
          veg%froot(h,is) = vegin%froot(is,veg%iveg(h))
        ENDDO
      ENDDO
    ELSE
```
What do you think? I haven't tested this yet, only that it compiles. If we implement this, then we also need to update cable_pft_params.F90 as the vegin%froot values in the current file are incorrect. I think they should be
```
    !vegin%froot(soil-level,veg-type) ESM1.5 values (all pfts the same)
    vegin%froot(1,1) =        0.050000
    vegin%froot(2,1) =        0.200000
    vegin%froot(3,1) =        0.200000
    vegin%froot(4,1) =        0.200000
    vegin%froot(5,1) =        0.200000
    vegin%froot(6,1) =        0.150000
```
and replicated across each of the PFTs.
My file is here:
/g/data/p66/rml599/ESM1.5-CABLE3-main/submodels/UM/umbase_hg3/src/atmosphere/CABLE/src/coupled/ESM1.5/cable_pft_params.F90

It also has updated parameters for running Medlyn and the CM2/ESM1.5 refl/taul parameters next to each other in the file so it is clearer to swap between them.

JhanSrbinovsky:
no problem. I dont know where else the froot values in there could've come from?

rml599gh:
It just looked like they'd been put into the subroutine incorrectly (pfts and soil levels swapped) i.e. all the pft1 values were 0.05, all the pft2 values were 0.2 etc and after 6 pfts, everything was zero.

JhanSrbinovsky:
I can imagine how such an error could occur. mixing up dimensions (pft,layer).

JhanSrbinovsky:
done :)

rml599gh:
I ran some one-month tests replicating these changes in my working copy. Plots are zonal mean GPP. Black and green are cases where I'd switched files in the build in order to change between froot and rootbeta. Red and blue use the new file and the access13roots switch i.e. a run-time rather than build-time choice.
image

And here's a difference plot, with the original froot case subtracted from each of the others. Black is switchable with froot used. Red and green are rootbeta with red hardwired and green switchable.
image

Ideally the black line should be zero and the red and green should be identical but something seems to have given divergent meteorology. Probably should at least check that we are still bit reproducible.

rml599gh (approved):

PR #458, opened 2024-11-04
322 use netcdf inputs for bios met

Whyborn:
# CABLE

## Description

This PR routes the BIOS inputs through the new input routines developed for the CRU/TRENDY experiments. It involves some minor restructuring of the existing routines to handle instances where an amount less than the full set of possible variable datasets are passed via the namelist. The ```cable_user%RunType='bios'``` still exists for the ancillaries. This should operate completely independently of the ancillary work.

The outputs of the 9 point ACT test are bitwise equivalent with 2 caveats:

- The handling of the ```WG%TempMaxDayPrev``` is corrected in the original BIOS routines (handling at first day of run was incorrect).
- The longitude/latitude values in the output differ on the order of the floating precision delta, since they are computed in different ways.

I am doing some longer term work in neatening up and improving the met routines to generalise them for more arbitrary sets of input variables, but this serves as an acceptable interim that should work seamlessly with Abhaas' work with the ancillaries.

Part of #(322)

## Type of change

Enhancement

## Checklist


offline/cable_cru_TRENDY.F90
```diff
@@ -451,18 +636,43 @@ SUBROUTINE cru_get_daily_met(CRU, LastDayOfYear)
     DummyYear = CRU%cYear + 1
   END IF
 
-  ! Was the Tmin recycled?
+  ! Now we have to handle each of the "next" variables individually
+  ! Tmax
   IF (CRU%isRecycled(Tmin)) THEN
```
ccarouge, line 641: Why does it say Tmax in the comment but looks at "isRecycled(**Tmin**)"?

offline/cable_cru_TRENDY.F90
```diff
@@ -408,38 +551,80 @@ SUBROUTINE cru_get_daily_met(CRU, LastDayOfYear)
...
+  END IF
+
+  DummyDay = 365
+  IF ((CRU%LeapYears) .AND. (is_leapyear(DummyYear))) THEN
+    DummyDay = 366
+  END IF
+
+  ! Now we just need to call cru_read_metvals with the Tmax Dataset reader and
+  ! the prevTmax array to write to
+  CALL read_metvals(CRU%MetDatasets(vph1500), CRU%Met(prevvph1500)%MetVals, land_x,&
```
ccarouge, line 616: Why do we check "isRecycled(vph**0900**)" but use vph**1500** in `read_metvals`?

I am also wondering if it would be worth having a function to "setDummyTime_and_ReadTheData" since it's always done in the same way and it's only the variable indexes that change. Would need either 2 functions or a logical to swap between previous time and next time calculations.

Such a function would avoid the mistake at the start of this comment, if it is a mistake.

offline/cable_cru_TRENDY.F90
```diff
@@ -451,18 +636,43 @@ SUBROUTINE cru_get_daily_met(CRU, LastDayOfYear)
...
-    DummyYear = CRU%RecycStartYear + MOD(DummyYear - 1501, CRU%metRecyc)
+    DummyYear = CRU%RecycStartYear + MOD(DummyYear - 1501, CRU%MetRecycPeriod)
   END IF
 
   CALL read_metvals(CRU%MetDatasets(Tmin), CRU%Met(nextTmin)%MetVals, land_x,&
     land_y, DummyYear, DummyDay, CRU%LeapYears, CRU%xDimSize, CRU%yDimSize,&
     CRU%DirectRead)
 
-END SUBROUTINE cru_get_daily_met
+  ! vph1500
```
ccarouge, line 649: The code afterwards is using vph0900 everywhere, so why does the comment say "vph1500"?

offline/cable_bios_met_obs_params.F90
ccarouge, line 1: Why should we keep this file if completely commented out? We can always retrieve it from the history with git.

ccarouge (changes requested):
It's good. Just a few comments and potentially a mistake.

offline/cable_cru_TRENDY.F90
```diff
@@ -29,22 +29,41 @@ MODULE CABLE_CRU
...
+! to stop the building process.
+REAL, PRIVATE, PARAMETER  :: SecDay = 86400.
+INTEGER, PRIVATE, PARAMETER  :: rain = 1, lwdn = 2, swdn = 3, pres = 4,&
+                                qair = 5, tmax = 6, tmin = 7, uwind = 8,&
+                                vwind = 9, wind = 10, vph0900 = 11, vph1500 = 12,&
+                                fdiff = 13, prevTmax = 14, nextTmin = 15,&
+                                nextvph0900 = 16, prevvph1500 = 17
+INTEGER, PRIVATE, PARAMETER  :: sp = kind(1.0)
+INTEGER, PRIVATE, PARAMETER  :: dp = KIND(1.0d0)
+INTEGER, PRIVATE, PARAMETER  :: nVariables = 13
```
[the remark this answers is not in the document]
Whyborn: I think the best way is to add some commentary. The nVariables is the number of "readers" we need, since we don't need to open another reader for the next day/previous day variables. The difference between the numbers of variables we actually need vs the number of datasets we need open is the reason for the ```nExtraVariables``` in the next line. Most of the iteration loops are based on this ```nVariables```, but we need storage for ```nVariables + nExtraVariables``` arrays of variables. Perhaps a rename of the variables? ```nCoreVariables``` and ```nDerivedVariables```? But even then, they're not "derived" as such.

offline/cable_cru_TRENDY.F90
```diff
@@ -365,11 +373,146 @@ SUBROUTINE CRU_GET_SUBDIURNAL_MET(CRU, MET, CurYear, ktau, kend)
...
+  END IF  ! End start of day accounting
+
+  CALL WGEN_SUBDIURNAL_MET(WG, CRU%mLand, NINT(Met%hod(1) * 3600.0 / Dt))
+
+  ! Now pass the data out to the land tiles
+  PassToTiles: DO iLand = 1, CRU%mLand
+    is = LandPt(iLand)%cStart
+    ie = LandPt(iLand)%cEnd
+
+    WRITE(*,*) WG%Temp
```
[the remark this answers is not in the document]
Whyborn: Yep, missed that one in the cleanup

offline/cable_cru_TRENDY.F90
```diff
@@ -408,38 +551,80 @@ SUBROUTINE cru_get_daily_met(CRU, LastDayOfYear)
...
+  ! 1. We want to retrieve that correct data for the relevant year used for
+  !   recycling, which implies leap years should be based on the met recycling
+  !   year. This prevents instances where we try to access the 366th day of a
+  !   year with 365 days, or only get 365 days from a 366 day year.
+  ! 2. We often want to perform yearly averages or accumulations over a year.
+  !   These accumulations/averages occur elsewhere in the code, and thus use the
+  !   number of days from the core simulation year. This imples leap years
+  !   should be based on the core simulation year, to prevent instances where
+  !   we accumulate meteorology over 366 days but average over 365 days and vice
+  !   versa.
```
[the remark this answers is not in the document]
Whyborn: I think this may come down to a question of exactly how much flexibility do we want with the forcing- in the original TRENDY set-up, there were specific ```RunType```s which recycled some variables but used the simulated year for others, which is why I included the tailored recycling options. I think there are instances where this tailoring could be useful, in instances where some variables have a longer history than others. But it would make the handling easier if we just set a blanket "Met recycling on or off".

offline/cable_bios_met_obs_params.F90
Whyborn: It still reads in the BIOS ancillaries from binary. I was debating on whether to keep the commented sections to make it easier to check back against existing behaviour until the ancillary side of things is finished.

offline/cable_cru_TRENDY.F90
Whyborn: Looks like I got a little mixed up here- need to get the ```vph1500``` from the previous day and ```vph0900``` from the next. I was retrieving the right data, but checking against the wrong recycling logical.

offline/cable_cru_TRENDY.F90
Whyborn: I think it might be better if I just pull the date calculation into a separate routine? Something like
```
SUBROUTINE get_met_date(SimYear, SimDay, IsRecycled, RecycleStart, RecyclePeriod, LeapYears, MetYear, MetDay)
! Some routine that modifies MetYear and MetDate based on the recycling parameters
END SUBROUTINE get_met_date
```
Then there's a workflow that works identically for any variable- compute the date, then read the data. We also get a nice division of responsibility that way- one function gets the date, one function gets the data.

ccarouge (changes requested):
I generally like it but a few more comments in. There is a weird one in there which could potentially be a bug, so not approving to make sure this is resolved before it gets merged.

offline/cable_cru_TRENDY.F90
```diff
@@ -365,104 +380,412 @@ SUBROUTINE CRU_GET_SUBDIURNAL_MET(CRU, MET, CurYear, ktau, kend)
...
 
-END SUBROUTINE cru_get_daily_met
+    ! This handles any future scenarios where we may change the day by >1
+    MetDay = 365 + SimDay
+    IF ((LeapYears) .AND. is_leapyear(MetYear)) THEN
+      MetDay = 366 + SimDay
+    END IF
+  ELSEIF (SimDay == (DaysInYear + 1)) THEN
+    MetYear = SimYear + 1
+    MetDay = SimDay - DaysInYear
```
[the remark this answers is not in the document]
Whyborn: Yep, should definitely by ```>=``` rather than ```==```. I originally built it specifically for +- one day, but figured why not make it a bit more generic?

offline/cable_cru_TRENDY.F90
```diff
@@ -365,104 +380,412 @@ SUBROUTINE CRU_GET_SUBDIURNAL_MET(CRU, MET, CurYear, ktau, kend)
...
-    DummyYear = CRU%RecycStartYear + MOD(DummyYear - 1501, CRU%metRecyc)
+  ! How many days in the year, for correct end of year handling
+  IF (LeapYears) THEN
+    DaysInYear = DaysInYear + leap_day(MetYear)
   END IF
 
-  CALL read_metvals(CRU%MetDatasets(Tmin), CRU%Met(nextTmin)%MetVals, land_x,&
-    land_y, DummyYear, DummyDay, CRU%LeapYears, CRU%xDimSize, CRU%yDimSize,&
-    CRU%DirectRead)
+  ! Check the validity of the passed day
```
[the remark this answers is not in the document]
Whyborn: Yep, this is true. I'll put that as the first check. I think I'll throw an informative message and exit the program if ```-364 > SimDay > DaysInYear + 365```.

offline/cable_cru_TRENDY.F90
```diff
@@ -365,104 +380,412 @@ SUBROUTINE CRU_GET_SUBDIURNAL_MET(CRU, MET, CurYear, ktau, kend)
...
+  ! check for recycling at the start of the routine, before doing any year
+  ! shifts when the day = 0 or DaysInYear. This means it is possible to retrieve
+  ! meteorology from outside the recycling period when handling neighbouring
+  ! days.
+  ! For example, we have a recycling period from 1951 to 1960, and we want the
+  ! meteorology from 01/01/1951, and consequently some of the previos day's
+  ! meteorology. The "previous day" with respect to the meteorology can be
+  ! either 31/12/1950 or 31/12/1960, depending on when the MODULO is applied.
+  ! The current approach would use 31/12/1949 as the previous day, which is less
+  ! likely to cause issues with large jumps in the time-local conditions.
```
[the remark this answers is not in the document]
Whyborn: Yep, I did mean 1950. I treat that as being out of the scope of this routine- this just determines what date to send to the ```SpatioTemporalDataset``` reader. Handling edge cases such as this is down to the rules of the ```SpatioTemporalDataset```, which at the moment is to get the date closest to the requested date i.e. the first day of the dataset if asking for a date before the dataset begins, the last for dates after the dataset finishes. 

This was chosen as it was consistent with the existing CRU behaviour. There are two options for handling this I think, either:

1) Current behaviour, get closest date to the requested date (treat 31/12/1950 as 01/01/1951).
2) Treat the year like a modulo and get the requested day from the nearest year (treat 31/12/1950 as 31/12/1951).

I think option 1 is better in the most common instance that this occurs, which is subdiurnalisation. Option 2 is probably better in scenarios where we're operating months/years outside the dataset range, because we'd still get the seasonality, but I don't think this scenario occurs anywhere in CABLE at the moment.

offline/cable_cru_TRENDY.F90
```diff
@@ -517,160 +523,267 @@ SUBROUTINE get_daily_met(CRU)
...
+    CRU%RecycStartYear, CRU%MetRecycPeriod, CRU%LeapYears, MetYear,&
+    MetDay)
+  CALL read_metvals(CRU%MetDatasets(Tmin), CRU%Met(nextTmin)%MetVals,&
+    land_x, land_y, MetYear, MetDay, CRU%LeapYears, CRU%xDimSize,&
+    CRU%yDimSize, CRU%DirectRead)
 
   ! vph0900
-  DummyYear = CRU%cYear
-  DummyDay = CRU%CTStep - 1
+  CALL get_met_date(CRU%cYear, CRU%CTStep-1, CRU%IsRecycled(vph0900),&
```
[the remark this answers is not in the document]
Whyborn: Corrected, thanks.

offline/cable_cru_TRENDY.F90
```diff
@@ -764,7 +754,14 @@ SUBROUTINE get_met_date(SimYear, SimDay, IsRecycled, RecycleStart,&
...
   ! Check the validity of the passed day
-  IF (SimDay <= 0) THEN
+  IF ((SimDay < -364) .OR. (SimDay > (DaysInYear + 365)) THEN
+    ! Looking more than a year into the past/future- this routine was never
+    ! meant for such tasks.
+    WRITE(ERROR_UNIT,*) "Looking too far into the past/future. "//&
+      "The get_met_date routine was only intended for nearby temporal "//&
+      "searching, not more than a year in the future/past."
+    EXIT 5
+  IF (SimDay < 1) THEN
```
ccarouge, line 764: Missing "END IF"?

ccarouge (commented):
The changes from the review are ok although it seems to be missing an END IF. Did it compile?

Also, you have conflicts with CABLE-POP_TRENDY branch to solve now.

har917:
@Whyborn @ccarouge Apologies I've just returned to trawling through github notifications and realised I hadn't looked at this from last month.  Could you go through and let me know if there are any unresolved questions from the thread?

In the main I'm less worried about maintaining reproducibility that I am about getting the science correct (but realise that's likely not how the NRI want to work)

Whyborn:
I'm in the process of merging into POP_TRENDY, just need to test after addressing the merge conflicts. I picked up a mistake I'd made in a prior PR while doing this, so got sidetracked fixing that.

PR #461, opened 2024-11-06
rm roughness update of met%tk

### LINKED ISSUE #459: remove `met%tk` update in cable_cbm when called from cable_explicit_driver(ESM1.5)
> quick comment - the line `met%tk = met%tk + ... ` _**HAS**_ to be removed - it is incorrect scientifically and its location in the code (in `cable_explicit:cbm`, i.e after `init_radiation` is called in the albedo code but before the radiation is evaluated fully) is a key source of energy imbalance.

I'll do this in a 2nd issue. not this cable_data issue

_Originally posted by @JhanSrbinovsky in https://github.com/CABLE-LSM/CABLE/issues/445#issuecomment-2458171723_

JhanSrbinovsky:
# CABLE

## Description

Incremental update of met%tk  following roughness calculation is removed. See issue #459 and links therein.

Fixes #459

## Type of change

- [X] Bug fix

## Checklist

- [Y] The new content is accessible and located in the appropriate section.
- [Y] I have checked that links are valid and point to the intended content.
- [Y] I have checked my code/text and corrected any misspellings

----
📚 Documentation preview 📚: https://cable--461.org.readthedocs.build/en/461/

### DIFF
--- a/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_cbm.F90
+++ b/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_cbm.F90
@@ -34,9 +34,6 @@ USE grid_constants_mod_cbl, ONLY : ICE_SoilType, lakes_cable
    USE cable_def_types_mod
    USE cable_roughness_module
    USE cable_air_module
-#ifndef NO_CASA_YET
-   USE casadimension,     only : icycle ! used in casa_cnp
-#endif
    USE cable_data_module, ONLY : icbm_type, point2constants 
    
    !ptrs to local constants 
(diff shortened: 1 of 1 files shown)


har917 (approved):
approved - this is the correct change (plus some whitespace changes).

@rml599gh @tammasloughran   I wonder whether we need to do a short run before/after though to quantify any systematic impact - I would expect that to first order this would lead to a cooling affect of around 0.3-0.4K over land (0.39K modified by roughness length and other feedbacks)

rml599gh:
I agree that it would be good to run a short test to check the impact. Probably best if @JhanSrbinovsky and I coordinate on that.

rml599gh:
I ran a 10 year test.
This is the difference in the 1.5m temperature field. As anticipated by @har917 there is a general cooling over land sometimes >0.4K presumably due to ocean variability. Are we happy that this is a sufficient test or is it worth continuing for another 10 years for a longer averaging period? @har917 - is it worth checking the energy balance in this run to see if it is improved?

image

har917:
@rml599gh It's a bit hard to tell given that everything is being dominated by internal variability - but this looks promising.  I wuld look at some of the fluxes as well (latent and sensible in particular, possibly GPP) to get an idea of the associated impact.  

It's worth checking the energy balance - however I'm not expecting that to be improved since the fundamental problem there is the mixed application of corrected and non-corrected fluxes* 

*the main issue in proto-CM2 was that the isothermal part of the longwave radiation balance (`init_radiation`) was evaluated down the albedo path (as needed) but not redone later on even though `met%tk` had been updated.  This led to ~1W/m2 constant imbalance - but I thought that `iniit_radiation` is called multiple times in the ESM.

JhanSrbinovsky:
keeping in mind that from cbm down, apart from specific ESM15 switches (either runtime or FPP) these below,  CABLE is 97% the same (can't say 100% the same as PRs pending that are holding this up). However nothing to do with calling init_rad. init_rad is only called from cbm in ESM1.5 however it is multiple times per timestep.

In terms of pushing this to main branch, I thought we were pretty much obliged to as it is wrong. Other corrections might need to be made, but this has to go in doesnt it?

:
coupled/ESM1.5/casa_landuse.F90:# define ESM15 YES
coupled/ESM1.5/casa_landuse.F90:#ifdef ESM15
science/casa-cnp/biogeochem_casa.F90:IF (cable_user%PHENOLOGY_SWITCH.EQ.'MODIS' .OR. cable_runtime%esm15 ) THEN
science/roughness/cable_roughness.F90:IF(cable_runtime%um_radiation .OR. cable_runtime%esm15                         &
science/soilsnow/cbl_surfbv.F90:    IF( .NOT. cable_runtime%UM  .OR. cable_runtime%esm15 ) THEN

har917:
The comment about `init_rad` relates to work in CM2 when we were trying to simplify/restructure things and this issue around `met%tk` manifested as a significant energy imbalance.

I can't easily figure out whether the poor energy balance of the ESM1.5 relates to this or not - regardless this update of `met%tk` needs to be removed.

JhanSrbinovsky:
ok - thanks. 

is there a bad energy imbalance in the ESM1.5?

PR #491, opened 2024-11-27
Fix to groundwater parameter initialisations that can impact the soil conductivity

### LINKED ISSUE #490: Missing variables in MPI messing up with soil conductivity
When `cable_user%soil_thermal_fix` is True, we calculate the soil conductivity using this function: https://github.com/CABLE-LSM/CABLE/blob/09a124303dc5154dc76d5f24a126ec9e6126391e/src/science/soilsnow/cbl_conductivity.F90#L11

It is called from `stempv`:
https://github.com/CABLE-LSM/CABLE/blob/09a124303dc5154dc76d5f24a126ec9e6126391e/src/science/soilsnow/cbl_stempv.F90#L57-L61

This new soil conductivity calculation is using the parameters `soil%sand_vec`, `soil%ssat_vec`, `soil%watr`, `soil%cnsd_vec`. These parameters were introduced for the groundwater scheme. It turns out these parameters were not fully implemented in main since the groundwater scheme isn't fully implemented. Initialisations and MPI implementations are faulty.

It could be these parameters or other parameters for the ground water are currently used in the code elsewhere. In this case, fixing this issue would have more impact than the soil conductivity alone. But I do not intend search to see if that is the case.

ccarouge:
# CABLE

## Description

Fixes #490 

Working on the groundwater implementation, I have found that the function `total_soil_conductivity` depends on some GW parameters that aren't correctly implemented. This is to fix this.

## Type of change

- [x] Bug fix

## Testing
Since this requires MPI to test the impact, I haven't used modelevaluation.org to analyse the results. I have run benchcab with the spatial configuration for:
1. main with `cable_user%soil_thermal_fix = .TRUE`
1. main with `cable_user%soil_thermal_fix = .FALSE`
1. this branch with `cable_user%soil_thermal_fix = .TRUE`
2. this branch with `cable_user%soil_thermal_fix = .FALSE`

Case 3 - Case 1 allow to gauge the total effect of the change. Case 2 - Case 1 give a measure of the change compared to the initial fix to the soil conductivity. Case 4 - Case 2 shows whether there are any impacts beyond the soil conductivity (without explicitly finding these other sources). 

The comparison of case 4 and case 2 with `cdo diffn` shows there are no differences when `cable_user%soil_thermal_fix` is false.

The largest effect for this branch is found on the ground flux but the effect is limited (right plot) compared to the effect of `soil_thermal_fix` (left plot):
ground_flux_effect
These are time-mean values of the differences between the simulations.

----
📚 Documentation preview 📚: https://cable--491.org.readthedocs.build/en/491/

### DIFF
--- a/src/offline/cable_mpicommon.F90
+++ b/src/offline/cable_mpicommon.F90
@@ -29,7 +29,7 @@ MODULE cable_mpicommon
 
   ! base number of input fields: must correspond to CALLS to 
   ! MPI_address (field ) in *_mpimaster/ *_mpiworker
-  INTEGER, PARAMETER :: nparam = 330
+  INTEGER, PARAMETER :: nparam = 340
    
   ! MPI: extra params sent only if nsoilparmnew is true
   INTEGER, PARAMETER :: nsoilnew = 1
--- a/src/offline/cable_mpimaster.F90
+++ b/src/offline/cable_mpimaster.F90
@@ -3183,6 +3183,66 @@ CONTAINS
             &                             types(bidx), ierr)
        blen(bidx) = 1
 
+       bidx = bidx + 1
+       CALL MPI_Get_address (soil%zse_vec(off,1), displs(bidx), ierr)
+       CALL MPI_Type_create_hvector (ms, r2len, r2stride, MPI_BYTE, &
+            &                             types(bidx), ierr)
+       blen(bidx) = 1
+
+       bidx = bidx + 1
+       CALL MPI_Get_address (soil%css_vec(off,1), displs(bidx), ierr)
+       CALL MPI_Type_create_hvector (ms, r2len, r2stride, MPI_BYTE, &
+            &                             types(bidx), ierr)
+       blen(bidx) = 1
+
+       bidx = bidx + 1
+       CALL MPI_Get_address (soil%cnsd_vec(off,1), displs(bidx), ierr)
+       CALL MPI_Type_create_hvector (ms, r2len, r2stride, MPI_BYTE, &
+            &                             types(bidx), ierr)
+       blen(bidx) = 1
+
+       bidx = bidx + 1
+       CALL MPI_Get_address (soil%clay_vec(off,1), displs(bidx), ierr)
+       CALL MPI_Type_create_hvector (ms, r2len, r2stride, MPI_BYTE, &
+            &                             types(bidx), ierr)
+       blen(bidx) = 1
+
+       bidx = bidx + 1
+       CALL MPI_Get_address (soil%sand_vec(off,1), displs(bidx), ierr)
+       CALL MPI_Type_create_hvector (ms, r2len, r2stride, MPI_BYTE, &
+            &                             types(bidx), ierr)
+       blen(bidx) = 1
+
+       bidx = bidx + 1
+       CALL MPI_Get_address (soil%silt_vec(off,1), displs(bidx), ierr)
+       CALL MPI_Type_create_hvector (ms, r2len, r2stride, MPI_BYTE, &
+            &                             types(bidx), ierr)
+       blen(bidx) = 1
+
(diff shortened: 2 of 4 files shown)


har917 (approved):
Approved:  code looks okay, comments are appropriate.

Testing is appropriate to the issue at hand - only aspect not assessed is the magnitude of the impact when vectorised soil parameters are used (but out of scope for now since the base case and inputs aren't available.).

NB - we may need to assess/check that these variables are initialised properly within the coupled model framework since we are defaulting to use `cable_users%soil_thermal_fix = .true.`

har917:
@ccarouge @JhanSrbinovsky Quick follow up on this PR as it relates to AM3 etc.

It would appear that most of the `_vec` variables and `soil%watr` are initialised from their single layer counterparts properly in AM3 - the exceptions being `clay_vec`, `sand_vec', `silt_vec` and `org_vec` as well as the `GW*_vec` (CASA vars and GW vars).  However the value of `soil%watr` is a bit questionable.

An element to pick up on later (between alpha and beta releases and/or with the merge of the GW science) is the meaning/role and consistency in value of `%watr` - offline seems to be setting this (in the absence of GW science) to 0.05 and/or 0.01 and/or 0.0, whereas CM2 and AM3 have it at 0.001.  I think there's some confusion as to whether `%watr` is a GW-only variable or more widely-used.

JhanSrbinovsky:
I was of the impression that the *_vec variables are not being used?

It looks like soil%watr is used IF soil_thermal_fixis TRUE. This is likely even why I switched it to false back when 4 timesteps was our longest run – from stempv

You should look at cable_um_init_soil. It is initialized there as .05 and later clobbered as .001 – intentional? My comment where it is clobbered reflects what is actually needed?

Jhan

From: har917 ***@***.***>
Date: Thursday, 28 November 2024 at 1:36 pm
To: CABLE-LSM/CABLE ***@***.***>
Cc: Srbinovsky, Jhan (Environment, Aspendale) ***@***.***>, Mention ***@***.***>
Subject: Re: [CABLE-LSM/CABLE] Fix to groundwater parameter initialisations that can impact the soil conductivity (PR #491)

@ccarouge<https://github.com/ccarouge> @JhanSrbinovsky<https://github.com/JhanSrbinovsky> Quick follow up on this PR as it relates to AM3 etc.

It would appear that most of the _vec variables and soil%watr are initialised from their single layer counterparts properly in AM3 - the exceptions being clay_vec, sand_vec', silt_vecandorg_vecas well as theGW*_vec(CASA vars and GW vars). However the value ofsoil%watr` is a bit questionable.

An element to pick up on later (between alpha and beta releases and/or with the merge of the GW science) is the meaning/role and consistency in value of %watr - offline seems to be setting this (in the absence of GW science) to 0.05 and/or 0.01 and/or 0.0, whereas CM2 and AM3 have it at 0.001. I think there's some confusion as to whether %watr is a GW-only variable or more widely-used.

har917:
> I was of the impression that the *_vec variables are not being used?

I hadn't fully realised this before - but Mark Dekker coded the `soil_thermal_fix` option for the soil thermal conductivity to use the `*_vec` variables (and `%watr`) all the time - regardless of whether `GW_MODEL = .true./.false.`.  Since we want (need) to use that science all the time this means that those variables need to carried and filled (in all applications) - ***or*** we need to rewrite the code to accommodate both cases.

I don't know what `%watr` is meant to represent - nor know how/why its default value has changed through the development process.  I just picked up its quirky initialisation when reviewing the ACCESS code in response to this PR request.

ccarouge:
The whole implementation of the `*_vec` variables is quirky and should not have been done this way. There is no reason to carry 2 sets of soil parameter variables in the code (except laziness around updating existing code to add the extra dim). And I agree it would have been nice to know what `watr` is!

PR #497, opened 2024-11-27
soilsnow developments from AM3

### LINKED ISSUE #494: merge soilsnow from AM3

JhanSrbinovsky:
# CABLE

## Description
Developments from AM3 work. Crazy number centre around using an ice density in place of liquid density, or 0.9. *  liquid density

iThis is why I left this til the end as there will be differences resulting from these mods

Fixes #494

## Type of change

- [x] model improvement

## Checklist

----
📚 Documentation preview 📚: https://cable--497.org.readthedocs.build/en/497/


src/science/soilsnow/cbl_soilsnow_main.F90
```diff
@@ -83,16 +84,15 @@ SUBROUTINE soil_snow(dels, soil, ssnow, canopy, met, bal, veg)
     ssnow%dtmlt = 0.0
     ssnow%osnowd = ssnow%snowd
 
-
-    wbliq = ssnow%wb - ssnow%wbice
+    ssnow%wbliq = ssnow%wb - ssnow%wbice
```
JhanSrbinovsky, line 87: this is a bug in main - it being local means it stays constant thru the run

rkutteh: There is not much I can say here of value except that the GW work did not alter in any way cbl_soilsnow_main.F90 (I double checked). I am not sure however if breaking the old soil_snow single module into many here introduced bugs, but that's of course unrelated to the GW work.

JhanSrbinovsky: that refactoring  was done long ago my friend. Thanks for looking at it

JhanSrbinovsky:
benchcab WILL reveal changes to output. In the process of downloading/uploading. My 3 site local test on gadi is pretty reasonable. Will post link to ME.org when done

JhanSrbinovsky:
still sorting out the ME.org plots. My testing shows seasonal diurnal quantities at Tumba (as expected) negligible. One of the sites I test at is Hyttialia (because there is snow): Note seasonal labels appropriate for Sothern Hemisphere

Albedo
Qh
Qle
Qmom

JhanSrbinovsky:
I think it looks as expected, I'll do a global one and see how it goes (I have to find the code first!)

JhanSrbinovsky:
This is using GSWP2 forcing. Most of the plots that flashed in front of my eyes as the scripts kept turning over were unremarkable I thought. But the surface SoilTemp caught my eye as it is one that it make sense to do a relative difference as it never gets near zero etc. As with all the others it is pretty bland, except for this bright spot in Northern Russia. On closer inspection this bullseye represents a ~10% diff. But it goes away after a few months. Odd, I wonder what could be causing this?  @har917 any ideas? Greenland is slightly cooler in Summer but not nearly as shocking. The Russian 10% difference appears to be an initialisation issue as it doesnt come back towards the end of the year. But still what is so dramatically different here. There was nothing here in AM3 was there?

RelativeDelta_jan

RelativeDelta_mar

RelativeDelta_jul

RelativeDelta_dec

har917:
@JhanSrbinovsky could be a feature of the initialisation - could be something to do with the ice fixes.

I'd be cautious about looking at %change in soil temperature - there's little context of a 10% difference in temperature -  depending on the units (C vs K) you can identify different regions for concern.  e.g. C will be particularly susceptible to divide by near-zero regions, using K will results in even large (+3K) differences only showing up at the 1% level.

JhanSrbinovsky:
@har917, its only really the ice fixes that are in here. but why should they apply to this bullseye in particular.  I just assumed it was Kelvin as almost everything in the UM is K. If it is in C, Im not so surprised. I just assumed this 1 spot was experiencing a ~30degree cooling, so probably C. I'll check

JhanSrbinovsky:
damn it. its in Kelvin

JhanSrbinovsky:
@har917 This lazy screenshot shows that almost everything is within -1% to +1% diff. That might be 3K and I can't comment on whether or not that is acceptable (or even expected) BUT the bullseye in Russia is 10 times that. Is this something to worry about, or should we move on?  

image

har917:
unfortunately I think we need to understand the origin of a ~30K difference in soil temperature - the most likely cause is that something that is now being used is getting through uninitialised (or to a default value of 0).

JhanSrbinovsky:
I dont have any ideas about what to look at. I might just look at the usual suspects and see what the moisture is doing at the same point. what else?
because it is isolated, and seems to go away by the next year I am hoping it is just a dodgy initialisation. Keep in mind as well that this is a no feedback, forced simulation.

ccarouge:
Since Ian says there is some digging to do to understand the differences, I've removed the request for review. This way I'm not coming back in here constantly to see if it's ready. @JhanSrbinovsky just ask for reviews again when ready.

JhanSrbinovsky:
Trouble spot zoomed in (rel. Diff ) T(K)
 
￼
Screenshot 2024-12-03 at 11 48 06 am




SoilMoist - Diff
￼
Screenshot 2024-12-03 at 12 06 37 pm



SoilMoistIce - Diff
Screenshot 2024-12-03 at 12 09 27 pm

￼



Iveg

Screenshot 2024-12-03 at 12 13 14 pm



￼


Rather than show you the dozens of field which are unremarkablekable:
snowdepth

Screenshot 2024-12-03 at 12 43 05 pm


Curiously the snow water equivalent does not show this. A clue that I can’t make sense of. I can imagine an initiallized snow depth which is all of a sudden subjected to a different density, but this wouldn't be the only place

JhanSrbinovsky:
This is interesting, the albedo:

Screenshot 2024-12-03 at 12 58 05 pm

JhanSrbinovsky:
the snowdepth which was this:
Screenshot 2024-12-03 at 1 02 16 pm

after 1 month is now:
Screenshot 2024-12-03 at 1 03 07 pm

JhanSrbinovsky:
are we fortuitously hitting a limit?

har917:
Jhan - can you check snow density?  There is a link between snow density and albedo (and a limit on that impact) in `surface_albedosn()`

This is a region with a lot of (frozen) lakes - remember we changed how `met%tk`/`ssnow%tggsn` got used - so that could be a factor as well.

JhanSrbinovsky:
sounds promising - I'll have to re-run both models - it isnt in my output. I imagine snow density output will be available through an easy switch

JhanSrbinovsky:
bugger - it isnt!! it is available in a restart but this isnt much good to us. I'll come up with something a bit later.

JhanSrbinovsky:
ok im back now. Does anyone know if it is possible to change the dump frequency in the offline model? This would be the easiest way I can think of to get a single point out of a global offline run?

JhanSrbinovsky:
Hey @SeanBryan51, you've recently been looking at the _drivers for offline. Is writing the restart coded after the timestep loop?

SeanBryan51:
> Hey @SeanBryan51, you've recently been looking at the _drivers for offline. Is writing the restart coded after the timestep loop? 

Apologies, I haven't looked into the restart IO in the drivers so I'm not too familiar on where the logic is in the code. This has not been touched yet so the restart behavior should be unchanged

JhanSrbinovsky:
no problem. Just thought you might've come across it.  I haven't been in that area for years it dumps the restart at the end of the time loop anyway. It would've been helpful in this issue if I could dump within the loop. Although, I only even want to do it because the IO dumps this particular field to the restart but not to the diagnostic output.

JhanSrbinovsky:
@har917 it took a bit of wrangling to get the model to print the snow density.  Then I was additionally thrown by the MPI component which I had forgotten about. Anyway, I haven't plotted anything yet, but I've isolated the region somewhat via the lat/lon co-ords. I couldn't see anything dramatic going in with density. So I output the soil Temp in the corresponding cell. On the 2nd timestep (and this looks like only on the 2nd timestep) - the Temp sinks to 25K and immediately recovers - although this is messing with the average. I suspect missing forcing data.

har917:
@JhanSrbinovsky Sounds like there may be something that's not being initialized properly in the ESM (so `cable_um_init_subs`) (but probably is in AM3) - could be `%tgg`, `%tggsn`, `%tss`.  Or there's some revised dependency that's using a currently unitialized variable.

Is it possible to force (compiler option) to initialise to e.g. NaN? That should help identify which/if a key variable is involved (through a segmentation fault on an identifiable variable/subroutine)

JhanSrbinovsky:
I'll try revised build again - this is offline(GSWP2) remember not ESM. I think it might be a revised initialisation. That it had before, but now doesnt. like the met%tk initialisation over lakes - but im not sure that is in here yet

ccarouge:
This sounds strangely identical to the issue @MartinDix found in AM3: https://github.com/CABLE-LSM/ACCESS3-JULES/issues/239

JhanSrbinovsky:
soil moisture goes crazy (-ve) on 2nd timestep. I "think" the reason it wasn't a recognised issue is because it goes away after a few months AND wbliq update was local and contained to soilsnow_main.   Why on 2nd timestep only - I think maybe something feeding into wb is wrongly initialized, which then sets wb entrance on the 2nd timestep.

JhanSrbinovsky:
so verification of one thing - bullseye disappears IF non-local wbliq is activated. As it is in AM3 HOWEVER NB. The crazy low temp is still there - its just on both sides now.

MartinDix:
> @JhanSrbinovsky Sounds like there may be something that's not being initialized properly in the ESM (so `cable_um_init_subs`) (but probably is in AM3) - could be `%tgg`, `%tggsn`, `%tss`. Or there's some revised dependency that's using a currently unitialized variable.
> 
> Is it possible to force (compiler option) to initialise to e.g. NaN? That should help identify which/if a key variable is involved (through a segmentation fault on an identifiable variable/subroutine)

This is what the rigorous build option with more recent versions of the UM uses. Compiler options `-init=snan -init=array`.  Possibly also need `-fpe0` to turn on trapping. Not sure whether ESM1.5 ever ran cleanly with this.

JhanSrbinovsky:
I think  i've found my problem at least. Should be fixed generically though.

wbice goes negative. I thought it was in cable_soilfreeze. I corrected it there by limiting it to >0, but it appears to be coming in -ve which I'll look at after a meeting this morning

JhanSrbinovsky:
sicemelt is greater than wbice on occasion. can clobber this to >0 but suggests an energy leak?

JhanSrbinovsky:
improves soil temp BUT it is still unrealistically cold

har917:
> sicemelt is greater than wbice on occasion.

Be careful here - sicemelt is an energy flux, wbice is a volumetric fraction of water that is frozen.  sicemelt will likely be realistically larger than wbice at times - but what we need is that the increment to wbice (because of the melting) lies between 0 and wbice.

If there is an issue here - I'm not sure how that comes about from the latest icefixes.  It suggests something else has gone really wrong beforehand.

I had a quick look through the code again and I'm definitely suspicious about where/when `%tss` is being first evaluated (in offline and ESM1.6).  It looks like this is in `cable_soil_snow_main` - CM2/AM3 has this in the `_um_init` section - but the variable is used in the albedo code, radiation and canopy sections.

JhanSrbinovsky:
tgg is diving on t=1 call to surfbv. re-enters soilsnow too low (for a few points). %snowd ~100 for surrounding points. At low T points %snowd ~565. Anyway, this can't be affecting it as it only feeds into the run off and this is zero as it is so cold. coming back again to same culprit in smoisture

    ! If redistribution of liquid has caused a layer to exceed frozen limit
    ! melt the excess and account for latent heat.
    dfactor = 1.0 - Cdensity_ice / Cdensity_liq
    DO k = 1, ms
       DO i = 1, mp
          IF (ssnow%wbice(i,k) > frozen_limit * ssnow%wb(i,k)) THEN
             sicemelt = (ssnow%wbice(i,k) - frozen_limit * ssnow%wb(i,k)) / &
                           (1.0 - frozen_limit*dfactor)
             ssnow%wbice(i,k) = MAX( 0.0_r_2, ssnow%wbice(i,k) - sicemelt )
             ssnow%wb(i,k) = ssnow%wb(i,k) - dfactor*sicemelt
             ssnow%tgg(i,k) = ssnow%tgg(i,k) -                                 &
                     sicemelt*soil%zse(k)*Cdensity_ice*CHLF / REAL(ssnow%gammzz(i,k) )
          END IF
       END DO
    END DO

JhanSrbinovsky:
actually I need to move diagonastics into smoisture as both wb and wbice are weird there - again only at a few points

har917:
@JhanSrbinovsky - the obvious thing to check is `%gammz` - I suspect that it's not being initialised properly (but thought it was in `stempv`)

Another aspect that could be the cause is that melt terms are usually linked to temperature increments via terms like
`snowmlt(j) * CHLF / ssnow%gammzz(j,1)` whereas here we have `sicemelt*soil%zse(k)*Cdensity_ice*CHLF / REAL(ssnow%gammzz(i,k) )` so there's a `%zse` and `Cdensity_ice` factor which is a candidate.

har917:
@JhanSrbinovsky (cc'ing @MartinDix @ccarouge for your thoughts) Two things (now that I'm back at the desk rather than travelling)

1. could you be a more precise on what/where you mention ensuring `ssatcurr > 0 ` (I'm worried that if can get `<=0` anywhere, anytime)?
2. since this is a first time step issue and related to new code - my new guess/thinking is that the code is okay but the initial conditions that are being used are incompatible with the new science.

On 2 - the way that this would work is - suppose that the old initial conditions are cold (i.e. soil moisture is nearly all frozen), and very close to saturated.  The new ice fixes reduce the available volume for ice by around 10% - this means that 10% of the ice in the initial conditions can no longer be supported.  Conservation of water/energy requires that unsupported ice to be melted (and passed to run off), that energy has to come from the soil, which leads to a very large, near instantaneous cooling. 

If 2 is the cause then this is really an particularly problematic case of using inputs with different origins that just so happen to be incompatible - always a risk.  The proper way froward would be to adjust the initial conditions (in gridinfo?) to avoid this - or we may need to add a `IF (first_call == .false.) THEN` around the %tgg line in the above.

NB - we could run into this problem with ESM1.6 as well.

JhanSrbinovsky:
I've got a better handle on things myself, the ssatcurr limiting "fixed" things slightly. The rest of the globe has only been affected by a fraction of a percemt. Even here, the biggest change is only ~4%, when the cooling was of order 10%. At any rate the initial temps seem quite high in these regions. Whatever, that could be the case. The crux of the discrepancy between main:494 is that main:  broadly continues this pattern, whereas the adjustment to tgg due to the latent heat of melting excess water/ice effectively "cools the hot spots". The region is more homogenous but IDK that this is necessarily what we want.  By hot spots I mean these cells are initially at ~273K where as surrounding region is commonly  20-40 below. I've already put a snippet of the code above. The bit with dfactor.  How do you want proceed @har917  ?

JhanSrbinovsky:
The same region - with the ssat limiting Straight K -K difference

Screenshot 2024-12-06 at 1 29 31 pm


and global
Screenshot 2024-12-06 at 1 33 12 pm

har917:
@JhanSrbinovsky I'm trying to unpick what you're showing and asking

By the 'ssat_curr' fix do you mean changing line 425(ish) in `cbl_smoisturev.F90` from

`ssatcurr(:,k) = soil%ssat - ssnow%wbice(:,k)`

to

`ssatcurr(:,k) = max(soil%ssat - ssnow%wbice(:,k),0.0)`?  

While this appears sensible I'm struggling to see how we get negative `ssatcurr` - i.e. there's something else going on which has allowed `%wbice > frozen_limit* %ssat` (equivalently `%wb > %ssat`)

I'm also struggling to understand what has been changed from the runs where you had a 30K difference to these latest runs where the differences appear to be more modest (up to 3K).  could you restate?

JhanSrbinovsky:
The two plots I just showed do NOT have the ssat range limiting, BUT yse it is like below:

> `ssatcurr(:,k) = max(soil%ssat - ssnow%wbice(:,k),0.0)`?

> While this appears sensible I'm struggling to see how we get negative `ssatcurr` - i.e. there's something else going on which has allowed `%wbice > frozen_limit* %ssat` (equivalently `%wb > %ssat`)

this I have mo insight into, I clobbered it here and changed tac
 
> I'm also struggling to understand what has been changed from the runs where you had a 30K difference to these latest runs where the differences appear to be more modest (up to 3K). could you restate?

so ALL the AM3 developments in soilsnow are included **EXCEPT** the code which increments tgg by the LE required to melt water above frozen limit as in the code in smoisture :

```
! If redistribution of liquid has caused a layer to exceed frozen limit
! melt the excess and account for latent heat.
dfactor = 1.0 - Cdensity_ice / Cdensity_liq
DO k = 1, ms
   DO i = 1, mp
      IF (ssnow%wbice(i,k) > frozen_limit * ssnow%wb(i,k)) THEN
         sicemelt = (ssnow%wbice(i,k) - frozen_limit * ssnow%wb(i,k)) / &
                       (1.0 - frozen_limit*dfactor)
         ssnow%wbice(i,k) = MAX( 0.0_r_2, ssnow%wbice(i,k) - sicemelt )
         ssnow%wb(i,k) = ssnow%wb(i,k) - dfactor*sicemelt
         ssnow%tgg(i,k) = ssnow%tgg(i,k) -                                 &
                 sicemelt*soil%zse(k)*Cdensity_ice*CHLF / REAL(ssnow%gammzz(i,k) )
      END IF
   END DO
END DO
```

And then it is only a problem at a few points globally. And then also, it seems to go away after spinning up for a while. I will see if I can spot it in the ESM16(-ish) output.  

I am having trouble reading it. I considered that it was just my misunderstanding or not knowing a missing piece of the puzzle, especially because it seems to work elsewhere in the globe. It seems to be kinda the reverse of what is in soilfreeze. It did occur to me though that the surrounding IF likely excludes almost everywhere anyway. I dont know how it gets in here. I haven't verified but almost certainly if these few lines alone are added to main, we'll get the same 30K discrepancy. Was it needed for the energy balance in this region in particular. It could be a PFT(=lakes) thing. I plotted iveg above but its not clear.

JhanSrbinovsky:
Yeah - definitely Tundra.

JhanSrbinovsky:
>i.e. there's something else going on which has allowed `%wbice > frozen_limit* %ssat` (equivalently `%wb > %ssat`)

Given that this is expected to be the case, then shouldn't getting into the loop in the first place be excluded. I wonder how many times it gets into the loop and there isnt a problem. 

Anyway, no sign of it in AM3. Curiously AM3 doesnt have these near 0 C Temps in the region. Its all 20-40 below.

har917:
@JhanSrbinovsky to pick up on a couple of things in here

- we need the `%tgg = %tgg - sicemelt*soil%zse(k)*Cdensity_ice*CHLF / REAL(ssnow%gammzz(i,k) )` line in there.  We get into that loop on occasions where evaporation/drainage reduces the liquid part of the soil water (in `surfbv`) so that `%wbice > frozen_fraction * %wb`.  It is is more likely to encounter that case now that we have different densities for liquid and frozen water.
- we shouldn't have been getting to `%wbice > frozen_limit*%ssat` 
- I would not be surprised if there are a small number of grid cells where the initial conditions in gridinfo are incompatible with the updated science.  Such cases will likely show up on the first time step.
- we've been a lot more careful around initialisation, especially around the application of `frozen_limit`, in the AM3 codebase - so again not that surprising that something has been triggered.
- I'd be happy if the new code ends up smoothing out any 'hotspots' or 'coldspots' 

----------------------------------
I'm not really sure about the way forward - it seems highly unlikely that we can create a codebase-gridinfo combination that preserves the previous offline performance while also not adding lots of adhocness to the codebase.

The most scientifically credible way forward is to either

1. include the `%tgg` line, take the initial temperature hit, run out for several years and then use that to create a new gridinfo file; or
2. do two phases: i) run a month or so with the `%tgg` line copied out (or wrapped in a if first_call condition), write to restart, then ii) add the `%tgg` line back in, rerun from restart for a while, then create a new gridinfo file.

1. is probably easier but does run the risk that the model doesn't come back to the same approximate climate. Either way we break comparability in a benchmarking sense. @ccarouge - any thoughts?

JhanSrbinovsky:
@har917 
> * we need the `%tgg = %tgg - sicemelt*soil%zse(k)*Cdensity_ice*CHLF / REAL(ssnow%gammzz(i,k) )` line in there.  

This being the case then I am in favour of below:

> 1. include the `%tgg` line, take the initial temperature hit, run out for several years and then use that to create a new gridinfo file; or

As overwhelming and dramatic as it may seem to re-create the grid info file 

Tthe grid info file is decades old anyway and no-one is really confident about where all the bits and pieces came from. 

Presumably we could correct this.

We should take this opportunity to establish an offline grid-info file that is reflective of the ACCESS online runs. There would be reasons to keep it in ACCESS resolution, however it might be better to adopt a more generic 1*1 degree resolution. 

It would be satisfying ti know what it is in the grid info file that is messing with us?

JhanSrbinovsky:
So this has been run out for 30-50 years in AM3. Is there someone at NRI that can create a grid info file @ccarouge ?

ccarouge:
@JhanSrbinovsky for the gridinfo file the only things we are working on are creating the file from UM output/restarts and creating the file using ANTS (that's far from being ready). Neither of these options is what you need here. So no, we don't have anything for what you want to do.

rml599gh:
A description of the gridinfo file I built based on ESM1.5 is here: https://forum.access-hive.org.au/t/cable-site-runs-with-access-forcing/2108/8 
I've only ever tested this in a single site set-up so have no idea if it would run a global case.

Whyborn:
I have built such a gridinfo file, based mostly off of Rachel's method in the forum post she linked. I ran a few spin-up stages of CABLE offline (on the CABLE-POP_TRENDY branch) using an "alpha" version of the gridinfo file, and it seemed to work fine. You can find the conversion script at [convert_from_fields.py](https://github.com/ACCESS-Community-Hub/Land-ancillary-creation/blob/CABLE-offline-as-ACCESS/ancillaries/CABLE_offline_as_ACCESS/convert_from_fields.py), which should just work out of the box in the ```hh5``` conda environment, assuming you have access to ```rp23```, ```fs38``` and ```vk83```.

Whether this gridinfo is something we want to use longer term, I can't say.

JhanSrbinovsky:
OK thanks @Whyborn , @rml599gh

 @Whyborn - I might try to use your grid info file as a test in this GSWP2 fiasco here. Can you please point me to it?

Whyborn:
The gridinfo I've run with is at ```/g/data/rp23/experiments/2024-03-12_CABLE4-dev/lw5085/CABLE-as-ACCESS/gridinfo_files/ACCESS-ESM1p5-1p875x1p25-gridinfo-CABLE_v0.1.nc```. This was on the ```CABLE-POP_TRENDY``` branch- I don't think there are major differences in the required gridinfo between ```main``` and ```CABLE-POP_TRENDY```?

har917:
@JhanSrbinovsky @Whyborn 
> I don't think there are major differences in the required gridinfo between main and CABLE-POP_TRENDY?

Probably not - however I think CABLE-POP (so TRENDY and BIOS applications) routinely uses a 4 tile gridinfo whereas as main will expect a 17 tile version.  I don;t know the offline i/o code well enough to advise whether this is an issue or not.

JhanSrbinovsky:
Unfortunately it doesnt like it. 

I'll see if I can use your python script. a couple of questions @Whyborn , 

1. (out of) Where did you get the fields mentioned
2. Was you CABLE run global?


 Nan in evap flux,        5219   65.50000      -17.50000
 fe nan        5219           2  2.8106968E-03  0.0000000E+00  0.0000000E+00
   220.3700      0.0000000E+00  0.0000000E+00   269.2953       7.291198
            NaN   946.4033                NaN            NaN            NaN
            NaN            NaN            NaN            NaN

JhanSrbinovsky:
reading the script it seems like an ESM1.5 dump is where you were grabbing things from, which is unfortunate as this is where I was thinking anyway, but it doesnt like something  :(

Whyborn:
It might be to do with ```iveg```? I'm a bit confused- do we treat ice as bare ground in CABLE offline? The default CABLE gridinfo appears to set most of Greenland and high arctic regions to be bare ground, while ACCESS-ESM sets those areas to the actual ice type (or at least appears to, based on the ancillaries). Do we have valid handling for the ice tile in offline?

JhanSrbinovsky:
Really, bare ground. Im not familiar with offline, I assumed it was treated as ICE (i.e. iveg=17). Im going to check GSWP, Im sure it was ICE but could be wrong. Maybe is just a TRENDY thing.

JhanSrbinovsky:
yeah GSWP treats Greenland as ICE and dodges the poles completely

rml599gh:
Not sure about Greenland but I don't think offline typically runs for Antarctica so the 1x1 gridinfo doesn't go that far south.
When I was making my access gridinfo, I did treat ice points differently. Even though there are values for the soil parameters in the restart file I didn't use these but overwrote them with the values for isoil=9 from def_soil_params.txt. I think this is what happens in ACCESS-ESM1.5 (another quirk that it would be good to clean up).  Hence my code:
```
           if (iveg(i,j).eq.17) then
                   isoil(i,j) = 9
                   rhosoil(i,j) = 910.
                   css(i,j) = 2100.
                   bch(i,j) = 7.1
                   clay(i,j) = 0.3
                   css(i,j) = 2100.
                   hyds(i,j) = 0.000001
                   sand(i,j) = 0.37
                   sfc(i,j) = 0.301
                   silt(i,j) = 0.33
                   ssat(i,j) = 0.479
                   sucs(i,j) = -0.153
                   swilt(i,j) = 0.216
           else
                   isoil(i,j) = 2
                   rhosoil(i,j) = 1600.
                   css(i,j) = 850.
           endif
```

har917:
This issue of ensuring consistency between `iveg` and `isoil` and how to fill in the soil/pft values in the offline appliciation has been a focus of some recent work (linked to GW model implementation) - see #338.

I think we're okay offline but the initialisation in ESM1.6 may need to be looked at again.

Whyborn:
@JhanSrbinovsky  I've made some fixes to the gridinfo generated from the ACCESS-ESM1.5, and now it seems to run for a few select points without producing ```NaN```. Had a quick look at ```GPP``` and ```Ebal``` and the results seem reasonable? The new version of the gridinfo is at ```/g/data/rp23/experiments/2024-03-12_CABLE4-dev/lw5085/CABLE-as-ACCESS/gridinfo_files/ACCESS-ESM1p5-1p875x1p25-gridinfo-CABLE_v0.3.nc```..

JhanSrbinovsky:
Thanks @Whyborn , I'll give it a shot

src/science/soilsnow/cbl_snowAccum.F90
```diff
@@ -47,6 +47,8 @@ SUBROUTINE snow_accum ( dels,  canopy, met, ssnow, soil )
 
              ssnow%tgg(i,1) = ssnow%tgg(i,1) + canopy%precis(i) * CHLF               &
                   / ( REAL( ssnow%gammzz(i,1) ) + Ccswat *canopy%precis(i) )
+             ssnow%dtmlt(i,1) = ssnow%dtmlt(i,1) + canopy%precis(i) * CHLF           &
```
har917, line 50: these are new terms from AM3 development. they are the counterpoint to the melt terms to `%dtmlt` and come about by the freezing of liquid precip onto frozen ground or snow pack. 

good that they've been picked up - but not that surprising given they sit in a different part of the code base to the rest of the ice fixes

JhanSrbinovsky:
Given the UM7 debacle, I've turned to a long standing debacle. This one is interfering with the effort to get the code the same across apps. FastTrack already has them  and they came out of AM3. Alternative global offline tests seem plausible BUT the problem is very likely in the grid info file. What is not clear at all is how the grid info file Lachlan made from ESM1.5 either still has a problem, or crashes. Anyway, further evidence to suggest the problem is in the grid info file. I dumped a restart from the 1986 run. And fed it back into to a re-run of the 1986 forcing. The problem more or less persists. Looking at the SoilTemp from the day before, the old-new might be 6-7K. The next day it's more like 60K. That 30K difference was the ave for Jan (old-new).

JhanSrbinovsky:
One possibility is that "the bug" in grid info was present in ESM1.5 and we need to redo it from ESM16 output, but Im not confident this is it at alll. @Whyborn  Is it possible that there is a problem with extrapolating to GSWP cells (excluding poles) from all globe grid info 192*145.? I suspect it might even be an initialixzation that is hard-coded and got by in the past

Whyborn:
I assume you are running from the ```main``` branch of CABLE? I managed to do a first stage of a TRENDY run from the ```CABLE-POP_TRENDY``` branch using the new gridinfo file, and it didn't throw NaNs anywhere at least. I'm working on getting a configuration working with the ```main``` branch.

JhanSrbinovsky:
In this caseI'm working with both 494-merge-soilsnow-from-am3 and main.  I have a TRENDY setup somewhere. I'll see if I can find it. I have a bad feeling though that it was on /scratch/ and has long since disappeared. I might have to get back to you about that. If we can run global offline with the "improved" grid info that should get it across the line

JhanSrbinovsky:
The only TRENDY forced stuff I have is:
 1. using CABLE-POP (src_pop/), so it won't work with main:src/. 
 2. Something from Ying-Ping 2or3 years ago I grabbed off mass data tape but as suspected it doesnt work. I imagine it has its own drivers to read the v5 CRU forcing and won't work with our CABLE.
 
Can we use your version to test whether this Siberian bullseye in SoilTemp occurs with your grid info? @Whyborn Just to clarify, is it just re-gridding or the value come from ESM1.5?

Whyborn:
@JhanSrbinovsky Not sure exactly what you mean by re-gridding or the value- the gridinfo I've been trying to build is at the ACCESS-ESM1.5 resolution (1.875degx1.25deg) using the values from the ACCESS restart. I've followed the process described by Rachel [here](https://forum.access-hive.org.au/t/cable-site-runs-with-access-forcing/2108/8).

Where can I find the CABLE configuration you're using for testing, and are you using the same configuration for ```main``` and this branch? Almost all of my experience is with configurations on the ```CABLE-POP_TRENDY``` branch, I'm not sure how much difference there is between that and ```main```.

JhanSrbinovsky:
@Whyborn OK - I remember reading that you followed @rml599gh 's use of the ACCESS restart (= ACCESS values in my previous unclear post). 

My cable configuration is here:

`/home/599/jxs599/gdata/494_AM3_soilsnow/validate/GSWP2_template/cable.nml
`

Although I dont think t hat will help. It's the source code that is different between models. 

This is an executable from CABLE main corresponding to the branching point of the other case. I suspect it won't work as CABLE-POP has specific reading code to handle TRENDY. It could though if we are lucky :) 

`/home/599/jxs599/gdata/494_AM3_soilsnow/validate/GSWP2_template/cable-mpi
`

JhanSrbinovsky:
Source of error reading ne grid_info was output - not input.  Commenting out this error check implemented by Vanessa lets code run with @Whyborn  's revised grid_info. Yet to plot or review code commented but not a particularly satisfying solution (although I "think" this has already been commented out in serial version). [Troublesome code](https://github.com/CABLE-LSM/CABLE/pull/497/commits/c2ee11d87ac485c7c50a3500130f67a7b56857d4) (@har917 , @rml599gh )

har917:
> Source of error reading the grid_info was output - not input. Commenting out this error check implemented by Vanessa lets code run with @Whyborn 's revised grid_info. Yet to plot or review code commented but not a particularly satisfying solution (although I "think" this has already been commented out in serial version). [Troublesome code](https://github.com/CABLE-LSM/CABLE/pull/497/commits/c2ee11d87ac485c7c50a3500130f67a7b56857d4) (@har917 , @rml599gh )

@JhanSrbinovsky Surely 'having to comment out this error check to get past a failure' indicates that at some point (time and tile) `canopy%fe` has gone to NaN?  That's a signal that something else is wrong.

JhanSrbinovsky:
> @JhanSrbinovsky Surely 'having to comment out this error check to get past a failure' indicates that at some point (time and tile) `canopy%fe` has gone to NaN? That's a signal that something else is wrong.

@har917 Hence the comment "not particularly satisfying". The link to the code was hopefully going to inspire a comment from yourself or Rachel that commenting out here lines is valid because ...... . Possibly it would be the same reason that someone(?) commented this out in cable_serial!?! Given the commenting style of "!$", I suspect it Wass Vanessa herself. Her reasoning surely would,e been that it was reasonable. Why - I have no idea. I was hoping someone more familiar with offline would know. I am skeptical that it is a real science problem given that this is likely many generations of CMIP old.

har917:
@JhanSrbinovsky I'm actually struggling to see what case is caught by this code, that wouldn't cause the model to crash elsewhere when the code is removed.  This is what you've effectively found.  

In principal I'd be in favour of removing this (and the equivalent code in offline serial) from the codebase - it looks very much like debug code that shouldn't be in there - but we need to understand what's going on as a safe guard.

JhanSrbinovsky:
> @JhanSrbinovsky I'm actually struggling to see what case is caught by this code, that wouldn't cause the model to crash elsewhere when the code is removed. This is what you've effectively found.
> 
> In principal I'd be in favour of removing this (and the equivalent code in offline serial) from the codebase - it looks very much like debug code that shouldn't be in there - but we need to understand what's going on as a safe guard.

@har917  - yeah agreed. I just got back to this. So far all I have found is that it never happens at Tumba or Ampero. Strangely Hytialia - I cant run. I wouldn't be entirely surprised if it is happening at some location where we dont have a single-site anyway and I am kinda hoping that it is happening on the first timestep.

har917:
Possibly irrelevant/a different issue:  There are a whole load of `REAL(...)` operators applied to variables through the code.  This shouldn't be necessary when those variables are defined as `REAL ::`* - but clearly at some point through the development of CABLE this was found necessary.

a thought:  when you ran the MIP-case with the code in at it crashed with these error trapping lines in (TBH more likely triggered the `MPI_ABORT()`) - did you look at the logs to see where it happened and what the other outputs were?  This would be a good way of narrowing things down a bit.

*only legitimate usage I can see is when we need to do things like `a = b * REAL(INTEGER variable)` or `c = REAL(complex-valued d)` to make sure that the KINDs are correct.

JhanSrbinovsky:
Are you sure they dont have ( , r_2) at the end?

Anyway, an other aside. I rebuilt cable-mph in debug mode. It crashes in cable_parameters at t=0. IDK if this is just GSWP2 thing or happens generally? @ccarouge

JhanSrbinovsky:
interesting - the NaN crash using @Whyborn ;s revised grid_info occurs in the same region as the original crash. On an island off the coast of Siberia. It occurs at dt=1. Artificially disallowing the test unit dt=1111, it still crashes

har917:
can you check the `soil%isoil` and tile index for that point(s)?

There have been some changes recently in how consistency between the two is enforced - #349 and #491 - and I'm wondering whether something isn't quite correct with the revised gridinfo file.

JhanSrbinovsky:
@har917 - there was a mismatch.

set 

cable_user%l_ice_consistency = .true. 

now pft=17, soil=ice but it still dies there

JhanSrbinovsky:
what strikes me though is that valid, rad%fvlai = 0.0 BUT then again - big deal PFT=soil=ice. So canopy%fe should be zero right? So maybe fe is not initialised to zero and remains zero unless a finite canopy response overwrites it?

JhanSrbinovsky:
@har917   - so I zeroed fe, fes, fev on first call and then printed them after dt=1. At trouble spot Fev is 0.0 as expected fee is NaN. so we're narrowing it down. potev, tgg are 0 as well.

har917:
> So `canopy%fe` should be zero right? So maybe fe is not initialised to zero and remains zero unless a finite canopy response overwrites it?

`canopy%fe` is the total latent heat flux - so includes contributions from both vegetation `%fev` (wet and dry leaf components) and soil (`%fes`) - all of these are initialised in `cable_parameters`.  If I understand the code tracing `%fes` gets set in `define_canopy` before it is used (in `update_zetar`) and definitely should have a value by the time it gets out of `cbm`.  

Basically I don't understand how the code is failing at that point in mpi_master - except is there something weird going on with the MPI code and one or other of the master/worker side exchanges isn't set up properly? 

now - seeing your comment - have we really got `%tgg=0.0`?  I understand all/most of the temperatures are in K so it should be in the range 260-300 at initialisation and should be coming from the restart file. 

however - since we're enforcing consistency between `%isoil` and `%iveg` we may be triggering an entirely new tile which wouldn't/may not have an initialisation in the restart file.

JhanSrbinovsky:
Sorry I meant NaN (tgg, potev)

har917:
okay `%tgg` and `%potev` being NaNs indicate that something upwind isn't quite correct - and this is consistent with `%fe` and `%fes` being NaNs.

Can you check what `veg%hc`, `rough%z0soilsn` and `rough%z0` are doing?  (The main suspect is `%rtsoil` - and possibly `%tss` or `%qstss` - this depends on these 3)

I'm wondering whether we've managed to get it (with the change in surface types) into a state where one part of the code thinks a tile is vegetated and other parts think it's not.  I can't see how `consistency_ice_veg_soil` could get us into that state but worth checking.

JhanSrbinovsky:
@har917  - I haven't looked at height and roughness etc yet.what would hc have to do with this on ice? Anyway, I suspect tss etc will be screwed up.  I was looking at what kills the debug run in cable_parameters. The value of ssat is zero at many points, and at our troublesome point. ssat is always finite in the nml file, but I suspect it is getting this from grid_info. ASSUMING that all the points where it messes up ssat are glacial, clobbering ssat with the ice value from our nml, it crashes at a later point in cable_parameters. Actually I'm finding a bunch of parameters that are zero. I wager that if I used the nml values it would at least get passed this build=debug issue. Another thing I could try is if this is still a problem with the old grid_info. I did consider that how is this working in the UM? I thought perhaps myassumption of these being ice points could be right. Although, I dont know how JULES manages with ice points then. Anyway, quick test seemed to suggest this was indeed the case.

JhanSrbinovsky:
I clobbered ssat for ice=9 and it indeed got past the initial crash. The next was one also involving a soil parameter. I decided to just use clobber all the ice points as we do in the UM, which would make sense as thats where @Whyborn 's grid info is coming from. Playing in cable_parameters to see where I should do this I cam across this little gem (see file differences from this commit). Someone has been here before with this problem. As per my inline comment some of the values line up with those in our nml file. Some have a units conversion which I can believe. But some are just whacky. Like for exampl e.g. the density =1455. How does ice end up as dense as clay!?!

JhanSrbinovsky:
BTW - crash in debug goes away with these clobbering :)

har917:
> what would hc have to do with this on ice? 

I am/was wondering whether in enforcing consistency between PFT and soil we've managed to get some weird combination of parameter values at the land point concerned.  hc is important as it's critical in determining whether CABLE-roughness thinks we have a vegetation canopy or not - and that is a control on `rtsoil`, then potev (which has gone to NaN).

However - you've identified that `ssat = 0.0` at that point.  This will cause NaNs all over the place (there's lots of xx/ssat through the soil_snow code). 

Now if I look at `consistency_ice_veg_soil` again in the sections that it checks whether a land point is soil_ice (so line 2500 and line 2523) the code is only overwriting `soil%rhosoil` and `soil%css` - but it's not overwriting the other soil parameters (`ssat` etc.).  I think this is wrong/incomplete - @ccarouge your thoughts?

This perhaps explains why a value of `ssat = 0.0` has gotten through the consistency check but I don't understand how it got a value of 0.0 in the first place - since it should have picked up a realistic value when it was set as non-ice soil originally.

JhanSrbinovsky:
> This perhaps explains why a value of `ssat = 0.0` has gotten through the consistency check but I don't understand how it got a value of 0.0 in the first place - since it should have picked up a realistic value when it was set as non-ice soil originally.

I think all the zeroes are coming from the grid info file which is from the UM. We clobber ice points, but we dont pass them back so presumably this might be all those zeroes. JULES must be doing something similar and have ice point specific values, otherwise I dont know how they mange to avoid  all the potential sources of (fatal( error

JhanSrbinovsky:
whoaaaaa - it doesnt help us at all with our original problem - it spreads it out - this is day one only - where the original magnitude of the bullseye on day 1 was ~66K

Screenshot 2025-01-14 at 1 48 34 pm

JhanSrbinovsky:
also keep in mind that this is the ESM1.5 grid info

JhanSrbinovsky:
woo hoo 👯 using branch main at the branch point of 494 VS 494 and this @Whyborn  grid info for both:

Screenshot 2025-01-14 at 3 03 23 pm


This is DAY one! I'll make up some more plots but I am hopeful that even averages over a month will calm this down. I dont think this exceeds what might be expected weather wise from the ice fixes? @har917 I've also got it restarting from a restart using the old grid info so I'll try to clean t his up as well.

JhanSrbinovsky:
@rml599gh - it has become clear that ice points are misrepresented using the grid info file coming out of ACCESS. Possibly you have encountered this and have made adjustments already? - but this would certainly impact online/offline comparison

har917:
@JhanSrbinovsky This is looking a lot more reasonable - the differences are mainly in regions where we expect melt and/or rain onto snow to occur - and that was one part of the ice fixes changeset.

I'd expect different (larger?) differences to emerge over time since that will involve soil moisture dynamics with different amounts of water around.

@rml599gh @ccarouge Given that the offline/online comparisons have focused on performance over vegetated tiles I'm not so sure that this is an issue for that work.  But we do need the gridinfo that's aligned to ACCESS-ESM to have appropriate values for the soil properties given that we may want to do LSMIP-style runs and use offline-as-ACCESS for spin up purposes in ESM3.

@JhanSrbinovsky I am a little lost with what additional code has had to go in here - in particular around the values for the soil parameters that have been hard-wired.  These should be the same as the Greenland values that exist in the default gridinfo file.

JhanSrbinovsky:
> @JhanSrbinovsky I am a little lost with what additional code has had to go in here - in particular around the values for the soil parameters that have been hard-wired. These should be the same as the Greenland values that exist in the default gridinfo file.

Thats what I was expecting. Possibly @rml599gh has some memory of them.  

Im getting lost with latest plots showing bizarre things. Have to go pick up Lena now

rml599gh:
When I was writing my ACCESS-based gridinfo for my offline testing, I did have a special case in the ice grid-cells. I assume I did this based on what was in the UM code for ESM1.5 which, from memory, just assumed one particular soil type under all ice grid-cells and then used the relevant soil parameter values (from or the same as in def_soil_params.txt.
Here's the relevant code from my 'writegridinfo.F90'

```
           if (iveg(i,j).eq.17) then
                   isoil(i,j) = 9
                   rhosoil(i,j) = 910.
                   css(i,j) = 2100.
                   bch(i,j) = 7.1
                   clay(i,j) = 0.3
                   css(i,j) = 2100.
                   hyds(i,j) = 0.000001
                   sand(i,j) = 0.37
                   sfc(i,j) = 0.301
                   silt(i,j) = 0.33
                   ssat(i,j) = 0.479
                   sucs(i,j) = -0.153
                   swilt(i,j) = 0.216
           else
                   isoil(i,j) = 2
                   rhosoil(i,j) = 1600.
                   css(i,j) = 850.
           endif

```

JhanSrbinovsky:
> When I was writing my ACCESS-based gridinfo for my offline testing, I did have a special case in the ice grid-cells. I assume I did this based on what was in the UM code for ESM1.5 which, from memory, just assumed one particular soil type under all ice grid-cells and then used the relevant soil parameter values (from or the same as in def_soil_params.txt. Here's the relevant code from my 'writegridinfo.F90'

These are the same values from our nml file but not the ones that someone has added (BLAME doesn't recognise this suggesting it was in the code 3 years ago). In either case they were not necessarily the values in cable_parameters where it was crashing. 

I've been wondering where the code came from. It is possible that I did it during HAC/offline comparisson. I dont specifically remember it but there was parameter clobbering at some stages.

JhanSrbinovsky:
Cant believe that yet again I have to re-write this- grrrr:

The crank values I was seeing was due to a versioning problem, which I am more than happy to share if you’re interested.

I have created a new branch from the main branch as of last night. 494_redux_local. This isolates the AM3 ice fixes. Necessary compensation to use the ESM1.5 gridinfo file is the altered cable_parameters.F90. For your perusal this is committed to 494_redux_local, and present in my local copy of main.  Comparison isolates impact of ice fixes.

Screenshot 2025-01-15 at 8 19 47 am


I thought I could change branch in the PR, Im sure I ve done this before, anyway goto #527 to see Files changed.

ccarouge:
@JhanSrbinovsky I don't follow your last comment. Are you saying this current PR is now useless, it has been superseeded by #527 and this PR (#497) should be closed? If so, can you close it?

To change the branch at the top click the edit button next to the title and then click the branch name.

JhanSrbinovsky:
> @JhanSrbinovsky I don't follow your last comment. Are you saying this current PR is now useless, it has been superseeded by #527 and this PR (#497) should be closed? If so, can you close it?
> 
> To change the branch at the top click the edit button next to the title and then click the branch name.

That was instinctively how I tried to do it but it didnt seem to work. I'll try again and close #527 if successful

JhanSrbinovsky:
Nope - every time after edit if I click on the branch it t takes me to the branch. The option to change the branch to compare to is there

JhanSrbinovsky:
The branch is a mess because:

The feature branch 494-merge-soilsnow-.... branched from main ~6 weeks ago. When I went to rebase, it rebased local main which was still at 6 weeks ago. It doesnt want to rebase again because it thinks it has already done it. It was easier to rebranch  from main and merge 494-....... which is what rebase does anyway.

Whyborn:
I've made a new gridinfo with the ice fixes applied as per Rachel's comment. Comparison between soil saturation (```ssat```) in the old gridinfo vs the new (note the difference in scales):

Previous gridinfo ```ssat```:
SSAT_CABLE_Offline_gridinfo_noicefix

Gridinfo ```ssat``` with ice fix:
SSAT_CABLE_Offline_gridinfo_icefixed

The new gridinfo file is at ```/g/data/rp23/experiments/2024-03-12_CABLE4-dev/lw5085/CABLE-as-ACCESS/gridinfo/ACCESS-ESM1p5-1p875x1p25-gridinfo-CABLE-withicefix.nc```. Something happened to all the NetCDF metadata which I'm working to fix, but I don't think CABLE looks at the gridinfo metadata so it should still work?

har917:
@JhanSrbinovsky It's probably worth testing on GSWP2/3 with the @Whyborn's updated grid but without the (re)implemented lines `cable_parameters` and the removed error catch in `cable_mpimaster`  (i.e. revert the codebase to just include the AM3 soilsnow ice fixes code).

Hopefully getting the parameter layers in the gridinfo correct will solve both problems.

We could look to refine later (especially to remove the error catch) - but it would be good idea to focus the PR.

JhanSrbinovsky:
sorry just got distracted by a super important phone call - apparently I've won $US1000 in crypto. I just have to l.og into some random website and collect :) - back from lunch - Yeah im confident this will work out of the box and the ice fixes can go in independent of any other code change

JhanSrbinovsky:
to be clear - when I say ice fixes Im referring to the code changes in soilsnow/ , not the same ice fixes @Whyborn is referring to

JhanSrbinovsky:
yep. It's the equivalent of what I had this morning.

JhanSrbinovsky:
moving this to #527

PR #501, opened 2024-12-03
reconcile ESM16: CABLE and CABLE:main

### LINKED ISSUE #500: trivial modifications to reconcile ESM1.6:CABLE and CABLE:main
This is the same CABLE sd CABLE:main except: PR to be submitted to main to rm double entry of line:

USE cable_common_module 
in src/science/soilsnow/cbl_surfbv.F90.

remove IF condition around"

IF( .NOT. cable_runtime%um ) THEN
      canopy%cansto =  canopy%oldcansto
ENDIF

in cable_canopy as AM3 comments out the IF condition anyway. Frankly I dont know that we need to as oldcansto is initialised in the "drivers" BUT we need to follow through the intent here as canopy is called multiple times per timestep. @har917 could missing water be the double implicit call and (re-)initialising this?

AND

Not addressed in this issue at all:

PR#497 in CABLE:main to include the developments from AM3 in soilsnow. Mostly to date these are dual densities for liquid/ice water. Single site was unremarkable. Silly me, I did a GSWP2 test which showed a bullseye in Northern Russia. We dont understand why yet, so halting the merger for now. More later.....

AND casa-cnp is not merged. We'll have to think about this. The casa-cnp from AM3 is operational but completely untested. That in CABLE:main is being used offline and in Rachels testing. The AM3 related differences could be trivial anyway, I haven't looked yet.

JhanSrbinovsky:
# CABLE

## Description

Reconcile ESM16: CABLE and CABLE:main excludes #497  and divergent casa-cnp/. Super quick

Fixes #500 

## Type of change

- [x] Bug fix

## Checklist

----
📚 Documentation preview 📚: https://cable--501.org.readthedocs.build/en/501/

### DIFF
--- a/src/science/canopy/cable_canopy.F90
+++ b/src/science/canopy/cable_canopy.F90
@@ -166,10 +166,8 @@ REAL :: soil_conductance(mp)
 REAL :: Surf_conductance(mp)
 ! END header
 
-    ! Not sure that this is appropriate for JULES standalone - HaC either
-    IF( .NOT. cable_runtime%um ) THEN
-      canopy%cansto =  canopy%oldcansto
-    ENDIF  
+canopy%cansto =  canopy%oldcansto
+
     ALLOCATE( cansat(mp), gbhu(mp,mf))
     ALLOCATE( dsx(mp), fwsoil(mp), tlfx(mp), tlfy(mp) )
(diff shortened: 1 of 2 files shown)


ccarouge (approved):
All good.

Issue #496, opened 2024-11-27
dewmm and cansto units change


JhanSrbinovsky:
In merging cable_canopy from AM3 there was a curious units change to both dewmm and cansto that I did not recall. BLAME in the ACCESS3_JULES repo attributes the change to :

https://github.com/CABLE-LSM/ACCESS3-JULES/commit/496bbcfcfd611b3d48314ac8912a222f769c62cf

This should be ported to main as well.

har917:
@JhanSrbinovsky this will be important for [ESM1.6 #56](https://github.com/ACCESS-NRI/UM7/pull/56)

PR #474, opened 2024-11-13
changes required for ESM16

JhanSrbinovsky:
# CABLE

Thank you for submitting a pull request to the CABLE Project.

## Description

With these minor changes CABLE:main can be cloned directly into the ESM1.6 / UM7.3 code under umbase_hg3. NB:A lot of the changes are replicated from PRs that are waiting to be reviewed.

Fixes #473, #468

## Type of change

For ESM1.6. A library can be built directly from this code

benchcab regression tessts successful:  
[benchmark_cable_qsub.sh.o128621374.txt](https://github.com/user-attachments/files/17748080/benchmark_cable_qsub.sh.o128621374.txt)

----
📚 Documentation preview 📚: https://cable--474.org.readthedocs.build/en/474/

Issue #162, opened 2022-11-09
moving the time stepping of canopy water content out of the canopy (energy balance) routine


penguian:
`keyword_keepgit` `owner:jxs599@nci.org.au` `type_JAC readiness`   |    by inh599

___

In preparation for JAC we would be advised to modify the structure of CABLE so that appropriate parts of the model can be called cleanly.  At the moment the canopy water content state variable (canopy%cansto) is time stepped within the cable_canopy routine - this is the only state variable that is updated this way, all the rest are updated in soil_snow, SLI, CASA-CNP etc.  

In JAC we anticipate that cable_canopy will be called multiple times per time step and will **only** need to evaluate the surface energy balance (and associated fast flux of carbon).  Correspondingly we need to reorganise the code so that the time stepping of the canopy water content can be completed separately from cable_canopy - while maintaining conservation and current performance.

A suggested method to do so is (in some detail based off the trunk at 9037)
1. remove the update of %cansto at line 49 of cbl_SurfaceWetness
2. update the calculation of %fwet at line 52 of cbl_SurfaceWetness to use %cansto + %wcint
3. modify lines 854-879 of cable_canopy to remove the updates to %cansto while retaining the correct evaluation of the flux terms %spill, %through and %precis.  Note that %wcint needs to be added to %cansto at various points when evaluating the fluxes to preserve the intent (because of change 1. above)
4. remove the evaluation of %delwc (line 879) from cable_canopy
4. create a new SUBROUTINE to time step %cansto and evaluate %delwc. This would be called immediately before soil_snow in cbl_model_driver_offline, before the call to soil_snow in cable_implicit_driver (in ESM1.5 and CM2) and before the call to soil_snow in the WRF and CCAM driver routines.

Connections/interdependence with the ground water model and SLI may need to be followed through.

In CM2 %cansto would still need to be placed within the prognostic bank structure.

There are other 'working state variables' that are updated in a similar manner within cable_canopy - canopy%rtsoil, canopy%fes_cor and canopy%ga (if the Penman_Monteith option for latent heat flux from the ground is used).  These are variables whose values (currently) need to be updated correctly from time step to time (e.g. in order to conserve) but this is a result of the algorithm design not because they represent a biophysical state of the land. JAC will need to be mindful of (i.e. address) these working state variables at some point in the development cycle.

___
_Issue migrated from [trac:352](https://trac.nci.org.au/trac/cable/352) at 2023-11-27 11:43:56 +1100_

PR #506, opened 2024-12-09
update logic surrounding timestepping of cable%cansto / %oldcanst

### LINKED ISSUE #505: time stepping of canopy%cansto.
> @har917 it isnt active in the AM3 model, or the CMIP6 version of ESM1.5. It is blocked by the IF condition. Although it is active in the ESM1.5+CABLE3 version. Probably because offline was doing it. I think your 1,2,3 are in place, I'll have to double check and get back yo you. For 1 however I think my intention was to move it back to offline/cbm where the time-stepping in cbm is the same as in define_canopy anyway and have the same shared canopy for everyone? 

 _Originally posted by @JhanSrbinovsky in [#500](https://github.com/CABLE-LSM/CABLE/issues/500#issuecomment-2526555426)_

JhanSrbinovsky:
# CABLE

## Description

Clarifying time stepping of cansto / oldcansto

Fixes #505

## Type of change

- [x] Bug fix

----
📚 Documentation preview 📚: https://cable--506.org.readthedocs.build/en/506/

### DIFF
--- a/src/coupled/ACCESS-CM2/control/explicit/cable_explicit_driver.F90
+++ b/src/coupled/ACCESS-CM2/control/explicit/cable_explicit_driver.F90
@@ -296,7 +296,6 @@ USE cable_phys_constants_mod,  ONLY: density_liq, density_ice
   !CM2!IF(l_vcmaxFeedbk) call casa_feedback(ktau_gl,veg,casabiome,casapool,casamet)
   !CM2!IF(l_laiFeedbk) veg%vlai(:) = casamet%glai(:)
 
-  canopy%oldcansto=canopy%cansto
   rad%otrad = rad%trad
 
   !---------------------------------------------------------------------!
@@ -305,7 +304,7 @@ USE cable_phys_constants_mod,  ONLY: density_liq, density_ice
   CALL cbm( ktau_gl,timestep, air, bgc, canopy, met, bal,                      &
             rad, rough, soil, ssnow, sum_flux, veg, climate )
   
-  !-------- End Unique subroutine body -----------
+  canopy%cansto=canopy%oldcansto
 
   return
 
--- a/src/coupled/ACCESS-CM2/control/implicit/cable_implicit_driver.F90
+++ b/src/coupled/ACCESS-CM2/control/implicit/cable_implicit_driver.F90
@@ -306,8 +306,6 @@ USE cable_phys_constants_mod,  ONLY: TFRZ, CAPP
   met%tvair = met%tk
   met%tvrad = met%tk
 
-  canopy%cansto = canopy%oldcansto
-
   CALL cbm( ktau_gl,timestep, air, bgc, canopy, met, bal,                             &
             rad, rough, soil, ssnow, sum_flux, veg, climate )
 
@@ -419,7 +417,7 @@ subroutine cable_store_prognostics()
   PB(ipb)%snow_age  = ssnow%snage
   PB(ipb)%snow_flg3l= ssnow%isflag
   PB(ipb)%snow_tile = ssnow%snowd
-  PB(ipb)%ocanopy   = canopy%oldcansto
+  PB(ipb)%ocanopy   = canopy%cansto
   !Jan 2018 new PB variables
   PB(ipb)%fes_cor   = canopy%fes_cor
   PB(ipb)%puddle    = ssnow%pudsto
@@ -462,7 +460,7 @@ SUBROUTINE cable_reinstate_prognostics()
   ssnow%snage   = PB(1)%snow_age
   ssnow%isflag  = PB(1)%snow_flg3l
   ssnow%snowd   = PB(1)%snow_tile
-  canopy%oldcansto = PB(1)%ocanopy
+  canopy%cansto = PB(1)%ocanopy
   !Jan 2018 new PB variables 
   canopy%fes_cor = PB(1)%fes_cor
   ssnow%pudsto  = PB(1)%puddle
--- a/src/coupled/AM3/control/cable/interface/explicit/cable_explicit_driver.F90
+++ b/src/coupled/AM3/control/cable/interface/explicit/cable_explicit_driver.F90
@@ -274,6 +274,8 @@ CALL update_data( row_length, rows, land_pts, nsurft, npft, sm_levels,         &
 CALL cbm_expl( mp, nrb, timestep_number, timestep, air, bgc, canopy, met, bal, &
           rad, rough, soil, ssnow, sum_flux, veg, climate )
 
+canopy%cansto =  canopy%oldcansto
+
 RETURN
 END SUBROUTINE cable_explicit_driver
 
--- a/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_explicit_driver.F90
+++ b/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_explicit_driver.F90
@@ -389,15 +389,14 @@ integer :: j
    IF(l_vcmaxFeedbk) call casa_feedback(ktau_gl,veg,casabiome,casapool,casamet)
    IF(l_laiFeedbk) veg%vlai(:) = casamet%glai(:)
 
-   canopy%oldcansto=canopy%cansto
-
-
    !---------------------------------------------------------------------!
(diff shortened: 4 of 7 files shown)


har917:
@JhanSrbinovsky @ccarouge 

Headline comment:  This looks okay - the science looks correct and it would certainly improve the consistency in logic between the different applications.  It's a sensible half-way stop as part of the broader effort to address #162 #496.

However - one big concerns to check the approach on and one request for change.

concern:  I'm not sure that I like edits to files in the `coupled/CM2` and `coupled/ESM1.5` directories - I would have thought these are now read only.  However, you can't make the edit to `cable_serial` and `cable_canopy` and keep the codebase functional without those changes.  So I'm at a bit of an impass on the way forward.

request for change: we need some comments in the code to describe what's going on better 

----------------------
In CM2/cable_explicit_driver, AM3/cable_explicit_driver and ESM1.5/cable_explicit_driver ahead of the new `%cansto` line add

!cbm/define_canopy() sets the value of %oldcansto then updates the value of %cansto
!we reset to value of %cansto to that at start of timestep here to work with the split structure of the ACCESS calls to CABLE


In cable_canopy ahead of new %cansto line include

!take a copy of the value of %cansto at the start of time step.  This is used to reset %cansto part way through ACCESS timestep
!and in evaluation of some water fluxes.

JhanSrbinovsky:
@har917 - hrmmm - you're right, but I have no idea of the way forward. Perhaps, and it will have to happen at some stage, MIPX models of tester-year will just need to be tagged and left alone.

ccarouge:
> concern: I'm not sure that I like edits to files in the coupled/CM2 and coupled/ESM1.5 directories 
This is because #474 was merged into src/coupled/ESM1.5 instead of src/coupled/esm16. I've opened issue #507 and asked Ben to sort this out. Changes for ESM1.6 should go into src/coupled/esm16 (which hopefully can be renamed for consistency).

And I have no idea why CM2 need to change? CM2 does not run with CABLE3 so it should not reflect changes for the latest CABLE.

har917:
@JhanSrbinovsky @ccarouge Another option would be to add conditions around `frozen_limit`, %wbice`, `%wb` and `%ssat` that reflect the updated use of the different densities in the initialisation section(s) - i.e. constrain the starting conditions to abide by the updated physics.  The key difference is that we wouldn't have to maintain conservation at that point.

This would, hopefully, avoid the initial temperature shock that you're seeing (which I still think is coming the restart not being in balance with the physics and showing up in `surfbv`).  Given that we're breaking bitwise comparability anyway I don't think it's too much of an issue.

The question is where would it need to go in, presumably
- offline: in between the restart read and first time step
- coupled: in the `init_soil` part of the initialisation routines

Head's up: the work that @rkutteh has been doing (both the work to ensure consistency between soil type and veg type, and the GW work) could well intersect with this awkwardly.

JhanSrbinovsky:
@ccarouge - true (I think). I have run CM2 semi-recently but I am not sure of the details. An argument though for just removing the coupled/CM2 directory completely?

@har917  - I'll have to double check but I don't remember having a restart for this GSWP2 run, although GSWP2 was the new kid on the block when I set this up so I might've just forgotten.

ccarouge:
@JhanSrbinovsky We need some test results with this. What's the impact on simulations?

JhanSrbinovsky:
Ive run this  through  benchcab. Surprisingly reproducibility fails. I can believe a rounding error perhaps, so I will have to plot the data, but my understanding is that it was just nodified so we could use the same time-stepping logic across apps, although the original time-stepping was not wrong per se.

har917:
I'm a bit surprised this has failed on reproducibility given that I thought that we default to using `cansat = 0.0`.   In that case this whole section of code should be effectively redundant (lots of 0.0 = 0.0 lines but no net impact).

However my bigger concern here is that there are changes to the ESM1.5 and CM2 `*driver*` files in this change set - I thought/hoped that we'd agreed to not modify the source code of past applications in the git repos and focus only on AM3 and ESM1.6.

More generally this area of science needs to be reviewed as part of [cansto time stepping](https://github.com/CABLE-LSM/ACCESS3-JULES/issues/74) and the refactoring of the code base.

JhanSrbinovsky:
hmmmm - I thought this was the conclusion if the review. Anyway, the coupled changes herein are likely in those apps already. They possibly appeared here before the decision had been made. I'll make sure that the coupled changes are indeed already in those apps and removed them. That shows up how crazy this is that it is not reproducible.  We've moved the single line from outside canopy to inside canopy

har917:
> I'm a bit surprised this has failed on reproducibility given that I thought that we default to using `cansat = 0.0`. In that case this whole section of code should be effectively redundant (lots of 0.0 = 0.0 lines but no net impact).

Correction - it appears that (in offline and ESM1.5 at least) we are using `cansat = veg%canst1 * veg*vlaiw` with `veg%canst1 = 0.1` (irrespective of the surface type!!)

JhanSrbinovsky:
@har917 dint we already do this in the Xmas lead up?

har917:
@JhanSrbinovsky Yes we did a lot of work in this.  There are two elements to this topic.

1. changes in ESM1.6/AM3 to sort out the time stepping
2. Hoped for changes to CABLE3 MAIN so that `%cansto` stops being the odd-state-variable-out - see #162 - which would ripple down to ESM1.6 and AM3.

Most of this PR is about 1. and should be in ESM1.6 - but as it's in the coupling code I don't know whether we sync'd back to MAIN - see also #505.  I'd be nervous about merging - especially the difference `cable_canopy`.

JhanSrbinovsky:
So is it correct in any of the applications (main/fastrack/AM3)?

har917:
@JhanSrbinovsky @ccarouge I think they are all correct at the moment - but highly confusing and could be rationalized.  What that opinion implies is that this PR isn't needed - except as the means to sync the different repos - but probably worth a second opinion on this.

------------------------------------------------------------
I have that the sequence of operations in offline - as per the CABLE3 MAIN repo - is: 
- %cansto in from restart (if it exists or set in cable_parameters and/or cable_initialize)
- %oldcansto = %cansto in serial driver, mpimaster and mpiworker.
- cbm does nothing
- %cansto = %oldcansto in define_canopy - which then increments %cansto
- %cansto (and %delwc) into balance_checks, restart and output

ESM1.6 (part 1) - as per the UM7 repo:
- %cansto pack'd from canopy_tile in initialize_canopy
- %oldcansto =%cansto in explicit_river
- %oldcansto=%cansto in cbm_expl
- %cansto=%oldcansto in define_canopy - which then increments %cansto
- %cansto = %oldcansto in cbm_expl

then ESM1.6 (part 2)
- %cansto=%oldcansto in implicit_driver
- %cansto=%oldcansto in define_canopy - which then increments %cansto
- canopy_tile unpack'd from %cansto in implicit_unpack

AM3 (part 1) - as per the ACCESS3-JULES repo:
- %cansto pack'd from canopy_tile in um_update
- %oldcansto = %cansto in um_update
- %cansto=%oldcansto in define_canopy - which then increments %cansto
- %cansto = %oldcansto

AM3 (part 2 and 3):
- pb%ocan = %oldcansto in cable_store_prognostics from implicit_main
- %cansto=%oldcansto in define_canopy - which then increments %cansto
- %cansto = %oldcansto *if ***NOT*** the last call each time step*
- %oldcansto = pb%ocan in cable_restore_prognostics from implicit_main
- canopy_tile unpack'd from %cansto in implicit_unpack

---------------------------------------------
If timing weren't an issue - the two immediate changes I'd make/test are
1. move the %oldcansto = %cansto line from serial driver etc. to offline cbm (this brings the cbm's closer together)
2. delete the %cansto=%oldcansto line in define_canopy

In the longer term:
1. sort out and refactor the incrementing of %cansto in define_canopy
2. make other adjustments across the applications 

This would eventually mean that %oldcansto can be removed from the canopy TYPE entirely and %cansto would be handled in the same way as all the other state variables (e.g. in/out of the PB%) and stop being a special case.

JhanSrbinovsky:
Thanks @har917. I'm pretty sure we actually put in the casnsto=oldcanstyo in define_canopy across apps as it wasnt the case before. Then the app specific treatment differs

PR #518, opened 2025-01-07
Added details around benchcab outputs to the PR template. Fixes #517

### LINKED ISSUE #517: Update PR template to include request for benchcab outputs
In order to expedite code review, the onus should be on the code author to provide test results that the code change still permits CABLE to compile and run without error. The simplest way to do this is to include a benchcab step with bitwise comparison, as it will cover both situations.

A simple update to the PR template should be sufficient to encourage authors to get in the habit of testing with benchcab prior to requesting a review.

bschroeter:
Added details to the PR template RE: benchcab outputs.

## Type of change

- [x] CI/Workflow

## Checklist

----
📚 Documentation preview 📚: https://cable--518.org.readthedocs.build/en/518/

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,8 +20,11 @@ Please delete options that are not relevant.
 
 ## Checklist
 
-- [ ] The new content is accessible and located in the appropriate section.
-- [ ] I have checked that links are valid and point to the intended content.
+- [ ] The new content is accessible and located in the appropriate section
+- [ ] I have checked that links are valid and point to the intended content
 - [ ] I have checked my code/text and corrected any misspellings
+- [ ] I have included relevant test results (i.e. benchcab outputs or path to readable logs) to ensure that the code compiles, runs and is bitwise compatible to the main branch*
+
+**Unless a a change in outputs is expected, in which case add someone with relevant scientific expertise as a reviewer*
 
 Please add a reviewer when ready for review.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -23,8 +23,11 @@ Please delete options that are not relevant.
 - [ ] The new content is accessible and located in the appropriate section
 - [ ] I have checked that links are valid and point to the intended content
 - [ ] I have checked my code/text and corrected any misspellings
-- [ ] I have included relevant test results (i.e. benchcab outputs or path to readable logs) to ensure that the code compiles, runs and is bitwise compatible to the main branch*
 
-**Unless a a change in outputs is expected, in which case add someone with relevant scientific expertise as a reviewer*
+## Testing
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -20,8 +20,11 @@ Please delete options that are not relevant.
...
 ## Checklist
 
-- [ ] The new content is accessible and located in the appropriate section.
-- [ ] I have checked that links are valid and point to the intended content.
+- [ ] The new content is accessible and located in the appropriate section
+- [ ] I have checked that links are valid and point to the intended content
 - [ ] I have checked my code/text and corrected any misspellings
+- [ ] I have included relevant test results (i.e. benchcab outputs or path to readable logs) to ensure that the code compiles, runs and is bitwise compatible to the main branch*
+
+**Unless a a change in outputs is expected, in which case add someone with relevant scientific expertise as a reviewer*
```
ccarouge, line 28: I was thinking of something a bit more beefy.

```suggestion

## Testing

- [ ] Are the changes bitwise-compatible with the main branch? If working on an optional feature, are the results bitwise-compatible when this feature is off? If yes, copy benchcab output showing successful completion of the bitwise compatibility tests or equivalent results below this line.

- [ ] Are the changes non bitwise-compatible with the main branch because of a bug fix or a feature being newly implemented or improved? If yes, add the link to the modelevaluation.org analysis versus the main branch or equivalent results below this line.
```

Especially, I'd like for people to put their test results directly in the PR description. Or is that a bad idea?

ccarouge (changes requested):
Mainly just want to discuss what we want included here.

.github/pull_request_template.md
bschroeter: Beefy is good.

As for the other bit, it depends. In many cases people will submit this, then add a comment for the outputs. Although, putting it upfront will trigger the necessity to actually do the tests before submitting the PR...

Or, in the case of bitwise incompatible result they may need to supply the actual benchcab logfile (or a path to it) on the PR for closer inspection by a reviewer with the scientific know-how to make a determination if the change is material enough to block the PR.

In that case, I think it is still important to leave the comment about assigning an expert, as tech-reviewers won't necessarily have the background to confidently approve a change that break bitwise.

This also raises another consideration, when last I looked the default behaviour of benchcab was simply to return non-zero in the case of bitwise incompatibility and then I was required to do my own `nccmp` to get the actual numbers out of it to investigate. Is this still the default?

ccarouge: The description is editable at any time. I didn't think people would put their results at the start but edit when they have them. I edit my PR descriptions all the time...

For the test results, I don't particularly care about the `nccmp` results if it's not bitwise comparable. In that case, I want to see the me.org analysis results or equivalent. Too much?

bschroeter: If that is the case then I think your text is fine - in truth the ME.org outputs are likely to be needed for a "scientific investigation" of any differences anyway.

ccarouge (approved):
I have to approve my own text :D

ccarouge:
@bschroeter merge?

PR #520, opened 2025-01-08
Compile with GFortran following Fortran 2008 Standard

mcuntz:
# CABLE

## Description

Bug fixes to compile with GFortran > v13 following the Fortran 2008 Standard.

CABLE gets compiled with the option to follow the Fortran 2008 Standard now.
The compiler flag was wrong for GFortran in `CMakeLists.txt`. It was `-std=f08` as for the Intel compiler but should be `-std=f2008`.

There were several non-standard functions used in the code, that do not work anymore with the `-std=f2008` flag:
1. getenv() - replaced by the F2003 function get_environment_variable
3. call exit(5) - replaced by stop 5
4. stop variable - stop cannot be followed by a variable (apparently), put the error code into the message above the stop
5. isnan - this was a GNU extension. Replaced it with code from [here](https://fortran-lang.discourse.group/t/challenge-testing-inf-and-nan-with-gfortran-13-ofast/6481/30)

## Type of change

- [x] Bug fix

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
     )
 endif()
 
-set(CABLE_GNU_Fortran_FLAGS -cpp -ffree-form -ffixed-line-length-132 -std=f08 -D__GFORTRAN__ -D__gFortran__ -D__NETCDF3__)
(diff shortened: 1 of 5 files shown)


Whyborn:
Looks good to me. Thanks for including a link to a thread describing the NaN checker functions, that is very useful. Happy to have this merged.

Whyborn (approved):

Issue #484, opened 2024-11-21
wbliq bug in cbl_soilsnow_main.F90


rml599gh:
wbliq is treated as a local variable in cbl_soilsnow_main.F90 but needs to be written into ssnow%wbliq. Otherwise ssnow%wbliq does not change over time with implications for its use when fwsoil_switch='Haverd2013'.

Originally noticed when trying to enable Haverd2013 with ESM1.5/ESM1.6 (#483) but will also impact offline runs using soilsnow and Haverd2013 but not groundwater.

Testing at selected single sites (with ACCESS met forcing) showed impacts for GPP.

10 day running mean GPP shown for a 1 year simulation. 
Black uses fwsoil_switch='standard'. Red is the result of just changing fwsoil_switch to 'Haverd2013'. Green is the result with code changes in cbl_soilsnow_main.F90 (two instances of `wbliq = ssnow%wb - ssnow%wbice` changed to `ssnow%wbliq = ssnow%wb - ssnow%wbice`)

![Image](https://github.com/user-attachments/assets/02e7481b-ba69-4217-8e31-1353d9696a77)

har917:
There is other development work occurring in the AM3 codebase which intersects with this (seeking to improve energy/water balance and allowing for liquid/frozen water to have different densities).  This bug could be a non-issue once CABLE3-MAIN is sync'd with that AM3 work (and ESM1.6 picks up the latest CABLE3)

rml599gh:
Thanks @har917 for the reminder that the simple fix I'd tested is assuming that rho_water = rho_ice. Sounds like we are best to wait for the more correct, AM3, version.

PR #237, opened 2024-04-18
Comparison of BLAZE_9184 and CABLE-POP_TRENDY branches

ccarouge:
This pull request is to document the differences between the 2 branches while we start on #232 . This can provide a space for discussion on specific differences or simply an overview of the differences.

### core/biogeochem/POP.F90

- A lot of formatting differences hiding important differences
- `rkill` added to `Landscape` type in POP_types module and `ZeroPOP()`
- Changes in `ADJUST_POP_FOR_FIRE()`
- Change is `INTERPOLATE_FIREMORTALITY()`

Updating the formatting in BLAZE_9184 branch would help identifying the scientific differences.

### core/biogeochem/casa_inout.F90

Additional variables added for BLAZE

### core/blaze

Presumably, we would want to keep these files unchanged from BLAZE_9184. Some differences here are linked to differences in other files (e.g. cable_common.F90)

### namelists/cable.nml

We will work from defined configurations outside this repository so differences in this namelist file should not matter.

### build differences

We should try and compile the code on various architectures as people will need to run that code on different machines.

PR #58, opened 2022-11-07
Fix typo in publish action of documentation

ccarouge:
# CABLE

Thank you for submitting a pull request to the CABLE Project.

## Description

The file to configure FORD for GW is named GW_ford.md and the action was looking for gw_ford.md

You can link issues by using a supported keyword in the pull request's description or in a commit message:

## Type of change

Please delete options that are not relevant.

- [x] Bug fix
- [ ] New or updated documentation

## Checklist

- [x] The new content is accessible and located in the appropriate section.
- [x] I have checked that links are valid and point to the intended content.
- [x] I have checked my code/text and corrected any misspellings

Please add a reviewer when ready for review.

PR #527, opened 2025-01-14
494 redux local ice fixes

JhanSrbinovsky:
# CABLE

## Description

Fixes #(issue)

## Type of change

- [ ] Bug fix

## Checklist

## Testing

----
📚 Documentation preview 📚: https://cable--527.org.readthedocs.build/en/527/


JhanSrbinovsky:
@ccarouge I have several issues with testing required (apart from having limited time):

 * The issue here  didnt even show up until global run (is this even possible in benchcab/me.org). Now that that is otherwise "resolved" probably single -site tests are ok
 * Anything short of uploading from gadi is untenable. My failed attempt and request for help (https://forum.access-hive.org.au/t/benchcab-me-org/4056) is still hanging

JhanSrbinovsky:
with end of #494 comments/modifiations

Screenshot 2025-01-16 at 4 07 56 pm

src/science/soilsnow/cbl_remove_trans.F90
```diff
@@ -31,14 +31,14 @@ SUBROUTINE remove_trans(dels, soil, ssnow, canopy, veg)
              ! which can be removed:
              xx = canopy%fevc * dels / CHL * veg%froot(:,k) + diff(:,k-1)   ! kg/m2
              diff(:,k) = MAX( 0.0_r_2, ssnow%wb(:,k) - soil%swilt) &      ! m3/m3
-                  * soil%zse(k)*1000.0
+                  * soil%zse(k)*Cdensity_liq
```
har917, line 34: I'm going to create another issue for this - since the evaluation of `diff` doesn't recognise that only liquid water can be transpired we risk creating a water imbalance by forcing `wb` down so that `wbice/`wb > frozen_limit`.

This issue has always been in the code - so not part of this PR - but it's potentially made worse by allowing ice to have a different density to liquid water.

src/science/soilsnow/cbl_snowAccum.F90
```diff
@@ -47,6 +47,8 @@ SUBROUTINE snow_accum ( dels,  canopy, met, ssnow, soil )
 
              ssnow%tgg(i,1) = ssnow%tgg(i,1) + canopy%precis(i) * CHLF               &
                   / ( REAL( ssnow%gammzz(i,1) ) + Ccswat *canopy%precis(i) )
+             ssnow%dtmlt(i,1) = ssnow%dtmlt(i,1) + canopy%precis(i) * CHLF           &
```
har917, line 50: This is the correct implementation of what has been discussed to date - however, there's something suspicious here under the circumstances of a shallow snow layer, `%tgg < CTRZ` and large `%precis`. At the moment all of the rain is being converted to snow and the temperature increasing past zero - at a later point this will remelt but that probably the incorrect way to do this and truly conserve energy (instead we should limit the amount of rain that can be frozen while keeping the  combined snow <=CTRZ)

Another new issue.

src/science/soilsnow/cbl_soilsnow_init_special.F90
```diff
@@ -64,25 +64,25 @@ SUBROUTINE spec_init_soil_snow(dels, soil, ssnow, canopy, met, bal, veg,heat_cap
          ssnow%wb(:,4) = 0.95 * soil%ssat
          ssnow%wb(:,5) = 0.95 * soil%ssat
          ssnow%wb(:,6) = 0.95 * soil%ssat
-         ssnow%wbice(:,1) = 0.90 * ssnow%wb(:,1)
-         ssnow%wbice(:,2) = 0.90 * ssnow%wb(:,2)
-         ssnow%wbice(:,3) = 0.90 * ssnow%wb(:,3)
-         ssnow%wbice(:,4) = 0.90 * ssnow%wb(:,4)
-         ssnow%wbice(:,5) = 0.90 * ssnow%wb(:,5)
-         ssnow%wbice(:,6) = 0.90 * ssnow%wb(:,6)
+         ssnow%wbice(:,1) = frozen_limit * ssnow%wb(:,1)
```
har917, line 67: how is this SUBROUTINE getting hold of `frozen_limit` and `Cdensity_ice`?  via `cable_common`?

Another comments on this routine for completeness - line 38 should be redundant.

src/science/soilsnow/cbl_soilsnow_main.F90
```diff
@@ -122,7 +122,7 @@ SUBROUTINE soil_snow(dels, soil, ssnow, canopy, met, bal, veg)
     ! snow aging etc...
     CALL snowl_adjust(dels, ssnow, canopy )
 
-   CALL stempv(dels, canopy, ssnow, soil, soil%heat_cap_lower_limit )
+   CALL stempv(dels, canopy, ssnow, soil, REAL(soil%heat_cap_lower_limit) )
```
har917, line 125: The only way I can see that the `REAL()` is needed is if there a change in precision on `soil%heat_cap_lower_limit` across the SUBROUTINE call. That suggests that we need to actually make the edit in `stempv` not here - or the `REAL()` isn't needed.

Ditto for the call to `soilfreeze` below.

src/science/soilsnow/cbl_surfbv.F90
```diff
@@ -7,12 +7,12 @@ MODULE surfbv_mod
 CONTAINS 
 
 SUBROUTINE surfbv (dels, met, ssnow, soil, veg, canopy )
-! subrs
-USE smoisturev_mod,          ONLY: smoisturev
 
+USE smoisturev_mod,               ONLY: smoisturev
+    USE cable_common_module
```
har917, line 12: predantic comment - odd use of white space here

har917 (changes requested):
A couple of new things noted - now as new issues.


Please check i) how `frozen_limit` and `Cdensity_ice` are getting into `spec_soilsnow_init` and ii) why we need a REAL() in one of the SUBROUTINE argument calls - if they're okay then this PR is approved.

src/science/soilsnow/cbl_snowAccum.F90
JhanSrbinovsky: can we just keep freezing quanta of snow WHILE tgg < 0, or does it evaluate the new temperature later on?

Did you add this as an issue?

src/science/soilsnow/cbl_soilsnow_init_special.F90
JhanSrbinovsky: `USE cbl_ssnow_data_mod`

ll 38 - really? It doesnt need to be zeroed?

JhanSrbinovsky:
So @ccarouge  - what are we going to do about testing here?

src/science/soilsnow/cbl_soilsnow_init_special.F90
har917: It does in coupled - but all of this is inside a `IF NOT %cable_runtime_coupled)` condition so this line will never be encountered.

I'm not actually that convinced we need the outer IF condition either as this routine is now only ever encountered via the cbl_serial and mpiworker routines.

src/science/soilsnow/cbl_snowAccum.F90
har917: Added as #531 - I think we keep that separate from this PR as it will require a fair bit more thinking and we really need to focus on syncing the codebases first.

Whyborn:
@JhanSrbinovsky there was a minor mistake in the new gridinfo- apparently xarray's ```DataArray.where```, which I used to apply the ice fixes, silently casts the ```DataArray``` to floating point type, which caused the reading of ```isoil``` to fail. I have updated the file at the same location as before.

JhanSrbinovsky:
> @JhanSrbinovsky there was a minor mistake in the new gridinfo- apparently xarray's `DataArray.where`, which I used to apply the ice fixes, silently casts the `DataArray` to floating point type, which caused the reading of `isoil` to fail. I have updated the file at the same location as before.

Odd - my run didnt fail? Anyway, how do we go about replacing the grid_info in CABLE repo OR are we even concerned with doing that?

Whyborn:
That's very strange, my test run failed with  ```floating invalid``` in ```NF90_GET_VAR``` (and apparently it casted all the floating point parameters to 64bit floats, rather than 32bit so it broke all of them as well). I haven't gone anywhere near the gridinfo in the CABLE repository, I don't think that's the place to store ancillaries personally.

JhanSrbinovsky:
> That's very strange, my test run failed with `floating invalid` in `NF90_GET_VAR` (and apparently it casted all the floating point parameters to 64bit floats, rather than 32bit so it broke all of them as well). 

This is a worry. Were you running in debug mode?

>I haven't gone anywhere near the gridinfo in the CABLE repository, I don't think that's the place to store ancillaries personally.

Neither do I - storing binaries in there is contrary to the whole point really.

JhanSrbinovsky:
tested version4 against version3 (gridinfo files). Identical results for me.

Whyborn:
Yes, this was running in debug mode. I have since corrected the problem and managed to get a short run, with only 6 grid points and a single year, going with the ACCESS-ESM1.5 gridinfo and meteorology. Now extending it to a longer global run.

JhanSrbinovsky:
> Yes, this was running in debug mode. 
Makes total sense then, I'm not worried now :)

>Now extending it to a longer global run.

above comment (vn3 vs vn4) was for global GSWP2, so should work for others.

src/science/soilsnow/cbl_soilsnow_init_special.F90
JhanSrbinovsky: ok - so I've looked at the whole file and not just the line 38.

I agree can remove the outer loop. If we want  to zero dgdtg on first step of each run we can add it to  um_init_*., without the ktau=1 thing. it looks like a per timestep var anyway so this should be fine just to give it a finite value.

ccarouge:
@JhanSrbinovsky for testing:
- show the flux site runs are identical (excerpt from benchcab log file) to show it doesn't change what it shouldn't change.
- for the spatial runs, in the discussion I see [these results](https://github.com/CABLE-LSM/CABLE/pull/527#issuecomment-2594552871), it would be good for @har917 to give a general acceptance of the results (the differences shown look plausible and the testing is enough for the change). @JhanSrbinovsky have you said anywhere what the simulations are? It doesn't look like you did. Updating the PR description to add some information on the simulations you have done would be best. And in general, having any test results within the description (that can be updated anytime) is preferred to having them in separate comments. 

And finally, @har917 needs to review again because he is still noted as requesting changes (or @JhanSrbinovsky you haven't put in all requested changes...).

For upload to me.org, Ben has been working on it and should get you a reply soon. me.org analysis is not needed here as we can't yet run ILAMB from me.org.

src/science/soilsnow/cbl_soilsnow_init_special.F90
JhanSrbinovsky: @ccarouge - I gather this was the outstanding change request you were referring to. Just waiting on Ians response

har917: @JhanSrbinovsky @ccarouge My point here is/was really that `cbl_soilsnow_init_special.F90` is only ever encountered via the serial driver and mpi_worker routines (and possibly even only on the first call) and so that implies that the some of logical inside the routine is irrelevant (because it's trapped by logic at a higher level).

As such:
- the IF condition at line 35 is irrelevant  (you never have `cable_user@cable_runtime_coupled = .TRUE.` in this subroutine).
- the IF conditions at line 37 and 81 are irrelevant (we shouldn't be encountering the subroutine twice)
- the IF condition at line 38 is irrelevant - this is an offline-only subroutine and anyway `%dgdtg` is always set in `define_canopy` before first use.


It could/would be sensible to include an initialisation of `%dgdtg` (and `%dfh_dtg`, `%dfe_dtg` and `%dfn_dtg`) `cbl_um_init_soilsnow` but that's not strictly needed.

It would be sensible to sort this out - if nothing else because it removes things like the need for the SAVE condition on ktau (i.e. applying the coding standards to these refactored parts of the code).

 ... BUT all of this is really PR creep.  Leaving these lines in won't do any damage (as far as I can tell).

JhanSrbinovsky: These have been here since ACCESS-1.3 days and we just keep adding switches. The canopy% fields (like Claire has been looking at int JAC version of climate% decs) are all zeroed at declaration,allocation as was required there. If not zeroing for the moment is not causing a problem then I suggest we wait until we pick this up and be sure to CALL the zeroing SUBR.  AND remove this code. Note aswell the coupled_runtime flag has effectively been re-purposed and should at least be renamed to reflect that it does not indicate whether it is in a coupled run or not. I'll make an issue from here

JhanSrbinovsky: @har917 The only use of ktau is at the bottom, outside of the runtime_coupled loop so it gets executed for everyone, and in the UM - (potentially) on every restart. BUT init_spec is never called in the UM anyway. 

 * Does offline ever run in chunks? 
 
IF no, we can just use ktau_gl then and get rid of ktau completely.

Also, it strictly should be using **ssnow%**xx=heat_cap_lower_limit(:,1)

JhanSrbinovsky: unless of course that is what is being passed but I dont remember editing the offline drivers recently

JhanSrbinovsky:
>show the flux site runs are identical (excerpt from benchcab log file) to show it doesn't change what it shouldn't change.

I reran the single site testing. It isnt bitwise identical. I never really expected it to be except maybe where the ice fixes (including ice density where there is any frozen liquid in soil) dont apply. Looking at the headline subR, soilsnow_main there is a kind=r_2  which is enough to destroy bitwise compatabiliy AND then there is correction of a bug in which wbliq was only local/.This made me wonder how they were ever identical. Returning to the early days - [they never were identical](https://github.com/CABLE-LSM/CABLE/pull/497#issuecomment-2505103819).


>for the spatial runs, in the discussion I see https://github.com/CABLE-LSM/CABLE/pull/527#issuecomment-2594552871, it 
would be good for @har917 to give a general acceptance of the results (the differences shown look plausible and the testing is enough for the change). 


>@JhanSrbinovsky have you said anywhere what the simulations are? It doesn't look like you did. 
Updating the PR description to add some information on the simulations you have done would be best. 

I will look at updating this

src/science/soilsnow/cbl_soilsnow_init_special.F90
har917: > the only use of ktau is at the bottom, outside of the runtime_coupled loop so it gets executed for everyone, and in the UM.

Not as far as I can tell:  at the moment
- `spec_init_soilsnow` is called from `cable_serial` and `mpi_worker` before the first time step and outside the timestepping loop.
- `spec_init_soilsnow` is not encountered in AM3

That leaves ESM1.6 - going from the ESM1.5 `cable_um_init` routine it appears that the intent is to call `spec_init_soilsnow` from within an `IF (first_call)` construct - if so then we can remove all mentions of ktau from this routine (and probably the if coupled conditions as well).

In rereading I also note that `%gamzz` set at line 75 is different to that set at 81 - which seems odd.

har917:
@JhanSrbinovsky @ccarouge On the question of impact - it's a bit hard to give a definitive answer (partly because this has been a moving target).

Referring to map above - I'm assuming this is the net impact on one of the temperatures (which one?) after one day.  If so this is reasonable since
- we expect a 'large' impact to come from the forced melting and discharge of frozen soils on the first time step
- forced melting induces a cooling of the soil - so MAIN - NEW will be mainly positive.
- temperature increment is mainly in the 1-2K range (physically plausible) - only regions outside of this range are a on couple of Artic islands (possible confused with the changes to the soil/veg conditions) and in Alaska (likely vert saturated soils.

I'd like a longer run to see if the temperature differences settle over time and/or we get something more systematic emerging - but no cause for concern so far.

src/science/soilsnow/cbl_soilsnow_init_special.F90
JhanSrbinovsky: its weird - if there is NOT enough snow for the 3-layer scheme to kick in then gammzz is mediated  by snowd, but if there is enough snow then its the same base calc as lline 75 version

JhanSrbinovsky:
@har917  
> Referring to map above - I'm assuming this is the net impact on one of the temperatures (which one?) after one day. If so this is reasonable since

SoilTemp (top layer)

> I'd like a longer run to see if the temperature differences settle over time and/or we get something more systematic emerging - but no cause for concern so far.

I might be able to resurrect the (I think) 5 yearGSWP2 run. We already have 1 year which I can show. Otherwise these same snow fixes have run longer term in AM3, I just cant think of which run

src/science/soilsnow/cbl_soilsnow_init_special.F90
har917: Actually that would makes sense - in a somewhat odd way.

If there's enough snow for three layers then the soil temperature equations get appended by 3 extras layers - in `stempv` and `surfbv`.  The effect of the snow of `%gammz` gets wrapped up in the equations for those layers so *shouldn't* be counted in `%gammz(:,1)` here.

In contrast if there isn't enough snow for the 3 layers then we may need to include the snow effect on `%gammz` since the uppermost soil layer and the snow are effectively modelled as one layer.

However the two lines in `spec_init` also differ in whether or not `IF( .NOT.cable_user%cable_runtime_coupled )` so it's more confused.

-----------------
With the current codebase (for offline) `%gammz` is
- set in `spec_init_soilsnow`
- reset in `soilsnow_main` in a `ktau<=1` condition (so first call only)
- used first in `cbl_snow_accum`

and the initialisations are different.

In AM3 `%gammz` is
- set in `cbl_um_init_soilsnow`
- reset in `soilsnow_main` in a `ktau<=1` condition (so first call only)
- used first in `cbl_snow_accum`

and the initialisations are the same.

------------------
I also note that in `stempv`, `%gammz` doesn't include the snow effect (line 184) - this is picked up by `sgamm` at line 169.

------------------
Net result - `%gammz` has to have a proper value by the time we get to `cbl_snowMelt`.  This means that it has to be set in `spec_init_soilsnow` and `cbl_um_init_soilsnow` *or* in `soilsnow_main` (both doesn't hurt but not needed).

I'm not convinced about the inclusion of the snow effect in the second evaluation in `spec_init` to be honest.

------------------
... and I've just realised that `%gammz` is one of these internal-to-CABLE variables that break restart reproducibility - aargh!

JhanSrbinovsky: so do you want to add this to the restart ?

JhanSrbinovsky: I was a little perplexed that this was called offline. I went to check and was even more concerned as there is no condition around it. But then I went back to the subr spec_imit, and most of it is blocked by  the IF(runtime_coupled) so it really only initialises gammz. For now I am happy to leave this, but dont you @har917  think we should split this into init_gammz and init_spec (which is only called for special cases which should probably be relegated to the dustbin of history)?

JhanSrbinovsky: it isnt called in ESM1.6 but thankfully we have initialised gammz in cable_um_init_soilsnow

JhanSrbinovsky: @har917 see 3178b8b2..31a233fb

har917: > @har917 so do you want to add this to the restart ?

I think at this point I think we leave it for now (since `%gammzz` is being initialised in `cbl_um_init_soilsnow`) - and create an issue to remind ourselves to return to things later.

har917: see conversation in [AM3 #237](https://github.com/CABLE-LSM/ACCESS3-JULES/issues/237) - this question of whether to inlcude `@gammzz` in the AM3 restart has been tagged as needing resolving there.

JhanSrbinovsky: ok - I'll do that post the fes_cor going in

JhanSrbinovsky:
@har917 all the AM3 and ESM16 runs include these ice fixes - can you look at long term impacts there I still haven't looked into resurrecting the 5yr GSWP2 (that I remember exiting, I might even be making it up)

har917:
> @har917 all the AM3 and ESM16 runs include these ice fixes - can you look at long term impacts there I still haven't looked into resurrecting the 5yr GSWP2 (that I remember exiting, I might even be making it up)

@JhanSrbinovsky I understood that the AM3 runs included these ice fixes, but that ESM1.6 (so far) doesn't (since that builds from CABLE3:MAIN) - or have we bypassed that requirement?

JhanSrbinovsky:
Generally no, but we wanted to get a bunch of things before Xmas. So CABLE:main is the remaining version that thats ice with the density of liquid water etc.

har917:
@JhanSrbinovsky Sorry - I don't quite follow your most recent comment.  

I thought this issue was the attempt to sync the AM3 - CABLE3 and hence ESM1.6 science code (around the ice density and associated fixes).  Are you saying that ESM1.6 (presumably case-2d) ended up ahead of CABLE3:MAIN?

If yes, then what is the status of the changes to `cbm` (specifically whether the correction terms `%ga_cor` and `%fns_cor` are in) in that ESM1.6 run?  If they are also in then we need to (re)review energy conservation in the ESM.

I had understood that those changes weren't in the case-2d codebase - and so were likely the explanation for the poor energy balance, with the rivers changes the explanation for the water imbalance [ESM1.6 #58](https://github.com/ACCESS-NRI/UM7/pull/58)

JhanSrbinovsky:
ice. fixes - yes https://github.com/ACCESS-NRI/UM7/pull/52

cbm - not that I remember. BLAME doesnt seem to indicate so either https://github.com/ACCESS-NRI/UM7/blame/dev-access-esm1.6/umbase_hg3/src/atmosphere/CABLE/src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_cbm.F90

JhanSrbinovsky:
for e.g. PI-case2d, cable.nml = 

  snmin = 0.11        ! depth at which snow switches to 3-layer
!
!
!! cable_user flags
!
  cable_user%GS_SWITCH = 'medlyn' ! alternative is 'leuning' - used for ESM1.5
  cable_user%CABLE_RUNTIME_COUPLED = .TRUE. ! controls whether snow initialised if
                                            ! ktau_gl=1
                                            ! must be set true for coupled run
                                            ! use true for atmosphere only runs if
                                            ! want to take snow from start dump
                                            ! default is .false.
  cable_user%FWSOIL_SWITCH = 'Haverd2013' ! Controls root water uptake function
                                        ! choices are:
                                        ! 'standard' - used for ESM1.5
                                        ! 'non-linear extrapolation'
                                        ! 'Lai and Ktaul 2000' NB TYPO IS IN CODE (should be Katul)
                                        ! 'Haverd2013'  - likely choice for ESM1.6
  cable_user%DIAG_SOIL_RESP = 'ON '  ! Controls soil respiration scheme when CASA-CNP not used
                                     ! 'ON' uses scheme tuned with parameter veg%vegcf
                                     ! 'OFF' uses scheme tuned with parameter veg%rs20
  cable_user%LEAF_RESPIRATION = 'OFF' ! Controls what is output into the GPP stash variable
                                      ! 'OFF' is true GPP (photosynthesis + leaf respiration)
                                      ! 'ON' is only photosynthesis, canopy%fpn
  cable_user%RUN_DIAG_LEVEL= 'BASIC'  ! Controls output from CABLE to standard out
                                      ! choices are:
                                      ! 'BASIC'
                                      ! 'NONE'
                                      ! 'zero' (in cable_explicit_driver - what does this do?)
  cable_user%ssnow_POTEV = ''    ! Controls method used for soil evaporation
                                 ! Choices are:
                                 ! 'P-M' Penman_Monteith, otherwise defaults to
                                 ! humidity deficit method
  cable_user%access13roots = .TRUE. ! TRUE uses prescribed froot values
                                    ! FALSE uses 'rootbeta' parameterisation (as ACCESS-CM2)
  cable_user%SOIL_THERMAL_FIX = .TRUE. ! TRUE used from CM2 onwards
                                       ! FALSE as ESM1.5 (didn't exist then)
!!--------------Test Switches--------------
  cable_user%l_new_roughness_soil  = .FALSE.   ! for new soil roughness, E.Kowalczyk Mar14
  cable_user%l_new_runoff_speed    = .FALSE.   ! for new increase in runoff speed, E.Kowalczyk Mar14
  cable_user%l_new_reduce_soilevp  = .FALSE.   ! to reduce soil evaporation, E.Kowalczyk Mar14

har917:
okay - so we need to add the lines

 ```
cable_user%L_REV_CORR  = .TRUE.  !apply fix to correctly diagnose energy balance in ESM1.6 runs
                                  !FALSE in ESM!.5 (as didn't exist) 
```

to that namelist.

**Comment by JhanSrbinovsky** [comment]
@har917 - DONE in ready to go PI-case2e. What shall we do with this PR though?

**Comment by har917** [comment]
> What shall we do with this PR though?

I think that benchcab and spatial runs tests with the 'final, final' code set need to be done and included as part of this conversation - and then reviews from @har917 and @ccarouge ticked off, and then it should be okay to go in.  

(but since this is going into MAIN @ccarouge has the key voice)

File: src/science/soilsnow/cbl_soilsnow_init_special.F90 (hunk 3 above)
Comment (har917) [reply src/science/soilsnow/cbl_soilsnow_init_special.F90:64]: I still think that line 38 is redundant - but it's not harming anything.

**Review by har917 — APPROVED** [review]
I'm happy for this to go in from a code perspective - `spec_init_soilsnow` will need to be reviewed at some point once we have a bit more clarity on a) `%gammzz` in the restart(s) and b) which applications that code is being used in.

We still need final testing to be included in this PR/conversation.

**Comment by JhanSrbinovsky** [comment]
To summarise the testing situation:

 * At the top is the 1st day of a 1 year run. 
 * 5 yers I haven't been able to do yet
 * IDK whether benchcab supports spatial runs
 * Even single-sites w no ice are NOT identical, requiring further analysis
 * uploading to me.org is at present not possible
 * Unfotunarely rsynching to me.org  isnt possible. It has to upload in one go. For 5 site analysis this took me three days unsuccessfully from home and then two from the office and I cant really get to the office until next Friday. An outside chance of Wednesday. 

> Please check i) how `frozen_limit` and `Cdensity_ice` are getting into `spec_soilsnow_init` and ii) why we need a REAL() in one of the SUBROUTINE argument calls - if they're okay then this PR is approved.

via cbl_soilsnow_data_mod

**Comment by ccarouge** [comment]
@JhanSrbinovsky If you tell me the path of the benchcab outputs, I can upload them for you. It is very fast from ANU. Although I would think we could get a benchcab version that would work in conda by the end of the day. I'll put an update here when that's ready.

**Comment by JhanSrbinovsky** [comment]
@ccarouge - that'd be great

/scratch/p66/jxs599/bench_example/runs/fluxsite/outputs

**Comment by ccarouge** [comment]
I've run the full 42 sites with benchcab for the branch and got [the evaluation on me.org](https://modelevaluation.org/analyses/anywhere/nFcjg4qqHGPkB9sqE/s6k22L3WajmiS9uGv/CDfKPXWFmETT32ZpX/all).

Interestingly, the cases with `FWSOIL_SWITCH=Haverd2013` show the largest differences. The `standard` option shows almost no differences (nothing to worry about). It might be a lead to find the cause of the non-reproducibility.

**Comment by har917** [comment]
> It's possible to have differences at single sites with snow.

We should be getting differences with frozen soils (as well as with melting snow and freezing rain).

I'm not overly sure on the `fwsoil_switch` question:  this science will effectively reduce the amount of water (in all forms) that sits in the soil column.  This will then reduce the water available for ET later in the simulation - the `fwsoil_switch` is far better at restricting transpiration in the case of limited water availability, so there could be some complicated interactions going on.

Have you been able to look at the Ebal and Wbal metrics for (some of) these runs - should be near the end of the standard output.  It would be good to check that these have improved/not overly degraded.

**Comment by har917** [comment]
> Have you been able to look at the Ebal and Wbal metrics for (some of) these runs 

Not sure if this is relevant - but looking at my current snow_aging runs - the configurations with 'Haverd2013' are (apparently) conserving much (much) worse at some sites than those with `standard' - with worse meaning several orders of magnitude.

That level of difference/sensitivity will inevitably feed into the me.org metrics - but does raise the question as to what's going on?

**Comment by ccarouge** [comment]
Here is what I see for the water balance for a very few sites where I'm pretty certain there is snow some of the time:

| station | main + Haverd2013 | main + standard | 494 + Haverd2013 | 494 + standard |
|--------|--------------------|------------------|--------------------|-----------------|
| FI-Hyy | -1.1594189E-02 | -1.0896737E-02 |  -8.411185 | -8.410690 |
| US-UMB |  -607.1661  | -2.1411697E-03 |  -3.697309    |  -4.025065    |
| CA-Qfo |  -1.0523711E-02 | -1.0954248E-02 |  -2.875305   |  -2.866241    |

US-UMB is weird for Haverd2013 in main. But for the other sites, there is little difference between either scheme for the water balance. And the ice fixes seem to degrade the water balance in offline.

**Comment by JhanSrbinovsky** [comment]
whoa - what's with:

US-UMB | -607.1661
-- | --

**Comment by har917** [comment]
okay - we have a problem - lines 502-514 in `cable_checks` are assuming that density of water is always 1000.0.  These lines feed into `delwb` and the `bal%wbal` variable.

Basically for any site where any soil layer freezes this will be incorrect.

I suspect that most of the wbal issue at US-UMB site will be in the first time step - reflecting a quirk with the initialisation.

**Comment by har917** [comment]
a quick check on CABLE3 MAIN raises a few more cases of 1000.0 or 1000. that aren't obviously a m->mm type unit conversion.  These are in `cable_psm`, `cable_output` (for `ssnow%wtd`) and `cable_parameters`.  I suspect that these are units conversions but aren't commented as such.

**Comment by har917** [comment]
> okay - we have a problem - lines 502-514 in `cable_checks`

I think the easiest way around this is to replace in local re-evaluation of the total water (liquid++frozen). and hence change in water content, with that evaluated in `%wbtot`.  This is available from `soil_snow` and `soil_snow_gw` - SLI already has it's own `wbal` version which correctly incorporates frozen/liquid densities.

In effect the `mass_balance` routine changes (variable definition):
```
! Local variables
REAL(r_2), DIMENSION(mp,2),POINTER, SAVE :: bwb         ! LOCAL total (liquid+ice) soil moisture
REAL(r_2), DIMENSION(mp)                  :: delwb                 ! change in soil moisture
```
then (first time step and main calculation)
```
IF(ktau==1) THEN
       ! initial value of soil moisture content - liquid/ice densities included
       bwb(:,1)=ssnow%wbtot
       bwb(:,2)=ssnow%wbtot
       delwb(:) = 0.0

ELSE
       ! Calculate change in soil moisture b/w timesteps:
       IF(MOD(REAL(ktau),2.0)==1.0) THEN            ! if odd timestep
             bwb(:,1)=ssnow%wbtot
             delwb(:) = bwb(:,1) - bwb(:,2)

       ELSE IF(MOD(REAL(ktau),2.0)==0.0) THEN    ! if even timestep
             bwb(:,2)=ssnow%wbtot
             delwb(:) = bwb(:,2) - bwb(:,1)
          
       END IF
END IF
```

There is an interesting quirk around how this works with the lakes (where we keep adding water anyway)

**Comment by har917** [comment]
@ccarouge Looking at one of the problematic cases with the `Haverd2023` fwsoil switch indicates that this is not varying (at least in the output).  This seems to be common across all sites with this switch at true.  Different cases seem to land on different values - but always constant.  Depending on what value it takes we could easily end up in a position where different parts of the model want to transpire/evaporate/extract different amounts of water.

Assuming that the output is working - this seems to indicate that `cbl_dryLeaf` is not calling `getrex_1d` properly - given that's where the value gets set with that switch - or that there something else gone awry around the sequence to find `fwsoil`

Looking at the CABLE-POP_TRENDY branch and MAIN in a bit more detail indicates that a possible problem is that `canopy%fwsoil` is reinitialised (to 1.0) each entry into `define_canopy` from MAIN but not with CABLE-POP_TRENDY.  Since `cbl_dryLeaf` with `Haverd2013` uses the value of `canopy%fwsoil` to set the local value `fwsoil` at first then this difference could be introducing a consequential difference.

I don't know when that difference was implemented - likely around the refactorisation of the code.

**Comment by ccarouge** [comment]
> > okay - we have a problem - lines 502-514 in `cable_checks`
> 
> I think the easiest way around this is to replace the local re-evaluation of the total water (liquid++frozen). and hence change in water content, with that evaluated in `ssnow%wbtot`. This is available from `soil_snow` and `soil_snow_gw`. SLI already has it's own `wbal` version in `cable_checks` which correctly incorporates frozen/liquid densities.

If we are going to change how the `wbal` stuff is calculated, can we rewrite it in this way:

```fortran
    REAL(r_2), DIMENSION(:),POINTER, SAVE :: owb         ! volumetric soil moisture
    REAL(r_2), DIMENSION(mp)                  :: delwb       ! change in soilmoisture

    IF(ktau==1) THEN
       ALLOCATE( owb(mp) )
       ! initial value of soil moisture
       owb=ssnow%wbtot
    END IF
           
    ! Calculate change in soil moisture b/w timesteps:
    delwb = ssnow%wbtot - owb
    
    ! Update old soil moisture to this timestep value.
    owb = ssnow%wbtot
```

It does the same thing and is faster to read (well, at least for me).

We could add this to this PR if we want to.

For the `fwsoil` value, it is worth looking into but let's not do that within this PR.

har917:
> We could add this to this PR if we want to.

I think we have to do this in this PR - it's a related part of the whole liquid/ice densities aren't equal topic and could create confusion later on if we don't do it now.

Happy to go with your preferred implementation.

har917:
opened new issue about the fwsoil quirk with 'Haverd2013' at #546

ccarouge:
I'm changing the water balance calculation then.

ccarouge:
New water balances after the fix to the densities:

| station | main + Haverd2013 | main + standard | 494 + Haverd2013 | 494 + standard |
|--------|--------------------|------------------|--------------------|-----------------|
| FI-Hyy | -1.1594189E-02 | -1.0896737E-02 |  -6.402957 | -6.238629 |
| US-UMB |  -607.1661  | -2.1411697E-03 |  -1.596815    |  -42.26907    |
| CA-Qfo |  -1.0523711E-02 | -1.0954248E-02 |  -1.574845   |  -0.8182139   | 

The fix seems to improve the cumulative water balances for this branch in general, although they are still worse than the main branch. Except for something weird at US-UMB for `fwsoil_switch='standard'` but considering US-UMB results on the main branch for Haverd2013, it looks like something site-specific.

har917:
@ccarouge @JhanSrbinovsky, @rml599gh has just pointed me to #484 

Just to note that the current branch includes the changes necessary to capture that issue.

har917:
@ccarouge @JhanSrbinovsky @rml599gh I now understand why we're getting somewhat inconsistent impacts from the 'Haverd2013' parameterisation.

Below are 4 plots showing time series (blue = 'standard' R0_S2, orange = 'Haverd2013', R0_S0) for fwsoil, latent heat Qle (basically transpiration), wbal (the integrated water imbalance) and soil moisture in each of the 6 layers - from the benchcab runs for UMB from #545.

As noted 'Haverd2013' is not varying the fwsoil, this allows unrestricted transpiration, which is driving the soil moisture below the standard case (and even below the wilting point swilt).  However it is only if and when this drives the water content below one or other of the hard limits (0 soil moisture in soil layer 4) that this manifests as a water imbalance.  If you don't hit such a case then 'Haverd2013' is conserving well, albeit the dynamics are incorrect - hence the inconsistency in results across cases.

All this is pointing towards the need to understand what's happening with fwsoil with Haverd et al. as the route to explain this issue.

fwsoil_US-UMB
QLE_US-UMB
Wbal_US-UMB
soilmoist_US-UMB
* dotted lines mark the wilting point and field capacity for this site.

JhanSrbinovsky:
Sorry - In and out today - what's the 4000-6000?

JhanSrbinovsky:
fwsoil stuck at .95 is suspicious isnt it ?

har917:
> what's the 4000-6000?

time steps in the output - i.e. 0 on the x-axis relates to time step 4000 in the output.  Sorry could have been cleaner.

har917:
@JhanSrbinovsky @ccarouge @rml599gh - see #546 

okay so it appears that @rml599gh was at least partially correct.  I took MAIN and *only* edited `cbl_soil_snow_main` to make the LOCAL `wbliq` into `ssnow%wbliq` - this allows `%fwsoil` to vary through time.   For the default Tumbarumba test case the water imbalance drops from -9.485mm to -6.07e-4 mm when fwsoil_switch = 'Haverd2013'

For reference, the 'standard' switch on this case gives a water imbalance of -4.42E-4mm .

However this doesn't explain what's happening with the #494 branch since I think that branch incorporates those edits to `cbl_soil_snow_main`.

src/offline/cable_checks.F90
```diff
@@ -474,48 +474,46 @@ END SUBROUTINE constant_check_range
...
-                  - (bwb(k,:,2)))*soil%zse)*1000.0
-          END DO
-       ELSE IF(MOD(REAL(ktau),2.0)==0.0) THEN    ! if even timestep
-          bwb(:,:,2)=ssnow%wb
-          DO k=1,mp           !  current smoist - prev tstep smoist
-             delwb(k) = SUM((bwb(k,:,2)                                         &
-                  - (bwb(k,:,1)))*soil%zse)*1000.0
-          END DO
-       END IF
+       owb=ssnow%wbtot
```
har917, line 509: @ccarouge sorry this is incorrect.  This updates `owb` before using it to evaluate `delwb` - i.e. `delwb` will always = 0

ccarouge: It's only on ktau == 1. so first time step initialisation. So at first time step we will have 0 but then it will work fine. Same behaviour as before.

ccarouge: If it helps on the review page on GitHub, it's possible to display the differences unified or split. Use the gear icon to choose (let me know if not clear on the screenshot).

Screenshot 2025-02-14 at 4 01 21 pm

JhanSrbinovsky:
> However this doesn't explain what's happening with the #494 branch since I think that branch incorporates those edits to `cbl_soil_snow_main`.


It does,. What's the wbal for Tumbarumba with #494 @ccarouge ?

JhanSrbinovsky:
also what's the idea behind the wbal tracking "main" until timestep~5000 (in the above plots)?

har917:
@ccarouge I think we need to look at the timeseries of `wbal` here - it could be possible that all of this imbalance is occurring in the first few time steps (which could reflect that the initial conditions may not be consistent with the new science).

Are these data still on gadi somewhere?

JhanSrbinovsky:
I just re-read it, so the soil moisture just t takes that long to dry out I guess

har917:
> also what's the idea behind the wbal tracking "main" until timestep~5000 (in the above plots)?

In effect the `Haverd2013` code is working at conserving water until the overly high value of `fwsoil` allows so much transpiration to occur that it drives the soil water content (layer 4) negative.  At this point (timestep ~5000) the code forces soil moisture back to (near) zero values (while not decreasing transpiration) - so water conservation is broken.  

Once there's some rain and the soil moisture content can naturally climb above zero again (time step ~5250) it reverts to conserving nicely at the wbal plot levels off. 

With a sensibly behaving `fwsoil` (as provided by the changes to `cbl_soil_snow_main`) CABLE shuts off transpiration before this occurs and `wbal` is much better behaved (for this run).

har917:
@JhanSrbinovsky In this #494 branch - is `%wbliq` initialised before the first call to `cable_canopy` and `dryLeaf`?  If not we're passing 'garbage' into the first calls to both `dryLeaf` and `fwsoil_calc_std` if GW active.

This was in `cable_parameters` but that file has been removed.  I suspect that we need a line in `cbl_soilsnow_init_special`

JhanSrbinovsky:
ok - thanks, that was my next question - so the recovery had to wait for the rain.

wbliq init in AM3 happens. IDK about offline. It used to be there. wbliq was messed up, but the initialization was still there. Did someone remove it?

har917:
@ccarouge @JhanSrbinovsky This is worse than it looks from the headline water balance numbers.

Looking at the time series of `wbal` from data on modelevaluation.org for #527 - it would appear that the end numbers picked out by Claire for these sites are actually end of year numbers related to something much worse.  There is a significant annual cycle to the water imbalance at these 3 sites (both fwsoil switches).

*BUT* I'm not sure which set of numbers I've picked up from me.org - the time stamp indicates 12/3/2025 whereas I thought we'd (well @ccarouge) redone the analysis on the 14th.  So it could be that this comment is out of date given the changes to `cable_checks` .

Also - my comment around `cable_parameters` was misleading - I got things confused.  `cable_parameters` is still in the branch/repo but `cable_params_mod` has been deleted.

ccarouge:
@har917 @JhanSrbinovsky Trying to look into this more systematically, I've created very rough plots for all the sites for:

- main
- main at the start of this branch
- the head of this branch
- this branch without the last fix for the water balance

The plots have timeseries of Wbal, Fwsoil and SoilMoist for all soil levels. There is a file for Haverd2013: 
[stations_S0.pdf](https://github.com/user-attachments/files/18950850/stations_S0.pdf)

And a file for standard `fwsoil_switch` (only difference between the runs): 
[stations_S2.pdf](https://github.com/user-attachments/files/18950875/stations_S2.pdf)

What I see from these is:

- this branch is fixing a problem with fwsoil_switch=Haverd2013. `fwsoil` used to be constant but it isn't anymore.
- the last fix on the Wbal calculation is essential to the water balance

So it looks like we have enough there to document the changes. I'm not too sure why the fwsoil_switch=Haverd2013 seems to work better with this branch changes but I'm fine with that. So we are good to go for me.

ccarouge (approved):
Good to go

PR #544, opened 2025-02-12
Add serial driver gswp3 support and fix serial-MPI non-reproducibility

### LINKED ISSUE #388: Offline serial driver does not support `cable_user%MetType = 'gswp3'`
Running CABLE offline in serial (`main` branch - commit 95b9b5e915581b0d0ed0ed407573cb448770c7b4) using the [crujra_accessN96_1h](https://github.com/CABLE-LSM/crujra_accessN96_1h) results in the following error:

```
 Cannot process command line definition of namelist fie
 using cable.nml as the namelsit
 THE NAME LIST IS cable.nml
 Use spatially-specific soil properties;          192         112
           0
           0
 When choosing spatially-specific soil properties,
 snow-free albedo is also overwritten by this data set.
 Reading CABLE_PFTPARM namelist...
 Reading CABLE_SOILPARM namelist...
 Total number of patches (countPatch):            0
  Could neither find restart file INPUT/restart.nc
  nor INPUT/restart.nc/STANDARD_1901_cable_rst.nc
  Pre-loaded default initialisations are used.
 OVER WRITING CSS_VEC, RHOSOIL_VEC, CNSD_VEC read from gw_type input
 Forcing values consistant with soil_thermal_fix = true
 Reading CABLE_PFTPARM namelist...
 Reading CABLE_SOILPARM namelist...
 Error defining y dimension in output file. (SUBROUTINE open_output_file)
 NetCDF: NC_UNLIMITED size already in use
```

The above configuration using a debug build crashes with the following error:

```
 Cannot process command line definition of namelist fie
 using cable.nml as the namelsit
 THE NAME LIST IS cable.nml
forrtl: error (72): floating overflow
Image              PC                Routine            Line        Source
cable              0000000000B482C4  Unknown               Unknown  Unknown
libpthread-2.28.s  00007F806F3ACD20  Unknown               Unknown  Unknown
cable              000000000041502C  MAIN__                    604  cable_driver.F90
cable              000000000040CBA2  Unknown               Unknown  Unknown
libc-2.28.so       00007F806EFFE7E5  __libc_start_main     Unknown  Unknown
cable              000000000040CAAE  Unknown               Unknown  Unknown
Aborted
```

The debug build crashes on [this line in cable_driver.F90](https://github.com/CABLE-LSM/CABLE/blob/95b9b5e915581b0d0ed0ed407573cb448770c7b4/src/offline/cable_driver.F90#L604). It looks like this is due to 
an uninitialised `dels` variable (which is caused by the `cable_user%MetType` namelist option being set to `gswp3` which is not supported in the serial driver).

### Steps to reproduce (Gadi)

1. Clone the [crujra_accessN96_1h](https://github.com/CABLE-LSM/crujra_accessN96_1h) repository.
2. Set `exe` in `config.yaml` to point to the absolute path of the serial executable.
3. Setup the configuration with `payu setup`.
```
# via hh5 modules
module load conda/analysis3-24.07
payu setup
```
4. Change directory into `work` and run the symlinked CABLE executable.

SeanBryan51:
Currently the offline serial driver does not support the namelist option `cable_user%MetType = 'gswp3'` - the option is only supported in the MPI drivers. This PR adds support for the namelist option to the serial driver so we can use these configurations for testing the MPI implementation with parallel-IO (which will be based off the serial driver). This PR also fixes several bugs found in the MPI drivers such that we have bitwise reproducibility between serial and MPI for the [gswp3 configuration](https://github.com/SeanBryan51/crujra_accessN96_1h) for outputs and restarts (run for 1 year on a reduced domain).

Fixes #388

## Type of change

- [x] Bug fix

## Checklist

## Testing

- [x] Are the changes non bitwise-compatible with the main branch because of a bug fix or a feature being newly implemented or improved? If yes, add the link to the modelevaluation.org analysis versus the main branch or equivalent results below this line.
  - Changes are bitwise compatible with main for site runs but not for spatial MPI runs due to the bug fixes in the MPI drivers:
    ```
    2025-03-05 17:37:00,335 - INFO - benchcab.benchcab.py:380 - Running comparison tasks...
    2025-03-05 17:37:00,358 - INFO - benchcab.benchcab.py:381 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
    2025-03-05 17:40:02,866 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
    ```

----
📚 Documentation preview 📚: https://cable--544.org.readthedocs.build/en/544/


src/offline/netcdf_utils.F90
SeanBryan51, line 1: - [ ] Formatting
- [ ] Documentation
- [ ] Support other types and dimensions
- [ ] Unit testing?

SeanBryan51:
I've now fixed a bug in the MPI drivers - the MPI and serial restart files now match.

SeanBryan51:
Running it again through benchcab now.

SeanBryan51:
I noticed I introduced a bug in the serial driver where the met file was being closed twice and was causing CABLE to crash. I've now fixed and run benchcab again (updated PR description with benchcab logs). @bschroeter should be good to go

SeanBryan51:
@har917 this PR includes several bug fixes in the MPI code - do you mind taking a look?

src/offline/cable_mpimaster.F90
```diff
@@ -680,6 +750,8 @@ SUBROUTINE mpidrv_master (comm, trunk_sumbal, dels, koffset, kend, PLUME, CRU)
 !$          CALL read_casa_dump( ncfile, casamet, casaflux, casa_it, kend, .FALSE. )
 !$        ENDIF
 
+          ! Zero out lai where there is no vegetation acc. to veg. index
+          WHERE ( iveg%iveg(:) .GE. 14 ) iveg%vlai = 0.
```
har917, line 754: This seems odd 
- the same line occurs further up
- I would have expected that if `iveg%vlai` is changed then `veg%vlai` should be changed 

and, while probably not this issue, it would be better if we can not hard-wire indexes.

src/offline/cable_serial.F90
```diff
@@ -338,6 +338,10 @@ SUBROUTINE serialdrv(trunk_sumbal, NRRRR, dels, koffset, kend, GSWP_MID, PLUME,
           LOY = 365
           kend = NINT(24.0*3600.0/dels) * LOY
 
+        CASE ('gswp3')
```
har917, line 341: I'm not sure of the detail of the GSWP3 data so comment around whether this is complete or not.  

However - for consistency with the other data - do we need to handle leap years and `kend`?

src/offline/cable_mpimaster.F90
```diff
@@ -500,6 +500,76 @@ SUBROUTINE mpidrv_master (comm, trunk_sumbal, dels, koffset, kend, PLUME, CRU)
...
+          imet%tvair(:) = met%tvair(:)
+          imet%tvrad(:) = met%tvrad(:)
+          imet%pmb(:) = met%pmb(:)
+          imet%ua(:) = met%ua(:)
+          imet%qv(:) = met%qv(:)
+          imet%qvair(:) = met%qvair(:)
+          imet%da(:) = met%da(:)
+          imet%dva(:) = met%dva(:)
+          imet%coszen(:) = met%coszen(:)
+          imet%Ndep(:) = met%Ndep(:)
```
har917, line 523: So the only difference between `imet%` and `met%` is `%Pdep`?  

I'm not sure of the reasoning why a new instance of the TYPE is needed - and possibly (for longer term use) seems easier to carry `%Pdep` in `imet%` as well.

src/offline/cable_mpimaster.F90
```diff
@@ -500,6 +500,76 @@ SUBROUTINE mpidrv_master (comm, trunk_sumbal, dels, koffset, kend, PLUME, CRU)
...
+
+          iveg%canst1(:) =  veg%canst1(:)
```
har917, line 525: so this is a full copy of `veg%` - why do we need a second instance?

har917 (commented):
@SeanBryan51 I think this is okay.  The MPI changes look consistent (maybe POINTERS not a copy would better for the new `iveg%` and `imet%` instances)

I'm less on top of the details of the GSWP3 data structure so can't really comment/approve.  One thing to check is whether the time setting in `cable_input` - i.e. `if cable_user%MetType .NE.'gswp'` at line 832 needs to be changed.  

I guess running a single site from these gridded forcing data isn't one of our test cases.

src/offline/cable_mpimaster.F90
```diff
@@ -500,6 +500,76 @@ SUBROUTINE mpidrv_master (comm, trunk_sumbal, dels, koffset, kend, PLUME, CRU)
...
+          iveg%g0(:) = veg%g0(:)
+          iveg%g1(:) = veg%g1(:)
+          iveg%rootbeta(:) = veg%rootbeta(:)
+          iveg%gamma(:) = veg%gamma(:)
+          iveg%F10(:) = veg%F10(:)
+          iveg%ZR(:) = veg%ZR(:)
+          iveg%clitt(:) = veg%clitt(:)
+          iveg%disturbance_interval(:,:) = veg%disturbance_interval(:,:)
+          iveg%disturbance_intensity(:,:) = veg%disturbance_intensity(:,:)
+
```
har917, line 572: okay - slight problem possibly - `master_in_types()` is being called before the resetting of `iveg%vlai`.  

I think this means that the `%vlai` that is getting passed to the workers won't be tweaked to be consistent with the `%iveg`  value - which I guess is the intent. I'm also nervous that by changing `%vlai` but not other parameters we could be creating some weird cases.

src/offline/cable_mpimaster.F90
SeanBryan51: > the same line occurs further up

I presume the same line occurs further up because the master process communicates the "read ahead" inputs to the workers via `get_met_data` (i.e. `imet` and `iveg`) prior to entering the main loop. I don't really understand why there is a need for `imet` and `iveg` but this is the implementation which we are stuck with.

> I would have expected that if iveg%vlai is changed then veg%vlai should be changed

I suppose it eventually does get updated because `veg%vlai` in the worker process is set to `iveg%vlai` in the master process. In the master process, `veg%vlai` will be the result of what the workers return back to the master and is then written to the output file.

src/offline/cable_serial.F90
SeanBryan51: To be honest 'gswp3' is actually be a bit meaningless in this context because often configurations which use the 'gswp3' option may not be using GSWP3 forcing data (the spatial test I'm using to test this is using CRU JRA forcing). I'm not too familiar with the GSWP3 data either so I'd prefer to not to comment.

There doesn't seem to be any support for handling leap years for the gswp3 case in the MPI driver so I didn't add any leap year support to the serial driver.

src/offline/cable_mpimaster.F90
SeanBryan51: Looks like I forgot to add `%Pdep` - `imet` should be an exact copy of `met` at this point in the code so I'll add this in for correctness.

src/offline/cable_mpimaster.F90
SeanBryan51: I don't fully understand the purpose behind the `imet` and `iveg` "read ahead" variables or why this was necessary - but it was necessary to explicitly copy the data across due to the initialisations in `load_parameters`. I did actually try replacing all occurrences of `imet` with `met` however this resulted in the MPI communication to fail.

src/offline/cable_mpimaster.F90
SeanBryan51: I believe `master_in_types` is only responsible for creating the MPI type for communicating input fields for the workers and doesn't actually perform the communication (this is done in `master_send_input`).

> I think this means that the %vlai that is getting passed to the workers won't be tweaked to be consistent with the %iveg value

You are right that `iveg` and `veg` become inconsistent after setting `iveg%vlai` - however from the perspective of the worker processes this does not matter as `iveg` is being sent to the workers where it is then assigned to `veg`.

The `veg` variable in the master process from this point on is set to the value of `veg` returned from the worker processes, then it is written to the CABLE output file.

> I'm also nervous that by changing %vlai but not other parameters we could be creating some weird cases.

To be honest, I'm nervous that weird cases were already present in the MPI code prior to this change (e.g. not zeroing out the LAI correctly). Without these changes, running in MPI mode will give different results when compared to running in serial (for this configuration).

SeanBryan51:
> maybe POINTERS not a copy would better for the new iveg% and imet% instances

This should be equivalent to replacing `imet` with `met` and `iveg` with `veg` everywhere in the driver. Unfortunately this doesn't work with the way the communication is set up.

> One thing to check is whether the time setting in cable_input - i.e. if cable_user%MetType .NE.'gswp' at line 832 needs to be changed.

https://github.com/CABLE-LSM/CABLE/blob/f603fc517cb81e27594c9253722adf01439d095d/src/offline/cable_input.F90#L832-L836

It looks like this will incorrectly assign a local time coordinate if we run at a single grid cell with gswp3 enabled (or any other spatial configuration for that matter). IMO it's a bit out of the scope for this PR as I'm not too concerned about running at a single grid cell, however I will make a note of it.

bschroeter (approved):
I can't comment on the points raised by @har917, but I am satisfied as long as the tests pass. LGTM.

PR #545, opened 2025-02-12
move snow aging call to avoid need for osnowd in restart

### LINKED ISSUE #539: investigate moving snow_Aging() routine
The snow aging routine (`cbl_snowAging()`) is currently called as part of the evaluation of the albedo (snow age is a contributor to albedo in snow-impacted areas).  The snow aging routine needs to know the change in snow amount (`%snowd - %osnowd`) from the previous time step.  Since the albedo is called first this requires that snow amount in previous time step is available - strictly this includes on the first time step and so `%osnowd` needs to be in the restart/initalisation.

It would be preferable (at least for the coupled configurations) that this not be a requirement.

However - conceptually at least - it should be possible to move the snow aging routine to avoid this need.  Two element

- set `%osnowd = %snowd` at the start of the time step (part of `cbm()`)
- move the call to `snowAging()` from current location to after the soil/snow time stepping - `snowAge` can then use the snow incremented during the current time step to age the snow.

This move would not be bitwise reproducible but would
- address restart issue previously noted
- consolidate the time stepping of CABLE's state variables in (roughly) the same part of the sequence (see also #505)

har917:
# CABLE

## Description
As per #539 it would be useful (for coupled applications) if there was no need for `%osnowd` in the restart file.
The current need stems from the call to `snow_aging()` prior to the first setting of `%osnowd` as part of the time stepping.
Moving the call to `snow_aging` to after the call to `soil_snow` (or `sli_main`) avoids this as both `osnowd` and `snowd` are evaluated in those routines.

This change will not be bitwise reproducible - but is likely scientifically of no impact. 

There are partner changes that are needed in the coupled routines (particularly CM2 and AM3) which will be included in this issue - but have not been tested.

Once accepted we may also be able to review/remove the need for `osnowd` in the offline restart and MPI sections.

Fixes #539 

## Type of change

- [ ] Bug fix
- [x] Minor change in science algorithm sequencing 

## Checklist

## Testing

- [x] Are the changes non bitwise-compatible with the main branch because of a bug fix or a feature being newly implemented or improved? If yes, add the link to the modelevaluation.org analysis versus the main branch or equivalent results below this line.

----
📚 Documentation preview 📚: https://cable--545.org.readthedocs.build/en/545/

### DIFF
--- a/src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
+++ b/src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
@@ -13,7 +13,7 @@ SUBROUTINE cbm_expl( mp, nrb, ktau,dels, air, bgc, canopy, met,                &
 USE cbl_albedo_mod,            ONLY: albedo
 USE cbl_init_radiation_module, ONLY: init_radiation
 USE cbl_masks_mod,             ONLY: fveg_mask, fsunlit_mask,  fsunlit_veg_mask
-USE snow_aging_mod,            ONLY: snow_aging
+!USE snow_aging_mod,            ONLY: snow_aging  !539 no longer needed
 USE cable_roughness_module,    ONLY: ruff_resist
 USE cable_air_module,          ONLY: define_air
 USE cable_canopy_module,       ONLY: define_canopy
@@ -110,8 +110,9 @@ CALL init_radiation( rad%extkb, rad%extkd,                                     &
                    ) !reducedLAIdue2snow 
 
 !Ticket 331 refactored albedo code for JAC
-CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
-                ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm) 
+!#539 - move snow aging. 
+!CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
+!                ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm) 
 
 !explicit ONLY
 CALL Albedo( ssnow%AlbSoilsn, soil%AlbSoil,                                &
@@ -302,6 +303,10 @@ CALL define_canopy(bal,rad,rough,air,met,dels,ssnow,soil,veg, canopy,climate, su
 
 CALL soil_snow(dels, soil, ssnow, canopy, met, bal,veg)
 ssnow%deltss = ssnow%tss-ssnow%otss
+
+!#539 - move snow aging. 
+CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
+                ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm) 
 
 ! reset tss & wetfac to value corresponding to beginning of timestep 
 IF( cycleno .NE. numcycles ) THEN
--- a/src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
+++ b/src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
@@ -109,11 +109,6 @@ CALL init_radiation( rad%extkb, rad%extkd,                                     &
                      canopy%vlaiw                                              &
                    ) !reducedLAIdue2snow 
 
-!Ticket 331 refactored albedo code for JAC
-!#539 - move snow aging. 
-!CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
-!                ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm) 
-
 !explicit ONLY
 CALL Albedo( ssnow%AlbSoilsn, soil%AlbSoil,                                &
              !AlbSnow, AlbSoil,              
@@ -304,7 +299,7 @@ CALL define_canopy(bal,rad,rough,air,met,dels,ssnow,soil,veg, canopy,climate, su
 CALL soil_snow(dels, soil, ssnow, canopy, met, bal,veg)
 ssnow%deltss = ssnow%tss-ssnow%otss
 
-!#539 - move snow aging. 
+!Ticket 331 refactored albedo code for JAC, #539 - moved snow aging. 
 CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
                 ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm) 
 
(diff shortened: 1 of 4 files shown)


har917:
re implementation in CM2 and AM3 - snow age state variable `ssnow%snage` was already in the prognostic bank so this does not need to be modified.

har917:
benchcab tests - unsurprisingly bitwise tests failed.  More interestingly this only happened at 7 (of 42 sites) - IT-Lav, US-FPe, US-GLE, US-Ha1, US-Me2, US-MMS and US-NR1.

It maybe interesting to chase these down further - are these sites the only sites with snow in the restart/initial conditions? For these runs it appears that initial conditions are coming from offline_CSIRO_1x1.nc

IT-Lav site: latitude = 45.95, longitude=11.28, snowd = 0.034m
US-Fpe :  latitude = 48.3, longitude=-105.1, snowd = 0.048m
US-GLE :  latitude = 41.3, longitude=-106.2, snowd = 0.070m
US-Ha1: latitude = 42.5, longitude=-72.2, snowd = 0.020m
US-Me2: latitude = 44.5, longitude=-121.6, snowd = 0.066m
US-MMS: latitude = 39.3, longitude=-86.4, snowd = 0.016m
US-NR1: latitude = 40.0, longitude=-105.5, snowd = 0.014m

i.e. the initial conditions all have snow, but with very small amount.  

The move of the snow_aging routine means that it is potentially sensitive to what happens on the first time step (snowd, osnowd, ssdn and tggsn that gets into the snow_Aging routine) - osnowd will be different on the first call to snow_Aging and ssdn and tggsn could well be different e.g. if the snow pack disappears during that first time step (as could happen with v. small amounts of snow).

Otherwise - it seems that the change has no impact!

Running through me.org

har917:
benchcab output file and yaml file for completed tests.
[benchmark_cable_qsub.sh.o135167668.txt](https://github.com/user-attachments/files/18762911/benchmark_cable_qsub.sh.o135167668.txt)
[config.yaml.txt](https://github.com/user-attachments/files/18762912/config.yaml.txt)

JhanSrbinovsky:
@har917 When this was implemented it was thought to be necessary for calculating snow albedo on the first timestep. IF it is moved to later what will this do the albedo

har917:
> IF it is moved to later what will this do the albedo.

History:  back in CABLE2.x days the albedo and snow aging routines were wrapped in together.  Issue 331 refactored the code so now CABLE3-albedo only depends on the snow depth, age, density and temperature, not  old snow depth.

However - it's not quite as simple as this.  This issue in effect moves the `snow_aging` routine from after the call to `albedo` (current time step) to the end of the previous time step - order of operations was `albedo` then `snow_aging`, it's now `snow_aging` then albedo.  Tracking dependencies is too complicated but my expectation was that there would be differences but not scientifically important ones.

The results seen so far indicate it's even better than this - and it is only making any difference for cases which initialise with a very small amount of snow.  I suspect that these come about because either i) the new routine allows for the complete removal of snow pack on the first time step whereas the old routine maintained it (or vice versa) and/or ii) the value of `%tss` that is getting passed into `snow_aging` is slightly different.

----------------------------------
Follow up on the second cause - looking at the code again I think I've (inadvertently) also fixed a bug.

In offline serial mode it looks like `%tss` is read in from the restart file (under `load_parameters`) and then promptly overwritten by `%tgg` (line 417 in `cable_serial`), i.e. the value of `%tss` that gets passed into snow_aging (CABLE3 MAIN) on the first time step is wrong.

The corresponding line in mpi_worker is not there.

This means that by moving snow_aging I've also removed the dependency on this bug in the serial model - @ccarouge is it worth removing that line as part of this PR?

What is wrong in both serial and mpi versions is that `%otss` and `%otss_0` (which I don't think is ever used anymore) are being initialised to `%tgg` not to `%tss`.

har917:
[me.org output](https://modelevaluation.org/modelOutput/display/9vXZpHnWdLHQozaKM) - at this aggregate level there is no discernable impact from this change.

har917:
@ccarouge sending for main review - noting questions around whether we need to address the surrounding i/o code better or not.

@JhanSrbinovsky this PR includes edits to the `cbm()` routines in each of the applications - these have not been tested.  We should devise a test this aspect of the PR as well (likely requires PRs in UM7 and AM3 repos)

ccarouge (commented):
Why do the change for CM2? I don't think we'll ever implement changes to CM2 (the user base is too small).

har917:
> Why do the change for CM2?

Because we should be trying to keep the repo internally consistent - not changing the CM2 code will mean that the cbm's in the different applications are fundamentally different -> confusion.

To be honest - the better way forward is to remove the CM2 (and ESM1.5) code from the repo entirely since we're not supposed to be modifying (expecting to modify) the code for these applications any more.

src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
JhanSrbinovsky, line 1: I'll port this file manually to ACCESS3-JULES

src/coupled/ESM1.5/CABLEfilesFromESM1.5/cable_cbm.F90
JhanSrbinovsky, line 1: I'll port this file manually to UM7

JhanSrbinovsky (approved):
I'll have to port this file manually to ACCESS3-JULES and UM7 - we need them now

har917:
@ccarouge Are you happy for this to go into MAIN? or do you want me to unpick the CM2 change?

@JhanSrbinovsky we probably need short AM3 and ESM1.6 runs with this in before merging I suspect.

JhanSrbinovsky:
output in feat-snow_aging-expt-c20a3efa/ for ESM1.6

AM3 - rose-calc is giving grief at the minute

src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
```diff
@@ -110,8 +110,9 @@ SUBROUTINE cbm_expl( mp, nrb, ktau,dels, air, bgc, canopy, met,                &
                    ) !reducedLAIdue2snow 
 
 !Ticket 331 refactored albedo code for JAC
-CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
-                ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm) 
+!#539 - move snow aging. 
+!CALL snow_aging(ssnow%snage,mp,dels,ssnow%snowd,ssnow%osnowd,ssnow%tggsn(:,1), &
+!                ssnow%tgg(:,1),ssnow%isflag,veg%iveg,soil%isoilm)
```
ccarouge, line 116: Why are we keeping the lines commented out? Nobody is going to come back to it and clean up after. Can we remove the lines before merging in please?

Otherwise I'm happy for this to go. Don't worry about the CM2 changes, that directory will be removed with the shuffling of files.

ccarouge (approved):

src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
har917: done - see 8ab0153

PR #551, opened 2025-02-25
reconcile with dev-access-esm1.6 - interface code

JhanSrbinovsky:
# CABLE

## Description
Reconcile developments in ESM1.6 with CABLE:main

There are changes across utilities since directories as well BUT here it is only those in coupled/ that are considered.

Post shuffling these files around to facilitate building a library. What happens to them is irrelevant really. 

These modified and added files are from dev-access-esm1.6. 

## Testing

So - I haven't actually done the testing BUT considering none of these fits are even in the build they cant impact anything

----
📚 Documentation preview 📚: https://cable--551.org.readthedocs.build/en/551/

### DIFF
--- a/src/coupled/ESM1.5/CABLEfilesFromESM1.5/allocate_soil_params_cbl.F90
+++ b/src/coupled/ESM1.5/CABLEfilesFromESM1.5/allocate_soil_params_cbl.F90
@@ -12,7 +12,7 @@ CONTAINS
 
 SUBROUTINE allocate_soil_parameter_type(var, mp)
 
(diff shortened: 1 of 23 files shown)


ccarouge (approved):
Good to go once you fixed the conflict with the snow aging stuff. Ian put some changes directly in CABLE repo for the ESM1.6 so needs to make sure what we need to see at the end. There are some commented lines that would be best to remove instead of using comments.

JhanSrbinovsky:
..... starting at line 157 in the new version

PR #572, opened 2025-03-19
Add documentation on running CABLE with CASA-CNP

SeanBryan51:
This PR adds documentation on how to run CABLE offline with CASA-CNP.

## Type of change

## Checklist

## Testing

----
📚 Documentation preview 📚: https://cable--572.org.readthedocs.build/en/572/


documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+
+The [CASA (Carnegie-Ames-Stanford Approach)-CNP (Carbon-Nitrogen-Phosphorous)](https://carbonwaterobservatory.csiro.au/casa.html) model simulates the biogeochemistry component of the CABLE land surface model. CASA-CNP models the dynamics of carbon pools and nitrogen and phosphorous limitations and is directly coupled to CABLE.
+
+## Getting started:
+
+### Initialisation (spin-up):
+
+Similar to other CABLE state variables, spin-up is required to obtain the initial conditions for CASA-CNP state variables (e.g. carbon and nitrogen pool sizes) for the simulation of interest. This is usually achieved by first running the model from a "cold start" over many periods of repeated meteorological forcing until there is no trend in the internal state variables.
+
+*Note:* this details the spin-up for CASA-CNP state variables only. The spinup for soil moisture, soil temperature and other CABLE state variables is enabled via the `spinup` (TODO(Sean): link) namelist option and is separate functionality.
```
SeanBryan51, line 11: Link to `spinup` namelist option.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+- `casafile%cnpflux = <output_file_path_for_casa_fluxes>`
+  - CASA output fluxes in CSV (`.csv`) format.
+- `casafile%out = <output_file_path_for_casa_output>`
+  - CASA output variables in NetCDF (`.nc`) format.
+
+To run CABLE with CASA-CNP from a CASA restart file (warm start), set `cable_user%CASA_fromZero = .FALSE.` and `casafile%cnpipool` to the input file path of the CASA restart file.
+
+### Accelerated spin-up (semi-analyical solution method):
+
+To avoid running long and computationally expensive spin-up simmulations, CABLE contains a semi-analytical solution method which reduces the spin-up time (TODO(Sean): cite 10.5194/gmd-5-1259-2012). Note, this method should be used only for simulations where the phosphorus cycle and its coupling with carbon and nitrogen cycles has been switched off (i.e. `icycle = 2`).
```
SeanBryan51, line 36: Cite 10.5194/gmd-5-1259-2012

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+
+TODO(Sean): confirm this cannot be used with phosphorus coupling enabled.
```
SeanBryan51, line 38: This is because in Xia's paper on semi-analytic spin-up, the phosphorus cycle and its coupling with carbon and nitrogen cycles was not activated.

@mcuntz @juergenknauer @har917 to confirm, do you know if the semi-analytic spin-up method works for phosphorus coupling?

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+
+The semi-analytic method first runs a short spin-up to obtain steady-state NPP, time averaged state variables for one period of repeated forcing variables, and C/N ratios. The second step is to analytically solve the differential equation to calculate steady-state carbon and nitrogen pool sizes. The last step is to make an addtional spin-up to meet the steady-state criterion for all pools.
+
+To run CABLE with CASA-CNP with the semi-analytic spin-up, CABLE to CASA dump files need to first be generated for each model year throughout one period of the repeated forcing. This is required so that the time averaged state variables can be calculated in the first stage of the semi-analytic spinup. The CABLE to CASA dump files can be generated by enabling the CABLE namelist option `cable_user%CASA_DUMP_WRITE = .TRUE.`. By default the dump files will be written to a path prefixed by the `casafile%c2cdumppath` CABLE namelist option.
```
SeanBryan51, line 42:
```suggestion
To run CABLE with CASA-CNP with the semi-analytic spin-up, CABLE to CASA dump files need to first be generated for each model year throughout one period of the repeated forcing. This is required so that the time averaged state variables can be calculated in the first stage of the semi-analytic spinup. The CABLE to CASA dump files can be generated by enabling the CABLE namelist option `cable_user%CASA_DUMP_WRITE = .TRUE.`. By default, the dump files will be written to a path prefixed by the `casafile%c2cdumppath` CABLE namelist option.
```

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+
+Once the CABLE to CASA dump files have been generated for one period of the repeated forcing, the semi-analytic spin-up can be enabled by setting `spincasa = .TRUE.`. This will spin-up CASA-CNP using the semi-analytical solution and then proceed to run CABLE.
+
+TODO(Sean): to confirm role of `cable_user%CASA_DUMP_READ` in semi-analytic spin-up.
```
SeanBryan51, line 46: @mcuntz @juergenknauer @har917 for the TRENDY and BIOS configurations, `cable_user%CASA_DUMP_READ = .TRUE.` is set for stages 3b. analytic quasi-equilibrium of biomass pools, 4b. analytic quasi-equilibrium of biomass pools and 5. First dynamic land use (Initialise land use). From looking at the code, it looks like the option causes CASA-CNP to be forced by the yearly CABLE to CASA dump files. What is the reasoning for doing this in TRENDY/BIOS? Is it relevant for the spin-up?

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Steady-state criterion for CASA-CNP spin-up:
+
+TODO(Sean): how to verify the spinup has converged to steady-state
```
SeanBryan51, line 50: Is there a check in the model that determines convergence? If not what output files and variables should the user look at to ensure convergence? What should the steady-state criterion be?

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
juergenknauer: the semi-analytic spinup does not work correctly without P cycling. Most likely the same problems exist when P cycling is turned on.

SeanBryan51:
@mcuntz @juergenknauer @har917 this is a first documentation draft on how to run CABLE offline with CASA-CNP. Any comments/feedback from you would be extremely helpful.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,58 @@
...
+- `casafile%cnpepool = <output_file_path_for_casa_restart>`
+  - CASA restart file containing CASA state variables. It can be fed back to another simulation for the purpose of spin-up via `casafile%cnpipool`.
+- `casafile%cnpflux = <output_file_path_for_casa_fluxes>`
+  - CASA output fluxes in CSV (`.csv`) format.
+- `casafile%out = <output_file_path_for_casa_output>`
+  - CASA output variables in NetCDF (`.nc`) format.
+
+To run CABLE with CASA-CNP from a CASA restart file (warm start), set `cable_user%CASA_fromZero = .FALSE.` and `casafile%cnpipool` to the input file path of the CASA restart file.
+
+### Accelerated spin-up (semi-analyical solution method):
```
SeanBryan51, line 34:
```suggestion
### Accelerated spin-up (semi-analytical solution method):
```

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
har917: @SeanBryan51 The two CABLE-POP communities are currently divided on the use of the quasi-analytic spin up.  @juergenknauer - as I understand it - does not use this capability in his TRENDY runs preferring to run the model normally for longer; for BIOS @AlisonBennett and I do.

We've found - that while it's not strictly correct - we do shorten the spin up needed for our runs.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
har917: Realistically this is a user choice - some use cases such as TRENDY have specified conditions on what convergence means as part of the protocol.

Typically assessing convergence will involve an expert determination as to whether the model spin up has converged - an a bit of gut feeling.  The determination will involve consideration of trends vs variability in the carbon and nitrogen pools, and long-term average and any trends in the key fluxes such as GPP, NPP and NBP.

Note there is capability in the driver code which enforces rerunning of the biophysical model until convergence model - I've never liked these because a) they're somewhat arbitrary (see expert determination above) and b) makes the code significantly more illegible.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
juergenknauer: I do use the quasi-analytic spinup for TRENDY runs. After applying the quasi-analytic spinup, I run a longer normal spinup to make sure that C pools have stabilised. That's because, from my observations, the quasi-analytic spinup works well for some regions but not so well for others. By using this approach I can make sure that pools have stabilised in most regions. Not doing the quasi-analytic spinup at all would (I think) require really long spinup time periods as some C pools (e.g. in boreal/tundra regions) could take millennia to reach equilibrium.

SeanBryan51: > The model then reruns the daily time stepping of CASA and POP `mloop=30` times.

@har917 during the `mloop` reruns (after solving for the equilibrium solution), do you know why we force CASA with the daily dump files as opposed to letting CASA be free running? I think this is the source of my confusion regarding the purpose of setting `cable_user%CASA_DUMP_READ = .TRUE.`

har917: @SeanBryan51 A bit of background:  CABLE inherently needs to run at sub-diurnal timesteps (ideally around 30 minutes, but some users push out to 3 hours). CASA always works on daily time steps - the inputs from CABLE to CASA are accumulated in the `sumcasaflux%` through the day and passed into CASA.  These inputs are needed regardless of whether CASA is being run coupled to CABLE or in isolation - the only part of the model that avoids this is the analytic `analyticpool` (which uses average the inputs up).

What option `cable_user%CASA_DUMP_READ = .TRUE.` does is to read in dump files of the daily variables passed into CASA (created during a previous phase where `cable_user%CASA_DUMP_WRITE = .TRUE.`) and use that to force CASA on its own.  This allows the user to avoid having to run the entirety of CABLE lots of times when spinning up the CNP pools (CABLE is a significant fraction of CABLE-CASA compute).

Now `analyticpool` forces the CNP pools to the analytic equilibrium solution assuming that all the inputs are constant - this isn't the same (but is close) to the equilbrium solution of CASA when forced by the true daily time series from `sumcasaflux%`.  Hence the model reruns CASA `mloop` times after the `analyticpool`.

----------------------
Shorter response - all of these steps are designed to get as close to a true spin up as possible in an efficient manner.  `analyticpool` solution gets you a lot of the way there, rerunning CASA with pre-determined inputs is the next step, then a long run using fully coupled CABLE-CASA is the last step.

SeanBryan51: Thanks, that clears up things. I just noticed that `icycle = 12` whenever `cable_user%CASA_DUMP_READ = .TRUE.` to run CASA without CABLE which would indeed speed things up. I will add a section on running "offline" CASA without CABLE.

SeanBryan51:
I've integrated the feedback - this is now ready for review.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+To run CABLE with CASA-CNP (cold start), the following CABLE namelist options will need to be configured:
+
+| Namelist option (`cable.nml`)            | Description                                                  |
+| ---------------------------------------- | ------------------------------------------------------------ |
+| `l_casacnp = .TRUE.`                     | Enables CASA-CNP (must be switched on when `icycle` is non-zero). |
+| `icycle = <icycle_value>`                | `icycle` value of  `1`, `2` or `3` enable coupled carbon, carbon-nitrogen or carbon-nitrogen-phosphorus dynamics respectively. |
+| `cable_user%CASA_fromZero = .TRUE.`      | Enable cold start without CASA restart file.                 |
+| `casafile%cnpbiome = <input_file_path>`  | Input file path to PFT lookup table used by CASA-CNP.        |
+| `casafile% = <input_file_path>`          | Input file path to vegetation phenology file. The vegetation phenology file contains a latitudinal listing of phenology dates for various vegetation types. |
```
ccarouge, line 19: Missing the option name

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+| `casafile%cnpepool = <output_file_path>` | Output file path to the CASA restart file containing CASA state variables. It can be fed back to another simulation for the purpose of spin-up via `casafile%cnpipool`. |
+| `casafile%cnpflux = <output_file_path>`  | Output file path to CASA output fluxes in CSV (`.csv`) format. |
+| `casafile%out = <output_file_path>`      | Output file path to CASA output variables in NetCDF (`.nc`) format. |
```
ccarouge, line 22: Does `casafile%out` contain the fluxes as well or are the fluxes only in `casafile%cnpflux`? The term "CASA output variables" is too vague to know.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+To run CABLE with CASA-CNP from a CASA restart file (warm start), set `cable_user%CASA_fromZero = .FALSE.` and `casafile%cnpipool` to the input file path of the CASA restart file.
+
+## Running CASA-CNP in offline mode:
+
+It can be useful to run CASA-CNP in offline (or CASA only) mode for the purpose of spinning up CASA-CNP more efficiently. Since CASA-CNP operates on a daily time step as opposed to CABLE which inherently runs on a sub-diurnal time step, CABLE makes up a significant fraction of CABLE-CASA compute. For the purposes of spinning up CASA-CNP, two way coupling between CABLE and CASA-CNP is not strictly necessary and CASA-CNP can be forced using CABLE to CASA input meteorology. This can reduce the amount of compute time spent on CASA-CNP spin-up.
```
ccarouge, line 28: I'm not sure "meteorology" is the right term here for "using CABLE to CASA input meteorology". But I don't know what fields are passed so maybe it is.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+To run CASA-CNP in offline mode, CABLE to CASA dump files need to first be generated for each model year throughout spin-up period. The CABLE to CASA dump files can be generated by enabling the CABLE namelist option `cable_user%CASA_DUMP_WRITE = .TRUE.`. By default, the dump files will be written to a path prefixed by the `casafile%c2cdumppath` CABLE namelist option.
```
ccarouge, line 30: What does "each model year" mean here? Since for a spin-up we are repeating a period of years several times, do we need to create the files only over one run of the period of years used for the spin-up or over the full repetition of the years? I'm guessing it's the former since the number of repetition might not be known.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+Once the CABLE to CASA dump files have been generated for the spin-up period, CASA-CNP can be run in offline mode by enabling the CABLE namelist options `CASAonly = .TRUE.` and `cable_user%CASA_DUMP_READ = .TRUE.` (and `cable_user%CASA_DUMP_WRITE = .FALSE.` to prevent overwriting the CABLE to CASA input meteorology). The CABLE to CASA input file paths will be read from the same path as they were written initially. Also, note that `CASAonly` may be set implicitly by setting `icycle` to a value greater than `10`.
+
+## Accelerated spin-up (semi-analytical solution method):
+
+To avoid running long and computationally expensive spin-up simmulations, CABLE contains a semi-analytical solution method which reduces the spin-up time ([Xia et al. 2012][gmd-5-1259-2012]). Note, this method should be used only for simulations where the phosphorus cycle and its coupling with carbon and nitrogen cycles has been switched off (i.e. `icycle = 2`).
```
ccarouge, line 36:
```suggestion
To avoid running long and computationally expensive spin-up simulations, CABLE contains a semi-analytical solution method that reduces the spin-up time ([Xia et al. 2012][gmd-5-1259-2012]). Note, this method should be used only for simulations where the phosphorus cycle and its coupling with carbon and nitrogen cycles have been switched off (i.e. `icycle = 2`).
```

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+The semi-analytic method first runs a short spin-up to obtain steady-state NPP, time averaged state variables for one period of repeated forcing variables, and C/N ratios. The second step then involves analytically solving the differential equation to calculate steady-state carbon and nitrogen pool sizes. Lastly, an addtional spin-up is done to meet the steady-state criterion for all pools.
+
+To run CABLE with CASA-CNP with the semi-analytic spin-up, CABLE to CASA dump files need to first be generated for each model year throughout one period of the repeated forcing (see [Running CASA-CNP in offline mode](#running-casa-cnp-in-offline-mode)). This is required so that the time averaged state variables can be calculated in the first stage of the semi-analytic spin-up; and for doing a series of additional offline CASA-CNP spin-up cyles.
+
+Once the CABLE to CASA dump files have been generated for one period of the repeated forcing, the semi-analytic spin-up can be enabled by setting `spincasa = .TRUE.`. This will spin-up CASA-CNP using the semi-analytical solution and then proceed to run CABLE.
+
+## Steady-state criterion for CASA-CNP spin-up:
```
ccarouge, line 44: Does this paragraph mean the code does not have an in-built check? Does that mean people run the CASA-CNP for a given of repeated forcing years, then check the results against their criteria and then need to reset another period of spin-up if their criteria aren't reach?

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Prognostic LAI and vcmax
+
+Normally, CABLE uses prescribed LAI read from the gridinfo file or from the meteorological forcing file. When the CABLE namelist variable `l_laiFeedbk` is switched on, CASA-CNP will calculate LAI values to be passed back to CABLE for use. These prognostic LAI values are written out in the output file.
+
+CASA-CNP can also be used to predict the vegetation parameter `vcmax` which is normally prescribed in the [vegetation parameter file][pft_params_nml]. This is controlled by the CABLE namelist variable `l_vcmaxFeedbk` and requires the nitrogen cycle to be turned on (i.e. `icycle` set to `2` or `3`).
```
ccarouge, line 52: I would avoid using "normally". I would use "CABLE can use prescribed ..." and "`vcmax` which can otherwise be prescribed ...".

This is simply because both options (prognostic or prescribed) are valid scientific choices and one isn't better than the other. "Normally" seems to imply there is a normal and abnormal way to do things.

ccarouge (changes requested):
I have gone through quickly. It's good overall. I have put comments everywhere I think the text is a bit vague and might need to be clarified.

You do not talk at all about how to do repeat years of forcing. Is it by choice?

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
SeanBryan51: Good question. I wasn't able to find any information on what these output files contain and there is no metadata in either of the files. The answer is probably in the code:
- `casafile%cnpflux`: https://github.com/CABLE-LSM/CABLE/blob/6102496e2a47c685bb7260d080b01fca70bb789a/src/science/casa-cnp/casa_inout.F90#L577-L677
- `casafile%out`: https://github.com/CABLE-LSM/CABLE/blob/6102496e2a47c685bb7260d080b01fca70bb789a/src/shared/casa_offline_inout.F90#L580-L1471

It might be worth asking others who are more familiar with the CASA-CNP outputs.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
SeanBryan51: I agree I've renamed "CABLE to CASA input meteorology" to "CABLE to CASA forcing data".

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
SeanBryan51: I've updated to "each model year for one cycle of the spin-up period"

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
SeanBryan51: Done

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
SeanBryan51: I've clarified on the lack of an in-built check. On the repeated forcing I think this can be omitted since I'd consider this to be assumed knowledge for this guide.

SeanBryan51:
> You do not talk at all about how to do repeat years of forcing. Is it by choice?

Yes. I've assumed that the user already has someway of running the model with repeated forcing. Since this is done differently for various tools/scripts/configurations, I'd prefer to not go into the details. I'd say it deserves its own section focused on running biophysics only CABLE

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
ccarouge: I've checked and `cnpflux` has a subset of the variables in `out`. From some comments in the code, I believe `cnpflux` might have been a temporary inclusion that ended up being permanent. But I am guessing it doesn't necessarily output the fluxes at the same time step as in the netcdf file (depending on options).

Anyway, do you want to change the text to something like:

```suggestion
| `casafile%cnpflux = <output_file_path>`  | Output file path to output only the CASA fluxes in CSV (`.csv`) format. Yearly output after the spinup period. |
| `casafile%out = <output_file_path>`      | Output file path to output the CASA fluxes and other CASA variables in NetCDF (`.nc`) format. Output frequency is given by `cable_user%casa_out_freq`. |
```

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+To run CASA-CNP in offline mode, CABLE to CASA dump files need to first be generated for each model year for one cycle of the spin-up period. The CABLE to CASA dump files can be generated by enabling the CABLE namelist option `cable_user%CASA_DUMP_WRITE = .TRUE.`. By default, the dump files will be written to a path prefixed by the `casafile%c2cdumppath` CABLE namelist option.
+
+Once the CABLE to CASA dump files have been generated for the spin-up period, CASA-CNP can be run in offline mode by enabling the CABLE namelist options `CASAonly = .TRUE.` and `cable_user%CASA_DUMP_READ = .TRUE.` (and `cable_user%CASA_DUMP_WRITE = .FALSE.` to prevent overwriting the CABLE to CASA forcing data). The CABLE to CASA input file paths will be read from the same path as they were written initially. Also, note that `CASAonly` may be set implicitly by setting `icycle` to a value greater than `10`.
```
ccarouge, line 32:
```suggestion
Once the CABLE to CASA dump files have been generated for the spin-up period, CASA-CNP can be run in offline mode by enabling the CABLE namelist options `CASAonly = .TRUE.` and `cable_user%CASA_DUMP_READ = .TRUE.` (and `cable_user%CASA_DUMP_WRITE = .FALSE.` to prevent overwriting the CABLE to CASA forcing data). The CABLE to CASA input file paths will be read from the same path as they were written initially. Also, note that `CASAonly` may be set implicitly by setting `icycle` to a value greater than `10` but it is not the recommended way to configure a CASA-only experiment.
```

I'm thinking we should prepare the ground for deprecating duplicate functionality. The `CASAonly` way is more explicit than `icycle>10`, so it should be preferred.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Accelerated spin-up (semi-analytical solution method):
+
+To avoid running long and computationally expensive spin-up simmulations, CABLE contains a semi-analytical solution method which reduces the spin-up time ([Xia et al. 2012][gmd-5-1259-2012]). Note, this method should be used only for simulations where the phosphorus cycle and its coupling with carbon and nitrogen cycles has been switched off (i.e. `icycle = 2`).
```
ccarouge, line 36: Isn't it `icycle = 2 or 1`?

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
```diff
@@ -0,0 +1,57 @@
...
+
+Once the CABLE to CASA dump files have been generated for one period of the repeated forcing, the semi-analytic spin-up can be enabled by setting `spincasa = .TRUE.`. This will spin-up CASA-CNP using the semi-analytical solution and then proceed to run CABLE.
+
+## Steady-state criterion for CASA-CNP spin-up:
+
+Unlike the spin-up of other CABLE state variables, CASA-CNP does not contain an in-built check for determining if it has reached a steady-state. The steady-state criterion is typically left as a choice for the user. Users should manually check that there criterion is met before running their simulation of interest. Some experiments (for example TRENDY) specify the steady-state criterion as part of the experiment protocol. Assessing convergence will typically involve an expert determination as to whether the spin-up has converged. The determination will involve consideration of trends vs variability in the carbon, nitrogen and/or phosphorous pools, and long-term averages; and any trends in the key fluxes such as GPP, NPP and NBP. It is also important to note that different regions will equilibrate at different rates and that convergence should be checked for each region.
```
ccarouge, line 46:
```suggestion
Unlike the spin-up of other CABLE state variables, CASA-CNP does not contain an in-built check for determining if it has reached a steady-state. The steady-state criterion is typically left as a choice for the user. Users should manually check that their criterion is met before running their simulation of interest. Some experiments (for example TRENDY) specify the steady-state criterion as part of the experiment protocol. Assessing convergence will typically involve an expert determination as to whether the spin-up has converged. The determination will involve consideration of trends vs variability in the carbon, nitrogen and/or phosphorous pools, and long-term averages; and any trends in the key fluxes such as GPP, NPP and NBP. It is also important to note that different regions will equilibrate at different rates and that convergence should be checked for each region.
```

ccarouge (approved):
It's good for me. I propose a few minor additional changes, but nothing I need to review again.

documentation/docs/user_guide/running_cable/running_cable_with_casacnp.md
SeanBryan51: Yes, I can confirm from the Xia's paper that the method applies to C only simulations as well as C-N.

```suggestion
To avoid running long and computationally expensive spin-up simmulations, CABLE contains a semi-analytical solution method which reduces the spin-up time ([Xia et al. 2012][gmd-5-1259-2012]). Note, this method should be used only for simulations where the phosphorus cycle and its coupling with carbon and nitrogen cycles has been switched off (i.e. `icycle = 1 or 2`).
```

SeanBryan51:
@har917 @juergenknauer I've added you as co-authors. Thanks for your help putting this together. Let me know if you are happy with the documentation and I will merge it in.

ccarouge:
@SeanBryan51 Shouldn't this be merged?

Issue #526, opened 2025-01-14
Test configuration for the "merge" of NESP2pt9 and CABLE-POP_TRENDY


ccarouge:
At the end of the "BIOS merge", we will need to test the resulting code with a BIOS configuration. We will need to setup the configuration based on the pseudo-parallel TRENDY config.

BIOS config:
- ACT9
- 1000pts
- 0.25° Australia-wide
- 0.05° Australia-wide
- need some historical and future tests
- no LUC, no BLAZE.

Success criteria: good if bitwise comparable but very likely not achievable.

ccarouge:
First step, keep the inputs for TRENDY and change the science configs to mirror the BIOS config.

har917:
Some more detail/thoughts around this:

The three (four) tests cases should be simply differentiated by the specification of a landmask:
- BIOS (AGCD) meteorology data is provided at 0.05 degrees - aligned with the revised gridinfo
- ACT9 case is then defined uniquely via the specification of a landmask file (serial only needed)
- 0.05 degree case needs no additional landmask file.
- 1000pts and 0.25 degree cases are potentially more complicated as they require the combination of an externally provided landmask and the TRENDY partitioning/recombination.  
 
So the problematic cases - 2 elements.  1) Can the partitioning part of the process work on top of an externally specified landmask file? i.e. the landmasks created/used by the serial jobs need to be the overlay of the external landmask and the act of splitting it up.  2) Can the recombination step provide output on an externally specified landmask (or does it default to the input meteorology)?  

Another aspect to think through - the TRENDY partitioning process used some kind of randomisation of grid cells to ensure a reasonable mix of fast and slow (to compute) grid cells in each job.  This allows for efficient kSU usage and through flow during the multi-stage process.  Do we need to do the same randomisation of cells for BIOS or can we use a geographically defined (lat-lon box) as the means to split things up?  How would this impact the recombination script?

Whyborn:
I have got the ACT9 test case running with the TRENDY pseudo-parallel configuration. Some minor adjustments of the original BIOS landmask were required:

1. Change the mask variable name to ```'land'```
2. Change the type of the mask variable to ```Int8```

The TRENDY partitioning process uses a "tiling" approach- effectively walks through the the points in the order they appear in memory, and assigns them cyclically to each process. Say we wanted to run the ACT9 test case with 4 parallel jobs, it would assign:

| Job | Points |
| ----- | ------ |
| 1 | 1, 5, 9 |
[...]

PR #603, opened 2025-07-23
Make the code compute the diffuse fraction by default

Whyborn:
# CABLE

## Description

The ```CABLE-POP_TRENDY``` branch of the code currently sets ```cable_user%calc_fdiff=.FALSE.``` by default. The only place this was modified was in the CRU meteorology setup, so would effectively break anyone else's simulation not using this workflow who was unaware of this change.

This change changes the default to ```cable_user%calc_fdiff=.TRUE.```, so that the code computes the direct and diffuse fractions via the spitter function. The check used in the CRU/TRENDY has been moved to initialisation. If the user turns off diffuse fraction calculation (via the ```ReadDiffFrac=.TRUE.``` option in the CRU namelist), but doesn't also provide a dataset to read from, the code will now throw an error.

This is not an ideal fix, as we should be checking against the ```cable_user%calc_fdiff``` option, rather than the ```CRU%ReadDiffFrac``` option. I think a better option is to remove the ```CRU%ReadDiffFrac``` option altogether and only have the ```cable_user%calc_fdiff```, given they effectively contain identical information. That would also require minor adjustments to the TRENDY scripts.

Fixes at least part of the problems discussed in #526.

## Type of change

- [x] Bug fix

### DIFF
--- a/offline/cable_cru_TRENDY.F90
+++ b/offline/cable_cru_TRENDY.F90
@@ -81,7 +81,6 @@ TYPE CRU_TYPE
   REAL, DIMENSION(:), ALLOCATABLE ::  NDepVals  ! Nitrogen deposition values
 
   LOGICAL   :: DirectRead         ! Whether to read in entire met array
-  LOGICAL   :: ReadDiffFrac       ! Read diff fraction or calculate it
   LOGICAL   :: LeapYears
 
   LOGICAL, DIMENSION(nCoreVariables)    :: isRecycled
@@ -164,17 +163,13 @@ SUBROUTINE CRU_INIT(CRU)
   END IF
 
   ! Check for the diffuse fraction
-  IF (CRU%ReadDiffFrac) THEN
-    cable_user%calc_fdiff = .FALSE.
-    ! Make sure we have a dataset for fdiff
-    IF (TRIM(InputFiles(fdiff)) == "None") THEN
-      WRITE(ERROR_UNIT,*) "Specified to read diffuse fraction from"//&
-        " file, but no diffuse fraction file given."
-      ERROR STOP 1
-    END IF
-  ELSE
-    cable_user%calc_fdiff = .TRUE.
+  IF ((.NOT. cable_user%calc_fdiff) .AND.&
+    (TRIM(InputFiles(fdiff)) == "NONE")) THEN
+    WRITE(ERROR_UNIT,*) "Specified to read diffuse fraction from"//&
+      " file, but no diffuse fraction file given."
+    ERROR STOP 1
   END IF
+
   ! Initialise the weather generator
   CALL WGEN_INIT(WG, CRU%mLand, Latitude, REAL(CRU%DtSecs))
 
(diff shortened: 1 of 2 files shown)


har917 (approved):
@Whyborn Code changes look okay - we'll need some testing (short TRENDY runs before/after at the minimum) to demonstrate that this doesn't break something inadvertently (but I can't see how).

juergenknauer:
It looks good to me as well. I would also support to only use one of the flags. I am not sure which one though. One could argue that this is a met input thing and it would be better placed in the cru namelist.

offline/cable_cru_TRENDY.F90
```diff
@@ -163,6 +163,18 @@ SUBROUTINE CRU_INIT(CRU)
...
+    cable_user%calc_fdiff = .FALSE.
+    ! Make sure we have a dataset for fdiff
+    IF (TRIM(InputFiles(fdiff)) == "None") THEN
+      WRITE(ERROR_UNIT,*) "Specified to read diffuse fraction from"//&
+        " file, but no diffuse fraction file given."
+      ERROR STOP 1
+    END IF
+  ELSE
+    cable_user%calc_fdiff = .TRUE.
+  END IF
```
ccarouge, line 177: Wouldn't this be clearer:

```suggestion
  ! Calculate diffusive fraction if we are not reading it in.
  cable_user%calc_fdiff = .NOT. CRU%ReadDiffFrac
  
  ! Make sure we have a dataset for fdiff if set to read in
  IF (CRU%ReadDiffFrac .AND. (TRIM(InputFiles(fdiff)) == "None")) THEN
    WRITE(ERROR_UNIT,*) "Specified to read diffuse fraction from"//&
      " file, but no diffuse fraction file given."
    ERROR STOP 1
  END IF
```

ccarouge:
> It looks good to me as well. I would also support to only use one of the flags. I am not sure which one though. One could argue that this is a met input thing and it would be better placed in the cru namelist.

It is a met related thing but not limited to the CRU met forcings, so we can't just have it in the CRU namelist.

Whyborn:
I have made the change that drops back to using strictly the ```cable_user%calc_fdiff```. All instances of ```CRU%ReadDiffFrac``` have been removed. The way to set this variable should be through ```cable.nml```, so if this route is chosen, there will need to be some minor updates to the TRENDY configuration.

The only changes would be to ```run_cable.sh```:

1. Swap the order of the logical assignment [here](https://github.com/CABLE-LSM/TRENDY/blob/28c299cc9adef088dc9db2bf3eb0818ea039bdde/run_cable.sh#L274-L279)
2. Remove ```ReadDiffFrac``` from ```cru.nml``` [here](https://github.com/CABLE-LSM/TRENDY/blob/28c299cc9adef088dc9db2bf3eb0818ea039bdde/run_cable.sh#L307)
3. Add ```cable_user%calc_fdiff=${fdiff_bool}``` to the ```cable.nml``` template [here](https://github.com/CABLE-LSM/TRENDY/blob/28c299cc9adef088dc9db2bf3eb0818ea039bdde/run_cable.sh#L333-L375)

Let me know your thoughts- I can go back to the previous commit (+ suggestions from Claire) if that would be preferred.

ccarouge:
I'm happy with the proposed solution, but I'm not using the TRENDY configuration. @juergenknauer @mcuntz do you have any objections to the proposed changes in the namelists?

juergenknauer:
It all looks good to me. Just note that I had to branch off from the main TRENDY branch because the current version does not work with my setup. After this year's TRENDY runs we can try and merge them again and hopefully I can use the new setup next year.

Whyborn:
Branch off the TRENDY configuration, or branch off the TRENDY branch of the CABLE code?

juergenknauer:
branch off the code, because the problem lies somewhere in there. Unfortunately I couldn't see where yet. I would need to checkout after each commit, compile, and see if it works or not.

PR #607, opened 2025-08-05
Move blaze input routine to blaze if clause

Whyborn:
# CABLE

## Description

The climate parameters required for BLAZE are read in even when BLAZE is not active. The routine should be moved to the ```if (CALL_BLAZE)``` clause. This finishes the work of removing all binary inputs from the standard ```CABLE-POP_TRENDY``` mode of operation.

Fixes #(606)

### DIFF
--- a/offline/cable_driver.F90
+++ b/offline/cable_driver.F90
@@ -750,11 +750,11 @@ PROGRAM cable_offline_driver
                  end if
               end if
 
-              ! additional params needed for BLAZE
-              if (trim(cable_user%MetType) == 'bios') &
-                   call cable_bios_load_climate_params(climate)
-
               if (cable_user%CALL_BLAZE) then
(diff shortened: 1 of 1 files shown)


offline/cable_driver.F90
```diff
@@ -750,11 +750,11 @@ PROGRAM cable_offline_driver
                  end if
               end if
 
-              ! additional params needed for BLAZE
-              if (trim(cable_user%MetType) == 'bios') &
+              if (cable_user%CALL_BLAZE) then
```
har917, line 753: Please retain the explanatory comment - but otherwise okay.

har917 (changes requested):
minor change requested - otherwise okay.

abhaasgoyal (approved):

PR #615, opened 2025-09-24
Rename and use r2, r1, and i4 as kinds

mcuntz:
# CABLE

## Description

The numrical kinds used in cable were renamed and unified as much as possible. Changed `r_2` to `r2`, `r_1` to `r1`, `i_d` to `i4`. It was also tried to remove the kind `sp` and `dp` as much as possible. However, the latter kinds are still used in POP, the reading of BIOS files and my numerical utilities in `mo_utils.f90`. `dp` and `sp` in the latter point to `r1` and `r2` in the latter, though.
The kinds between POP and CABLE were synchronised. Only the definition of `sp` in BIOS is different to CABLE (it is also defined differently in POP but not used in the code).

As an example: a double precision numerical constant is now `1.0_r2` instead of `1.0_r_2` or `1.0_dp`. 

Fixes #(513)

## Type of change

- [X] Refactoring

## Checklist

- [X] The new content is accessible and located in the appropriate section

## Testing

- [X] Are the changes bitwise-compatible with the main branch?
Local tests were run, not benchcab. I used `cdo diff` on all restart and output files. However, Only one science configuration was tested.

### DIFF
--- a/core/biogeophys/cable_diag.F90
+++ b/core/biogeophys/cable_diag.F90
@@ -35,7 +35,7 @@
 
 !#define UM_BUILD YES
 MODULE cable_diag_module
-  use cable_def_types_mod, only : r_2
+  use cable_def_types_mod, only : r2
    IMPLICIT NONE
    INTEGER, PARAMETER :: gok=0
    INTEGER :: galloctest=1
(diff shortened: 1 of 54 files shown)


core/biogeophys/cable_diag.F90
```diff
@@ -35,7 +35,7 @@
 
 !#define UM_BUILD YES
 MODULE cable_diag_module
-  use cable_def_types_mod, only : r_2
+  use cable_def_types_mod, only : r2
```
ccarouge: Unused at this level, could remove since it's imported within individual subroutines.

PR #630, opened 2025-10-22
(#629) - Sync with the work on JULES

### LINKED ISSUE #629: Update following JAC's removal of CABLE science code
As part of the change of licence for JULES, it was decided that JAC would be pursued by keeping the CABLE code in its own repository and connecting the two code sources at compilation time.

As a consequence, the CABLE code as well as some interface code added to JULES has been removed for version 4.0. This issue is to add that code to this repository as necessary since it will be its new home.

ccarouge:
# CABLE

## Description

Fixes #629

This goes with the [work in JULES](https://code.metoffice.gov.uk/trac/jules/ticket/1633) to remove the CABLE files from JULES for the licence change.

The changes are as follow:

- remove files we decided belong to JULES for licensing reasons. Some of the files will need to be modified in JULES when we figure out the compilation to use the CABLE constant modules. But we can't keep the original of these files here (licensing reasons) so we'll need to refer to the JULES history for the details.
- Some science files had small differences between JULES and CABLE. Most were trivial and would not affect the results. A difference about `spitter()` was present but correspond to relatively new work in CABLE so I have chosen to accept the new version from CABLE. See notes:
   - *params/cable_phys_constants_mod.F90*: formatting and documentation differences only
   - *science/roughness/roughnessHGT_effLAI_cbl.F90*: differences in documentation only
   - *science/albedo/cbl_snow_albedo.F90*: different from JULES one but only because the `sfact` parameters have been given names instead of being hard-coded directly within the algorithm
   - *science/radiation/cbl_init_radiation.F90*: differences due to the documentation and the recent `spitter()` bug (#355) . Original lines for the radiation split in JAC:
```
RadFbeam(:,1) = spitter(mp, cpi, metDoy, coszen, SW_down(:,1))
RadfBeam(:,2) = spitter(mp, cpi, metDoy, coszen, SW_down(:,2))
```

Two parameter files show differences between JULES and CABLE and need to be brought in sync in future work. These are copied under JAC/params/cable

## Type of change

- Sync with JAC work

## Checklist

## Testing

The changes do not impact offline CABLE at all. They are linked to the JAC project and will only be used once we work on the new compilation to link JULES and CABLE.

----
📚 Documentation preview 📚: https://cable--630.org.readthedocs.build/en/630/

### DIFF
--- a/src/coupled/JAC/params/cable/cable_other_constants_mod.F90
+++ b/src/coupled/JAC/params/cable/cable_other_constants_mod.F90
@@ -36,7 +36,6 @@ INTEGER, PARAMETER ::                                                          &
   mf = 2,            & ! types of leaves (sunlit / shaded)
   r_2  = SELECTED_REAL_KIND(12, 50), &!this will be removed
                        ! double precision real dimension
-  niter = 4,         & ! number of iterations for za/L
   n_assim_rates = 3, & ! Rubisco, RuBP and Sink-limited rates of photosynthesis
   n_soiltypes = 9      ! number of soil types
 


src/coupled/JAC/params/cable/cable_other_constants_mod.F90
```diff
@@ -0,0 +1,53 @@
...
+                       ! (above divided into direct / diffuse)
+  mf = 2,            & ! types of leaves (sunlit / shaded)
+  r_2  = SELECTED_REAL_KIND(12, 50), &!this will be removed
+                       ! double precision real dimension
+  niter = 4,         & ! number of iterations for za/L
+  n_assim_rates = 3, & ! Rubisco, RuBP and Sink-limited rates of photosynthesis
+  n_soiltypes = 9      ! number of soil types
+
+REAL, PARAMETER ::                                                             &
+  max_snow_depth = 50000.0,  & ! maximum depth of lying snow on tiles (kg/m2)
```
har917: as an aside - this parameter (at least) should be converted to a namelist read since it varies between ESM1.6, CM2, AM/CM3 and offline.

I'm also not sure why some of the dimension setting numbers aren't set in the `grid_constants_mod`

src/coupled/JAC/params/cable/grid_constants_cbl.F90
```diff
@@ -0,0 +1,57 @@
...
+! Req'd to be defined at compile time to read in pars. strictly speaking these
+! only need to be greater than ntiles, nsoils (below). However, there is no
+! point in allocating useless space here
+INTEGER, PARAMETER :: ntype_max = 17 ! Max # tiles ! compile time constant
+INTEGER, PARAMETER :: nsoil_max = 9  ! Max # soils ! req'd to read in pars
+INTEGER, PARAMETER :: nsl       = 6  ! # soil layers !sm_levels in JULES IO
+INTEGER, PARAMETER :: nsnl      = 3  ! # snow layers
+INTEGER, PARAMETER :: nrb       = 3  ! # rad bands VISual/NIR + Legacy incl LW
+INTEGER, PARAMETER :: nrs       = 4  ! # streams (VIS+NIR)*(Direct+Diffuse)=4
+INTEGER, PARAMETER :: nsCs      = 2  ! # soil carbon stores
```
har917: why is this 2? there are 3 soil carbon stores.

har917 (changes requested):
Two changes needed I think
1. `niter` is defined twice
2. number of carbon stores `nsCs` should surely be 2

@JhanSrbinovsky please note point 2.   I am completely confused here.  Yes, there is a difference between the CASA variables and the bgc% variables - but the luc3 code (which is based on CASA) also appears to use `nsCs`, as does the sli code.

Basically worth checking - and there is a longer-term question around whether to maintain  the older `bgc%` code for carbon cycle work.

src/coupled/JAC/params/cable/grid_constants_cbl.F90
har917: In CASA there are 3 soil carbon stores (fast, slow, passive) - but those variables are defined with `msoil=3`.

`nsCs` is used in the `bgc%` TYPE - and co-opted by `SLI`/`landuse3` - and it could be that it is supposed to be 2.  

Perhaps leave for now.

src/coupled/JAC/params/cable/grid_constants_cbl.F90
```diff
@@ -0,0 +1,57 @@
...
+INTEGER, PARAMETER :: nvCs      = 3  ! # vegetation carbon stores
+INTEGER, PARAMETER :: ICE_SoilType = 9 ! SoilType Index (soilparm_cable.nml JAC)
+INTEGER, PARAMETER :: lakes_cable  = 16! SoilType Index (soilparm_cable.nml JAC)
+
+INTEGER, PARAMETER :: mf = 2          ! # leaves (sunlit, shaded)
+INTEGER, PARAMETER :: niter = 4       ! number of iterations for za/L
```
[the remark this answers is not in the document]
har917: Since it's JAC we won't ever have got to the point where the potential for conflict is an issue. 

Most of what is in `grid_constants` and `other_constants` is common across all apps - so ideally these MODULES would be in the library.  However for compilation purposes we'll need a version of these in JULES I think [I'm getting lost what's where now]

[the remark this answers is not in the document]
ccarouge: @har917 all CABLE constants will come from the library. For JULES currently (i.e. before the compilation is worked out), the constants are hard-coded in the JULES files with a comment to change when we work on the compilation.

Since these constant files are not currently used, I'll remove the `niter` declaration in `grid_constants_cbl.F90` (because `niter` is not a constant defining the grid).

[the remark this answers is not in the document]
ccarouge: Sorry, just realised `niter` is currently (and annoyingly) a grid constant. So should be in `grid_constants` for now.

src/coupled/JAC/params/cable/grid_constants_cbl.F90
ccarouge: I've checked in offline. landuse3 is using `ncs=2` from `cable_define_types.F90` only for the `bgc%` variable and it is using `msoil` from casadimension for the casa variables. So it's ok to keep the value 2 for now. Making it clearer, dropping deprecated feature would be welcome but not necessary here.

I couldn't find anything about this in SLI.

har917: > I couldn't find anything about this in SLI.

Looking again you're likely correct - what I thought I'd spotted wasn't a direct usage of `nsCs` but something like an assumed shape, via `ncs->nsCs=2`. 

There is usage of `var(i)%csoil(:,1)` and `var(i)%csoil(:,2)` and similar in SLI - but at that point `%csoil` is a heat capacity of soil, not a soil carbon variable. Importantly it has dimension `(1:mp,1:ms)` (so number of points, number of soil layers) with these set by `cable_define_types`.

However, I'll keep an eye out - for this.

ps. I agree it would be nice to remove `cable_carbon` and its dependencies

ccarouge:
Ready for review @har917 . I would think the `nsCs` and `ncs` and casadimension nightmare is for another day.

har917 (approved):
okay by me.

As an aside - we likely need an exercise to update/correct the FORD documentation in light of all the the code re-organisation.  I suspect that some of the inline links are out of date

PR #634, opened 2025-10-22
meaningful description of switch

### LINKED ISSUE #536: Review cable_runtime%coupled_run switch
These have been here since ACCESS-1.3 days and we just keep adding switches. The canopy% fields (like Claire has been looking at int JAC version of climate% decs) are all zeroed at declaration,allocation as was required there. If not zeroing for the moment is not causing a problem then I suggest we wait until we pick this up and be sure to CALL the zeroing SUBR.  AND remove this code. Note aswell the coupled_runtime flag has effectively been re-purposed and should at least be renamed to reflect that it does not indicate whether it is in a coupled run or not. I'll make an issue from here

_Originally posted by @JhanSrbinovsky in https://github.com/CABLE-LSM/CABLE/pull/527#discussion_r1921852396_

JhanSrbinovsky:
# CABLE

## Description

The history of this switch was that in ACCESS1.3, when we first introduced 17 tiles, and tiled soils, manipulation of fields files was way harder than what it is now. Plus, we didn't have a similar field to use as a base. We began with files of constant temperature, moisture etc. Progressively made amendments. In one iteration (lasting months perhaps), we clobbered values on snow fields in particular by calling a special initialization routine. This was more or less the first time was coupled (to the UM) - hence the naming of the flag as runtime_coupled. As time went on we occasionally used it in other scenarios. The meaning of the flag and its declaration initialization got mixed up until eventually it was determined that while the flag was TRUE, the conditions should be IF .NOT. . It is all very convoluted evolution to basically set a switch that triggers a special (albeit basic) initialization of soilnow fields in the instance where the initializations through a restart (etc) cannot be trusted.

It make way more sense to rename the switch, declare it as FALSE, and update the conditionals rather than than have this crazy double negative situation.

Fixes #536 

## Type of change

CLARIFICATION

## Checklist

## Testing

----
📚 Documentation preview 📚: https://cable--634.org.readthedocs.build/en/634/

### DIFF
--- a/src/science/gw_hydro/cable_gw_hydro.F90
+++ b/src/science/gw_hydro/cable_gw_hydro.F90
@@ -997,7 +997,7 @@ CONTAINS
     END IF
 
      !> 7. MMY???
-    IF( (.NOT.cable_user%cable_runtime_coupled ) .AND. (first_gw_hydro_call)) THEN
+    IF( (cable_user%soilsnow_init_spec ) .AND. (first_gw_hydro_call)) THEN
 
        IF (cable_runtime%um) canopy%dgdtg = 0.0 ! RML added um condition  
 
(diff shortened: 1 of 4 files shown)


ccarouge (approved):
LGTM

PR #635, opened 2025-10-22
pass iter-ith iteration of zetar

### LINKED ISSUE #521: Passing of iter and niter  in compute friction velocity subroutine
Can implement this at a later date following implementation of:

```
 * In comp_friction_vel, define zetar as REAL(mp) 
 * Call comp_friction_vel with : canopy%zetar(:,iter)
 * Then we can remove iter and NITER from the definition of comp_friction_vel entirely.

```

if possible. For now I just want a branch to test that asymmetric dimension declaration on both sides of the CALL is not the source of fluctuation in AM3, and should be able to do that here if indeed the output is identical.

Following DM with @har917 and @ccarouge

JhanSrbinovsky:
# CABLE

## Description

Received arg declared erroneously. Rather than resolve only need to send current ittteration (from define_canopy() ) part of the array anyway.

see #521 

Fixes #521 

## Type of change

- [x] Bug fix

## Testing

----
📚 Documentation preview 📚: https://cable--635.org.readthedocs.build/en/635/

### DIFF
--- a/src/science/casa-cnp/biogeochem_casa.F90
+++ b/src/science/casa-cnp/biogeochem_casa.F90
@@ -10,7 +10,7 @@ CONTAINS
        nleaf2met,nleaf2str,nroot2met,nroot2str,nwood2cwd,         &
        pleaf2met,pleaf2str,proot2met,proot2str,pwood2cwd)
 USE cable_def_types_mod
-USE cable_common_module, ONLY : cable_user
+USE cable_common_module, ONLY : cable_runtime, cable_user
 USE casadimension
 USE casa_cnp_module
 USE casa_inout_module, ONLY : casa_cnpflux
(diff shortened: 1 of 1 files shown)


src/science/casa-cnp/biogeochem_casa.F90
ccarouge: Shouldn't the changes for `cable_runtime%esm15` come separately? There is no link with the `iter` stuff in `cbl_friction_vel.F90`

JhanSrbinovsky: strange - I went to revert it, it didn't like that - so I just compared it to another local version of the file and the change wasn't even there. I pushed - it refreshed GH - and the change is gone. Of course I could've done it yesterday and just forgotten about it

PR #671, opened 2026-01-11
Change all instances of !$ to !, unless followed by OMP

### LINKED ISSUE #670: Instances of "!$" used for comments
There are a lot of instances of "$!" being used as comments. This leads to breakages any time we want to use OpenMP directives. All of these instances should be replaced with just "!". If there's any code commented in this way for the intention of selective activation, it should be braced with preprocessor directives rather than this reserved method.

Whyborn:
# CABLE

## Description

There were many instances where the "!$" sequence was misused as a regular comment. These lines are treated as code when certain directives are passed e.g. `-qopenmp` as we do in the ACCESS-AM3 build. There are some lines which are truly OpenMP directives, and these lines are untouched. All other instances of '!$' are changed to be regular comments with '!'.

@JhanSrbinovsky Do you know if this was ever used to contextually include code, through misuse of the `-qopenmp` option? If this was ever used for practical reasons, those bits should be moved to standard preprocessor directives.

Fixes #670

## Type of change

- [x] Bug fix

## Testing

----
📚 Documentation preview 📚: https://cable--671.org.readthedocs.build/en/671/

### DIFF
--- a/src/offline/CASAONLY_LUC.F90
+++ b/src/offline/CASAONLY_LUC.F90
@@ -255,10 +255,10 @@ USE casa_offline_inout_module, ONLY : WRITE_CASA_OUTPUT_NC
            CALL POP_IO( pop, casamet, YYYY, 'WRITE_EPI', &
                 ( YYYY.EQ.cable_user%YearEnd ) )
 
-!!$               WHERE (pop%pop_grid(:)%cmass_sum_old.gt.0.1 .and. pop%pop_grid(:)%cmass_sum.gt.0.1 )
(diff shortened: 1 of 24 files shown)


JhanSrbinovsky:
@Whyborn I removed some that were directly interfering. The ones I removed were made by Vanessa Haverd. I hope a test shows this is bitwise comparable and I can confidently avoid having to review every instance over 26 files :)

JhanSrbinovsky (approved):
There were even more of these! I only got rid of the ones that were interfering with whatever it was we were doing at the time (AM3/ESM1.6) From some of the comments I recognize that it was Vanessa Haverd that use "!$" as a comment - very annoying.

I hope that you can do a back to back. test and verify it is bitwise the same so I can confidently avoid reviewing 26 files :)

PR #676, opened 2026-01-19
Merge phenvariable module into casavariable

### LINKED ISSUE #674: Merging the phenology data type file with casa_variable.F90
At the moment, all the CASA derived data types except the phenology are contained in `src/science/casa-cnp/casa_variable.F90`, except the phenology, which is in `src/science/casa-cnp/casa_phenology.F90`. The phenology file doesn't contain any additional functionality, so it would be a good idea to move it to the `casa_variable.F90` file so they're all in the same place.

Whyborn:
# CABLE

## Description

Merges the `src/science/casa-cnp/casa_phenology.F90` file into `src/science/casa-cnp/casa_variable.F90`. That file already defines the rest of the CASA derived types- there's no reason for it not to define this as well. Also makes the merging with AM3 easier, as it's version of `casa_variable.F90` and the offline version will have a unified purpose.

Fixes #674 

## Type of change

- [x] Code clean up

## Testing

----
📚 Documentation preview 📚: https://cable--676.org.readthedocs.build/en/676/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,7 @@ if(CABLE_LIBRARY)
         src/science/casa-cnp/casa_feedback.F90
         src/science/casa-cnp/casa_inout.F90
         src/science/casa-cnp/casa_param.F90
-        src/science/casa-cnp/casa_phenology.F90
+        #src/science/casa-cnp/casa_phenology.F90
         src/science/casa-cnp/casa_readbiome.F90
         src/science/casa-cnp/casa_rplant.F90
         src/science/casa-cnp/casa_sumcflux.F90
@@ -186,7 +186,7 @@ else()
         src/science/casa-cnp/casa_feedback.F90
         src/science/casa-cnp/casa_inout.F90
         src/science/casa-cnp/casa_param.F90
-        src/science/casa-cnp/casa_phenology.F90
+        #src/science/casa-cnp/casa_phenology.F90
         src/science/casa-cnp/casa_readbiome.F90
         src/science/casa-cnp/casa_rplant.F90
         src/science/casa-cnp/casa_sumcflux.F90
--- a/src/offline/CASAONLY_LUC.F90
+++ b/src/offline/CASAONLY_LUC.F90
@@ -11,7 +11,6 @@ SUBROUTINE CASAONLY_LUC( dels,kstart,kend,veg,soil,casabiome,casapool, &
   USE casadimension
   USE casaparm
   USE casavariable
-  USE phenvariable
   USE POP_Types,  ONLY: POP_TYPE
   USE POPMODULE,            ONLY: POPStep, POP_init_single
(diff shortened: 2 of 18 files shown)


CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ if(CABLE_LIBRARY)
         src/science/casa-cnp/casa_feedback.F90
         src/science/casa-cnp/casa_inout.F90
         src/science/casa-cnp/casa_param.F90
-        src/science/casa-cnp/casa_phenology.F90
+        #src/science/casa-cnp/casa_phenology.F90
```
JhanSrbinovsky: why not just kill this?

JhanSrbinovsky (commented):
I do like having a separate module for phenology parameters, in principal phen parameters could be used in the absence of CASA. My bigger concern is that this renders the CABLE in CABLE:main different to that in ESM1.6?

JhanSrbinovsky:
We are far away from incorporating CASA into a JAC framework. This whole AM3 version was in the direction to accomplish that. However, I think that now it might be more prudent to proceed in the direction of staying closer to the ESM1.6 version of CABLE/CASA? 

Typically what happens is we do a CMIP using CABLE version X and then for the next 5+ years all research stems from CMIP using CABLE-X.  Variation on the code level are much less common than changes in configuration, requested output - and thats if they even run the model again. Probably most common is they just use the CMIP data.

In the meantime CABLE will evolve. I suppose should this be relative to a common CMIP7 version of CABLE? or the ESM/AM3 version? The version of CABLE roughly corresponding to the science/ directory, or what has been included in the library. You have probably noticed that the directory structure of CABLE follows that of JULES. This was a CABLE 3 thing that made the mapping trivial. At that time the UMST were adamant that the code be present in their repo if it was intended to ship with JULES releases. They seem to have changed their position on this. Prior to CABLE 3 there was 

```
offline/
coupled/
core/
```

We "basically" pulled things out of core/ to create science/ utils/ params/ etc. coupled/ contained a further breakdown of e.g. ESM/ CM2/.

In summary, perhaps WRT to the changes in science/ we should keep them as in ESM1.6. Perhaps stash the AM3 style changes into a branch that we can pick up later if necessary?

har917:
@Whyborn @ccarouge @JhanSrbinovsky Not sure I like this merge in principle (the practicalities make sense).

Reasoning - there are several use cases where we would want to develop interactive phenology that doesn't connect to CASA, e.g. a running-climate dependent LAI to allow for extended growing seasons in a warming climate.  Forcing the merge of the two TYPEs would be a barrier to this.

JhanSrbinovsky:
hmm - now Im confused - what was the purpose of moving when type decs inside casa_varaible. In AM3 it is a separate module - like MOST of the others. i.e. in JULES/src/util/casa/casa_phenology_type.F90

PR #681, opened 2026-02-05
src/offline/landuse_inout.F90: Replace write interfaces with direct calls to netCDF

SeanBryan51:
This change is a bug fix and also a refactor. Currently, enabling the land use scheme (via `l_landuse`) causes the following error when writing out the land use restart file:

```
 writing cable restart: land use
 Error writing iveg parameter/variable to output file (SUBROUTINE write_output_p
 arameter_r1)
 NetCDF: Not a valid ID                                                         
```

This is because the incorrect netCDF file ID is used in `cable_output.F90` when writing the land use restart file:

https://github.com/CABLE-LSM/CABLE/blob/b2d3f8df5c055a5942f1f5e86288f736582afd74/src/offline/cable_output.F90#L51

Rather than fixing this directly, this change replaces all calls to cable_write.F90 procedures with direct calls to netCDF. This is done so that in the future we can safely replace `cable_output.F90` and `cable_write.F90` with a renewed output module as part of the parallel I/O output module changes (see https://github.com/CABLE-LSM/CABLE/pull/641 and https://github.com/CABLE-LSM/CABLE/pull/655).

Note that the land use scheme is not yet currently functional, even with these changes, as `landuse3.F90` is partially implemented.

## Type of change

- [x] Bug fix
- [x] Refactor

## Checklist

## Testing

```
2026-02-05 14:20:00,805 - INFO - benchcab.benchcab.py:380 - Running comparison tasks...
2026-02-05 14:20:00,831 - INFO - benchcab.benchcab.py:381 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
2026-02-05 14:22:56,227 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
```

----
📚 Documentation preview 📚: https://cable--681.org.readthedocs.build/en/681/

### DIFF
--- a/src/offline/landuse_inout.F90
+++ b/src/offline/landuse_inout.F90
@@ -1138,8 +1138,6 @@ END SUBROUTINE landuse_getdata
     use cable_abort_module,         ONLY : nc_abort
     USE cable_IO_vars_module,       ONLY : latitude,longitude,timeunits,calendar,time_coord, timevar
     USE cable_checks_module,        ONLY : ranges
-    USE cable_output_module,        ONLY: check_and_write, output_par_settings_type
-    USE cable_write_module
     USE cable_common_module,        ONLY : filename,CurYear,cable_user
     USE landuse_variable,           ONLY : landuse_mp
 
@@ -1159,7 +1157,6 @@ END SUBROUTINE landuse_getdata
 
     INTEGER           :: ncid_restart ! netcdf restart file ID
     ! REAL, POINTER,DIMENSION(:,:) :: surffrac ! fraction of each surf type
-    INTEGER           :: dummy ! dummy argument in subroutine call
     INTEGER           :: mlandID, mpID, radID, soilID, napID,                       &
                          soilcarbID, plantcarbID, tID, snowID ! dimension IDs
 
@@ -1180,14 +1177,6 @@ END SUBROUTINE landuse_getdata
 
     INTEGER              ok
 
-    TYPE(output_par_settings_type) :: out_settings
-
-    LOGICAL, PARAMETER :: patchout_var = .TRUE.
-
-    out_settings = output_par_settings_type(met=met, restart=.TRUE.)
-
-    dummy = 0 ! initialise
-
     WRITE(logn, '(A24)') ' Writing restart file...'
     IF ( TRIM(filename%path) .EQ. '' ) filename%path = './'
     frst_out = TRIM(filename%path)//'/'//TRIM(filename%restart_out)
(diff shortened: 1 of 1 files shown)


src/offline/landuse_inout.F90
```diff
@@ -1488,158 +1652,154 @@ SUBROUTINE create_landuse_cable_restart(logn,dels,ktau,soil,mpx,lucmp,cstart,cen
...
     ok = NF90_PUT_VAR(ncid_restart, mstype_id,mstype)
     IF(ok /= NF90_NOERR) CALL nc_abort(ok,                                     &
          'Error writing mstype parameter to '    &
-         //TRIM(frst_out)// '(SUBROUTINE create_restart)')
-     
+         //TRIM(frst_out)// '(SUBROUTINE create_landuse_cable_restart)')
+
 
     ! Write parameters:
-    out_settings%dimswitch = "integer"
```
abhaasgoyal: I believe this specific `dimswitch` is used to fill in the missing values from `cable_write` (`-9999999`). Is this reflected Do we want to fill the missing values as well? Same for `real`.

Also just wanted to confirm (from seeing the code) we want to simplify the shape of dimensions being written by not calling `write_ovar` when we move over to Parallel IO?

SeanBryan51: > I believe this specific `dimswitch` is used to fill in the missing values from `cable_write` (`-9999999`). Is this reflected Do we want to fill the missing values as well? Same for `real`.

Due to the tile based format of the restart file, there actually aren't any missing values in the arrays being written, so I thought we could omit these for the time being.

> Also just wanted to confirm (from seeing the code) we want to simplify the shape of dimensions being written by not calling `write_ovar` when we move over to Parallel IO?

Yep, from my understanding, the land use code can change the number of active patches on a grid cell, so this means that the parallel I/O decomposition based on the initial distribution of active patches in the simulation is no longer valid. There is the idea that changing the format of the restart file from writing arrays with dimension `mp` to  `mland, max_veg_patches` could make this simpler - this would also harmonise the output and restart file formats.

src/offline/landuse_inout.F90
```diff
@@ -1201,284 +1190,459 @@ SUBROUTINE create_landuse_cable_restart(logn,dels,ktau,soil,mpx,lucmp,cstart,cen
...
+    ok = NF90_PUT_ATT(ncid_restart, snowdID, 'long_name', 'Liquid water equivalent snow depth')
+    if (ok /= NF90_NOERR) call nc_abort(ok, 'Error defining snowd variable attributes in restart file. (SUBROUTINE create_landuse_cable_restart)')
+    ok = NF90_PUT_ATT(ncid_restart, snowdID, 'units', 'mm')
+    if (ok /= NF90_NOERR) call nc_abort(ok, 'Error defining snowd variable attributes in restart file. (SUBROUTINE create_landuse_cable_restart)')
+
+    ok = NF90_DEF_VAR(ncid_restart, 'snage', NF90_FLOAT, (/mpID/), snageID)
+    if (ok /= NF90_NOERR) call nc_abort(ok, 'Error defining snage variable in restart file. (SUBROUTINE create_landuse_cable_restart)')
+    ok = NF90_PUT_ATT(ncid_restart, snageID, 'long_name', 'Snow age')
+    if (ok /= NF90_NOERR) call nc_abort(ok, 'Error defining snage variable attributes in restart file. (SUBROUTINE create_landuse_cable_restart)')
+    ok = NF90_PUT_ATT(ncid_restart, snageID, 'units', '??')
```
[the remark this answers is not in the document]
SeanBryan51: Yeah I'm not sure what the unit is here, would have to do some digging. There's a few bogus units in the restart and output file - I think we can address these later when making the output files CF compliant?

PR #695, opened 2026-02-23
Fix snowmelt declaration that should be illegal

Whyborn:
# CABLE

## Description

The declaration of the `ssnow` dummy variable in `snow_melting` seems like it *should* be illegal, as it's using a non-parameter value to declare an explicit array shape. The sequence around this call is also somewhat non-sensical, as the dummy variable is `INTENT(OUT)`, so any information set prior to this call is meaningless.

## Type of change

- [x] Bug fix

## Checklist

## Testing

----
📚 Documentation preview 📚: https://cable--695.org.readthedocs.build/en/695/

### DIFF
--- a/src/science/soilsnow/cbl_snowMelt.F90
+++ b/src/science/soilsnow/cbl_snowMelt.F90
@@ -27,7 +27,7 @@ IMPLICIT NONE
 
     REAL, DIMENSION(mp,0:3) :: smelt1
 
-    ALLOCATE(ssnow(SIZE(ssnow%smelt)))
+    ALLOCATE(snowmlt(SIZE(ssnow%smelt)))
     snowmlt= 0.0
     smelt1 = 0.0
 
--- a/src/science/soilsnow/cbl_snowMelt.F90
+++ b/src/science/soilsnow/cbl_snowMelt.F90
@@ -13,7 +13,7 @@ IMPLICIT NONE
 
     REAL, INTENT(IN) :: dels   ! integration time step (s)
 
-    REAL, DIMENSION(:), INTENT(OUT) :: snowmlt ! snow melt
+    REAL, DIMENSION(:), ALLOCATABLE, INTENT(OUT) :: snowmlt ! snow melt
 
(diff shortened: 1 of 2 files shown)


src/science/soilsnow/cbl_snowMelt.F90
```diff
@@ -1,19 +1,19 @@
...
 
 SUBROUTINE snow_melting (dels, snowmlt, ssnow, soil )
 
-    USE cable_common_module
 IMPLICIT NONE
 
     REAL, INTENT(IN) :: dels   ! integration time step (s)
 
-    REAL, DIMENSION(mp), INTENT(OUT) :: snowmlt ! snow melt
+    REAL, DIMENSION(:), ALLOCATABLE, INTENT(OUT) :: snowmlt ! snow melt
```
har917, line 16: I'd prefer `snowmlt` to ALLOCATED in `ssnow_main` and then passed down into the SUBROUTINEs `snow_melting()`, `cbl_thermal` etc. consistently rather than in a lower routine, then up and down.

src/science/soilsnow/cbl_thermal.F90
```diff
@@ -28,12 +28,10 @@ SUBROUTINE snow_processes_soil_thermal(dels,ssnow,soil,veg,canopy,met,bal)
     TYPE(veg_parameter_type), INTENT(INOUT)  :: veg
     TYPE(met_type), INTENT(INOUT)            :: met ! all met forcing
     TYPE (balances_type), INTENT(INOUT)      :: bal
-    REAL, DIMENSION(mp)                      :: snowmlt !track snow melt 
+    REAL, DIMENSION(:), ALLOCATABLE                      :: snowmlt !track snow melt 
 !    REAL, DIMENSION(:),  INTENT(INOUT)       :: snowmlt  ! replaced by rk4417 - phase2
     INTEGER             :: k,i
 
-    snowmlt = 0.0  ! inserted by rk4417 - phase2
```
har917, line 35: Need to check whether this is a science change or not

PR #689, opened 2026-02-23
Fix Area variable

SeanBryan51:
This change writes the Area variable (`casamet%areacell`) as a parameter instead of a time-varying variable, as `casamet%areacell` is only ever set on initialisation in `load_parameters`. The Area variable is also only written when CASA is enabled, however it is always defined in the output file. This change additionally fixes the definition of the Area variable such that it is only defined when CASA is enabled.

## Type of change

Please delete options that are not relevant.

- [X] Bug fix

## Checklist

- [X] The new content is accessible and located in the appropriate section
- [X] I have checked that links are valid and point to the intended content
- [X] I have checked my code/text and corrected any misspellings

## Testing

- [x] Are the changes non bitwise-compatible with the main branch because of a bug fix or a feature being newly implemented or improved? If yes, add the link to the modelevaluation.org analysis versus the main branch or equivalent results below this line.

Benchcab comparison tests fail as the Area variable is no longer being defined as CASA-CNP is disabled. Ignoring the Area when running `nccmp` restores bitwise compatibility.

----
📚 Documentation preview 📚: https://cable--689.org.readthedocs.build/en/689/

PR #702, opened 2026-02-27
Fix uninitialised access of `met` variables during range checks

SeanBryan51:
There are a few instances where range checks are performed before the variables `met%hod`, `met%doy`, and `met%year` are initialised in the `get_met_data` subroutine. When debug flags are enabled, floating point exceptions are produced when attempting to access these arrays. This occurs for configurations where `MetType /= " "` and `check%ranges /= 0` (i.e. range checks are enabled), and when a variable that is outside the range of its allowed values is detected. I hit this case when writing the `Area` variable in https://github.com/CABLE-LSM/CABLE/pull/689. This change makes the `met` arguments optional when range checks are performed during the initialisation phase to avoid the uninitialised access of `met%hod`, `met%doy`, and `met%year`.

## Type of change

- [x] Bug fix

## Testing

```
2026-02-27 14:46:49,061 - INFO - benchcab.benchcab.py:380 - Running comparison tasks...
2026-02-27 14:46:49,087 - INFO - benchcab.benchcab.py:381 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
2026-02-27 14:49:32,135 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
```

----
📚 Documentation preview 📚: https://cable--702.org.readthedocs.build/en/702/

### DIFF
--- a/src/offline/cable_mpimaster.F90
+++ b/src/offline/cable_mpimaster.F90
@@ -419,7 +419,7 @@ CONTAINS
 
           IF (check%ranges /= NO_CHECK) THEN
             WRITE (*, *) "Checking parameter ranges"
-            CALL constant_check_range(soil, veg, ktau=0)
+            CALL constant_check_range(soil, veg, ktau)
           END IF
 
           IF (CABLE_USER%POPLUC .AND. TRIM(CABLE_USER%POPLUC_RunType) .EQ. 'static') &


src/offline/cable_mpimaster.F90
```diff
@@ -419,7 +419,7 @@ SUBROUTINE mpidrv_master (comm, dels, koffset, kend, PLUME, CRU)
 
           IF (check%ranges /= NO_CHECK) THEN
             WRITE (*, *) "Checking parameter ranges"
-            CALL constant_check_range(soil, veg, 0, met)
+            CALL constant_check_range(soil, veg, ktau=0)
```
ccarouge: Considering `ktau` is already initialised to 0 in the master, couldn't we change this one to use ktau in the argument list instead of 0? 

ktau isn't initialised in the worker so we can't do the same (no point introducing an initialisation at this stage).

PR #700, opened 2026-02-26
Add new error handling capability

SeanBryan51:
This change adds a new error handling capability. This is basically a more polished version of the error handler that was introduced with the parallel I/O changes (https://github.com/CABLE-LSM/CABLE/pull/641) and addresses broader error handling requirements.

Closes #699

## Type of change

Please delete options that are not relevant.

- [x] New feature
- [x] New or updated documentation

## Checklist

- [x] The new content is accessible and located in the appropriate section
- [x] I have checked that links are valid and point to the intended content
- [x] I have checked my code/text and corrected any misspellings

## Testing

- [x] Are the changes bitwise-compatible with the main branch? If working on an optional feature, are the results bitwise-compatible when this feature is off? If yes, copy benchcab output showing successful completion of the bitwise compatibility tests or equivalent results below this line.

```
2026-03-13 14:59:06,973 - INFO - benchcab.benchcab.py:380 - Running comparison tasks...
2026-03-13 14:59:07,000 - INFO - benchcab.benchcab.py:381 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
2026-03-13 15:01:52,104 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
```

----
📚 Documentation preview 📚: https://cable--700.org.readthedocs.build/en/700/

PR #706, opened 2026-03-10
Add Parallel I/O Infrastructure

SeanBryan51:
This change brings in the interface layer for working with the NetCDF Fortran and ParallelIO (PIO) libraries in CABLE. PIO allows for each MPI rank to participate in I/O operations collectively and is a first step in adding MPI support to the serial offline driver, and eventually, to replace the legacy MPI implementation (https://github.com/CABLE-LSM/CABLE/issues/358).

To keep CABLE dependencies as minimal as possible for running in serial mode (without MPI), the interface layer is designed such that PIO support is optional.

To build CABLE with PIO, PIO version 2.6.8 or greater is required.

## Type of change

- [x] New feature

## Checklist

## Testing

```
2026-03-17 12:43:56,608 - INFO - benchcab.benchcab.py:380 - Running comparison tasks...
2026-03-17 12:43:56,634 - INFO - benchcab.benchcab.py:381 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
2026-03-17 12:46:38,124 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
```

----
📚 Documentation preview 📚: https://cable--706.org.readthedocs.build/en/706/

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -116,7 +116,7 @@ if hostname -f | grep gadi.nci.org.au > /dev/null; then
             [[ -n ${mpi} ]] && module add intel-mpi/2019.5.281
             ;;
         gnu)
-            module add gcc/13.2.0
+            module add gcc/14.1.0
             compiler_lib_install_dir=GNU
             [[ -n ${mpi} ]] && module add openmpi/4.1.4
             ;;
--- a/documentation/docs/user_guide/inputs/cable_nml.md
+++ b/documentation/docs/user_guide/inputs/cable_nml.md
@@ -134,6 +134,10 @@ The cable.nml file includes some settings that are common across all CABLE appli
 | cable_user%l_limit_labile        | logical            | .TRUE. .FALSE.                                             | .FALSE.                          | Limit labile in spinup.                                                                                 |
 | cable_user%NtilesThruMetFile     | logical            | .TRUE. .FALSE.                                             | .FALSE.                          | Specify Ntiles through met file.                                                                        |
 | cable_user%l_ice_consistency     | logical            | .TRUE. .FALSE.                                             | .FALSE.                          | If true, ensures consistency between soil and vegetation tiles with permanent ice. All tiles with permanent ice for soil will have ice for vegetation and vice-versa. All the parameters for these new ice tiles are updated to the ice parameters.                       |
+| pio_settings%io_tasks            | integer            | any integer greater than 0                                 | 1                                | Number of PIO I/O tasks used to perform distributed I/O operations.                                     |
+| pio_settings%base                | integer            | any integer greater than 0                                 | 1                                | The starting MPI rank index (1-based) used to assign the subset of I/O ranks from compute ranks.        |
+| pio_settings%stride              | integer            | any integer greater than 0                                 | 1                                | The stride used to assign the subset of I/O ranks from compute ranks.                                   |
+| pio_settings%rearr               | character(len=16)  | 'box' 'subset'                                             | 'box'                            | PIO rearranger to use.                                                                                  |
 
 
 <!-- markdown-link-check-disable-line --> [Walker]: https://doi.org/10.1002/ece3.1173
--- a/src/offline/cable_driver_common.F90
+++ b/src/offline/cable_driver_common.F90
@@ -17,6 +17,7 @@ MODULE cable_driver_common_mod
     l_landuse,                    &
     l_laiFeedbk,                  &
     l_vcmaxFeedbk,                &
+    pio_settings,                 &
     cable_runtime
   USE cable_IO_vars_module, ONLY : &
     soilparmnew,                   &
@@ -98,7 +99,8 @@ MODULE cable_driver_common_mod
     satuParam,      &
     snmin,          &
     cable_user,     & ! additional USER switches
-    gw_params
+    gw_params,      &
+    pio_settings
 
   PUBLIC :: cable_driver_init
   PUBLIC :: cable_driver_init_gswp
--- a/src/util/netcdf/nf90/cable_netcdf_nf90.F90
+++ b/src/util/netcdf/nf90/cable_netcdf_nf90.F90
@@ -1,0 +1,1483 @@
+! CSIRO Open Source Software License Agreement (variation of the BSD / MIT License)
+! Copyright (c) 2015, Commonwealth Scientific and Industrial Research Organisation
+! (CSIRO) ABN 41 687 119 230.
+
+module cable_netcdf_nf90_mod
+  !* The NetCDF Fortran implementation of the netCDF file handling interface in CABLE.
+  !
+  ! For more information on NetCDF Fortran please refer to:
+  !
+  ! - <https://docs.unidata.ucar.edu/netcdf-fortran/current/>
+  ! - <https://github.com/Unidata/netcdf-fortran>
+  !
+  ! @note
+  ! This implementation does not support parallel I/O and can only be used when CABLE is running in serial.
+  ! @endnote
+
+  use cable_netcdf_mod
+
+  use cable_error_handler_mod, only: cable_abort
+  use cable_array_utils_mod, only: array_index
+
+  use netcdf, only: nf90_create
+  use netcdf, only: nf90_open
+  use netcdf, only: nf90_close
+  use netcdf, only: nf90_sync
+  use netcdf, only: nf90_strerror
+  use netcdf, only: nf90_def_dim
+  use netcdf, only: nf90_def_var
+  use netcdf, only: nf90_put_att
+  use netcdf, only: nf90_get_att
+  use netcdf, only: nf90_put_var
+  use netcdf, only: nf90_get_var
+  use netcdf, only: nf90_inq_dimid
+  use netcdf, only: nf90_inq_varid
+  use netcdf, only: nf90_inquire_dimension
+  use netcdf, only: nf90_inquire_variable
+  use netcdf, only: nf90_enddef
+  use netcdf, only: nf90_redef
+  use netcdf, only: NF90_NOERR
+  use netcdf, only: NF90_NETCDF4
+  use netcdf, only: NF90_CLASSIC_MODEL
+  use netcdf, only: NF90_CLOBBER
+  use netcdf, only: NF90_NOCLOBBER
+  use netcdf, only: NF90_WRITE
+  use netcdf, only: NF90_NOWRITE
+  use netcdf, only: NF90_UNLIMITED
+  use netcdf, only: NF90_INT
+  use netcdf, only: NF90_FLOAT
+  use netcdf, only: NF90_DOUBLE
+  use netcdf, only: NF90_FILL_INT
+  use netcdf, only: NF90_FILL_FLOAT
+  use netcdf, only: NF90_FILL_DOUBLE
+  use netcdf, only: NF90_MAX_VAR_DIMS
+  use netcdf, only: NF90_GLOBAL
+
+  implicit none
+
+  private
+
+  public :: cable_netcdf_nf90_io_t
+
+  type, extends(cable_netcdf_decomp_t) :: cable_netcdf_nf90_decomp_int32_t
+    !! A decomposition for 32-bit integer variables in the NetCDF Fortran implementation.
+    integer, allocatable :: dims(:)
+        !! An array of the global dimensions used to describe the shape of the data in the netCDF file
+    integer, allocatable :: compmap(:)
+        !* An array of data offsets describing where each element of the
+        ! in-memory array is located in the netCDF variable data on disk.
+    integer(kind=CABLE_NETCDF_INT32_KIND), allocatable, private :: values_filled(:)
+        !* A buffer for reading/writing data from the in-memory array to the
+        ! netCDF variable data on disk.
+  end type
+
+  type, extends(cable_netcdf_decomp_t) :: cable_netcdf_nf90_decomp_real32_t
+    !! A decomposition for 32-bit real variables in the NetCDF Fortran implementation.
+    integer, allocatable :: dims(:)
+        !! An array of the global dimensions used to describe the shape of the data in the netCDF file
+    integer, allocatable :: compmap(:)
+        !* An array of data offsets describing where each element of the
+        ! in-memory array is located in the netCDF variable data on disk.
+    real(kind=CABLE_NETCDF_REAL32_KIND), allocatable, private :: values_filled(:)
+        !* A buffer for reading/writing data from the in-memory array to the
+        ! netCDF variable data on disk.
+  end type
+
+  type, extends(cable_netcdf_decomp_t) :: cable_netcdf_nf90_decomp_real64_t
+    !! A decomposition for 64-bit real variables in the NetCDF Fortran implementation.
+    integer, allocatable :: dims(:)
+        !! An array of the global dimensions used to describe the shape of the data in the netCDF file
+    integer, allocatable :: compmap(:)
+        !* An array of data offsets describing where each element of the
+        ! in-memory array is located in the netCDF variable data on disk.
+    real(kind=CABLE_NETCDF_REAL64_KIND), allocatable, private :: values_filled(:)
+        !* A buffer for reading/writing data from the in-memory array to the
+        ! netCDF variable data on disk.
+  end type
+
+  type, extends(cable_netcdf_io_t) :: cable_netcdf_nf90_io_t
+    !! The NetCDF Fortran implementation of the netCDF I/O handler interface in CABLE.
+  contains
+    procedure :: init => cable_netcdf_nf90_io_init
+    procedure :: finalise => cable_netcdf_nf90_io_finalise
+    procedure :: create_file => cable_netcdf_nf90_io_create_file
+    procedure :: open_file => cable_netcdf_nf90_io_open_file
+    procedure :: create_decomp => cable_netcdf_nf90_io_create_decomp
+  end type
+
+  type, extends(cable_netcdf_file_t) :: cable_netcdf_nf90_file_t
+    !! The NetCDF Fortran implementation of the netCDF file handling interface in CABLE.
+    integer, private :: ncid
+  contains
+    procedure :: close => cable_netcdf_nf90_file_close
+    procedure :: end_def => cable_netcdf_nf90_file_end_def
+    procedure :: redef => cable_netcdf_nf90_file_redef
+    procedure :: sync => cable_netcdf_nf90_file_sync
+    procedure :: def_dims => cable_netcdf_nf90_file_def_dims
+    procedure :: def_var => cable_netcdf_nf90_file_def_var
+    procedure :: put_att_global_string => cable_netcdf_nf90_file_put_att_global_string
+    procedure :: put_att_global_int32 => cable_netcdf_nf90_file_put_att_global_int32
+    procedure :: put_att_global_real32 => cable_netcdf_nf90_file_put_att_global_real32
+    procedure :: put_att_global_real64 => cable_netcdf_nf90_file_put_att_global_real64
+    procedure :: put_att_var_string => cable_netcdf_nf90_file_put_att_var_string
+    procedure :: put_att_var_int32 => cable_netcdf_nf90_file_put_att_var_int32
+    procedure :: put_att_var_real32 => cable_netcdf_nf90_file_put_att_var_real32
+    procedure :: put_att_var_real64 => cable_netcdf_nf90_file_put_att_var_real64
+    procedure :: get_att_global_string => cable_netcdf_nf90_file_get_att_global_string
+    procedure :: get_att_global_int32 => cable_netcdf_nf90_file_get_att_global_int32
+    procedure :: get_att_global_real32 => cable_netcdf_nf90_file_get_att_global_real32
+    procedure :: get_att_global_real64 => cable_netcdf_nf90_file_get_att_global_real64
+    procedure :: get_att_var_string => cable_netcdf_nf90_file_get_att_var_string
+    procedure :: get_att_var_int32 => cable_netcdf_nf90_file_get_att_var_int32
+    procedure :: get_att_var_real32 => cable_netcdf_nf90_file_get_att_var_real32
+    procedure :: get_att_var_real64 => cable_netcdf_nf90_file_get_att_var_real64
+    procedure :: inq_dim_len => cable_netcdf_nf90_file_inq_dim_len
+    procedure :: inq_var_ndims => cable_netcdf_nf90_file_inq_var_ndims
+    procedure :: put_var_int32_0d => cable_netcdf_nf90_file_put_var_int32_0d
+    procedure :: put_var_int32_1d => cable_netcdf_nf90_file_put_var_int32_1d
+    procedure :: put_var_int32_2d => cable_netcdf_nf90_file_put_var_int32_2d
+    procedure :: put_var_int32_3d => cable_netcdf_nf90_file_put_var_int32_3d
+    procedure :: put_var_real32_0d => cable_netcdf_nf90_file_put_var_real32_0d
+    procedure :: put_var_real32_1d => cable_netcdf_nf90_file_put_var_real32_1d
+    procedure :: put_var_real32_2d => cable_netcdf_nf90_file_put_var_real32_2d
+    procedure :: put_var_real32_3d => cable_netcdf_nf90_file_put_var_real32_3d
+    procedure :: put_var_real64_0d => cable_netcdf_nf90_file_put_var_real64_0d
+    procedure :: put_var_real64_1d => cable_netcdf_nf90_file_put_var_real64_1d
+    procedure :: put_var_real64_2d => cable_netcdf_nf90_file_put_var_real64_2d
+    procedure :: put_var_real64_3d => cable_netcdf_nf90_file_put_var_real64_3d
+    procedure :: write_darray_int32_1d => cable_netcdf_nf90_file_write_darray_int32_1d
+    procedure :: write_darray_int32_2d => cable_netcdf_nf90_file_write_darray_int32_2d
+    procedure :: write_darray_int32_3d => cable_netcdf_nf90_file_write_darray_int32_3d
+    procedure :: write_darray_real32_1d => cable_netcdf_nf90_file_write_darray_real32_1d
+    procedure :: write_darray_real32_2d => cable_netcdf_nf90_file_write_darray_real32_2d
+    procedure :: write_darray_real32_3d => cable_netcdf_nf90_file_write_darray_real32_3d
+    procedure :: write_darray_real64_1d => cable_netcdf_nf90_file_write_darray_real64_1d
+    procedure :: write_darray_real64_2d => cable_netcdf_nf90_file_write_darray_real64_2d
+    procedure :: write_darray_real64_3d => cable_netcdf_nf90_file_write_darray_real64_3d
+    procedure :: get_var_int32_0d => cable_netcdf_nf90_file_get_var_int32_0d
+    procedure :: get_var_int32_1d => cable_netcdf_nf90_file_get_var_int32_1d
+    procedure :: get_var_int32_2d => cable_netcdf_nf90_file_get_var_int32_2d
+    procedure :: get_var_int32_3d => cable_netcdf_nf90_file_get_var_int32_3d
+    procedure :: get_var_real32_0d => cable_netcdf_nf90_file_get_var_real32_0d
+    procedure :: get_var_real32_1d => cable_netcdf_nf90_file_get_var_real32_1d
+    procedure :: get_var_real32_2d => cable_netcdf_nf90_file_get_var_real32_2d
+    procedure :: get_var_real32_3d => cable_netcdf_nf90_file_get_var_real32_3d
+    procedure :: get_var_real64_0d => cable_netcdf_nf90_file_get_var_real64_0d
+    procedure :: get_var_real64_1d => cable_netcdf_nf90_file_get_var_real64_1d
+    procedure :: get_var_real64_2d => cable_netcdf_nf90_file_get_var_real64_2d
+    procedure :: get_var_real64_3d => cable_netcdf_nf90_file_get_var_real64_3d
+    procedure :: read_darray_int32_1d => cable_netcdf_nf90_file_read_darray_int32_1d
+    procedure :: read_darray_int32_2d => cable_netcdf_nf90_file_read_darray_int32_2d
+    procedure :: read_darray_int32_3d => cable_netcdf_nf90_file_read_darray_int32_3d
+    procedure :: read_darray_real32_1d => cable_netcdf_nf90_file_read_darray_real32_1d
+    procedure :: read_darray_real32_2d => cable_netcdf_nf90_file_read_darray_real32_2d
+    procedure :: read_darray_real32_3d => cable_netcdf_nf90_file_read_darray_real32_3d
+    procedure :: read_darray_real64_1d => cable_netcdf_nf90_file_read_darray_real64_1d
+    procedure :: read_darray_real64_2d => cable_netcdf_nf90_file_read_darray_real64_2d
+    procedure :: read_darray_real64_3d => cable_netcdf_nf90_file_read_darray_real64_3d
+  end type
+
+contains
+
+  function type_nf90(type)
+    integer, intent(in) :: type
+    integer :: type_nf90
+    select case(type)
+    case(CABLE_NETCDF_INT)
+      type_nf90 = NF90_INT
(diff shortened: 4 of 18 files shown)


SeanBryan51:
Hi @Whyborn, now that #700 has been merged, these changes are ready to go in. Do you mind giving this your review?

documentation/docs/user_guide/inputs/cable_nml.md
```diff
@@ -134,6 +134,10 @@ The cable.nml file includes some settings that are common across all CABLE appli
...
 | cable_user%l_ice_consistency     | logical            | .TRUE. .FALSE.                                             | .FALSE.                          | If true, ensures consistency between soil and vegetation tiles with permanent ice. All tiles with permanent ice for soil will have ice for vegetation and vice-versa. All the parameters for these new ice tiles are updated to the ice parameters.                       |
+| pio_settings%io_tasks            | integer            | any integer greater than 0                                 | 1                                | Number of PIO I/O tasks used to perform distributed I/O operations.                                     |
+| pio_settings%base                | integer            | any integer greater than 0                                 | 1                                | The base used to assign the subset of I/O ranks from compute ranks.                                     |
```
Whyborn: Can you expand this explanation a bit? How is the base used?

src/util/netcdf/nf90/cable_netcdf_nf90.F90
```diff
@@ -0,0 +1,1485 @@
...
+  end type
+
+contains
+
+  function type_nf90(type)
+    integer, intent(in) :: type
+    integer :: type_nf90
+    select case(type)
+    case(CABLE_NETCDF_INT)
+      type_nf90 = NF90_INT
```
Whyborn, line 187: I'm assuming this is for short `INT`? What about long ints? Or am I getting this confused with the data types that NetCDF files can contain?

tests/test_cable_netcdf.F90
```diff
@@ -0,0 +1,251 @@
...
+      test_suite("parallel", test_list_t([&
+        test_case_pio("test_cable_netcdf_pio_write_read_darray_int32", test_write_read_darray_int32),&
+        test_case_pio("test_cable_netcdf_pio_write_read_darray_real32", test_write_read_darray_real32),&
+        test_case_pio("test_cable_netcdf_pio_write_read_darray_real64", test_write_read_darray_real64)&
+      ]))&
+    ])
+
+  end function cable_netcdf_test_list
+
+  logical function check_decomp(compmap, start, end, block_per_pe) result(result)
```
Whyborn: Should this be called `create_decomp`? The name implies to me that it checks that the arguments form a valid decomp, but it's more like checking that the actual test setup is valid, then creating the decomp, if I'm understanding it right.

build.bash
```diff
@@ -116,7 +116,15 @@ if hostname -f | grep gadi.nci.org.au > /dev/null; then
...
-            module add gcc/13.2.0
+            if [ $do_tests -eq 1 ]; then
+                # This is required to allow for passing a factory procedure
+                # with a polymorphic function result to various test cases. See
+                # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118372 for more
+                # details.
+                module add gcc/14.1.0
+            else
+                module add gcc/13.2.0
+            fi
```
Whyborn: What's the reason for keeping `gcc/13.2.0` at all? Seems a bit redundant to do the tests with one compiler, but do production compilation with another?

src/util/netcdf/pio/cable_netcdf_pio.F90
```diff
@@ -0,0 +1,1199 @@
...
+  end subroutine
+
+  subroutine cable_netcdf_pio_file_def_dims(this, dim_names, dim_lens)
+    class(cable_netcdf_pio_file_t), intent(inout) :: this
+    character(len=*), intent(in) :: dim_names(:)
+    integer, intent(in) :: dim_lens(:)
+    integer :: i, tmp
+    do i = 1, size(dim_names)
+      if (dim_lens(i) == CABLE_NETCDF_UNLIMITED) then
+        call check_pio(pio_def_dim(this%pio_file_desc, dim_names(i), PIO_UNLIMITED, tmp))
```
Whyborn, line 354: Can we assign `CABLE_NETCDF_UNLIMITED = PIO_UNLIMITED`, so that we can drop the `if (dim_lens(i) == CABLE_NETCDF_UNLIMITED) then` clause?

src/util/netcdf/pio/cable_netcdf_pio.F90
```diff
@@ -0,0 +1,1199 @@
...
+    type(pio_var_desc_t) :: tmp
+    if (.not. present(dim_names)) then
+      call check_pio(pio_def_var(this%pio_file_desc, var_name, type_pio(type), tmp))
+      return
+    end if
+    allocate(dimids(size(dim_names)))
+    do i = 1, size(dimids)
+      call check_pio(pio_inq_dimid(this%pio_file_desc, dim_names(i), dimids(i)))
+    end do
+    call check_pio(pio_def_var(this%pio_file_desc, var_name, type_pio(type), dimids, tmp))
```
Whyborn, line 377: I'm a bit confused about this block- what happens when you try to call `pio_def_var` without the `dimids` argument? How is this not required? It seems the [PIO documentation](https://ncar.github.io/ParallelIO/group___p_i_o__def__var.html) is incomplete, as it only mentions that first signature. But I don't understand the meaning of defining a variable, without also defining its dimensions.

src/util/netcdf/pio/cable_netcdf_pio.F90
```diff
@@ -0,0 +1,1199 @@
...
+  end subroutine
+
+  subroutine cable_netcdf_pio_file_put_att_global_string(this, att_name, att_value)
+    class(cable_netcdf_pio_file_t), intent(inout) :: this
+    character(len=*), intent(in) :: att_name, att_value
+    call check_pio(pio_put_att(this%pio_file_desc, PIO_GLOBAL, att_name, att_value))
+  end subroutine
```
Whyborn, line 384: Can we use a type wildcard here for the `att_value`, and condense all of these definitions into one for global attributes, and one for variable attributes?

Edit: Realised this wouldn't work, as I think the underlying PIO/NetCDF routines use an interface, rather than a `case` block.

src/util/netcdf/pio/cable_netcdf_pio.F90
```diff
@@ -0,0 +1,1199 @@
...
+        call check_pio(pio_def_dim(this%pio_file_desc, dim_names(i), dim_lens(i), tmp))
+      end if
+    end do
+  end subroutine
+
+  subroutine cable_netcdf_pio_file_def_var(this, var_name, dim_names, type)
+    class(cable_netcdf_pio_file_t), intent(inout) :: this
+    character(len=*), intent(in) :: var_name
+    character(len=*), intent(in), optional :: dim_names(:)
+    integer, intent(in) :: type
```
Whyborn: Change order of arguments so the `optional` argument is after the required arg? I know the current ordering is more "logical", but I think having an optional argument before a required is too likely to cause problems. Either that or just make `dim_names` required- as I noted in the next comment, I don't see what the meaning of a variable without dimensions is.

src/util/netcdf/pio/cable_netcdf_pio.F90
```diff
@@ -0,0 +1,1199 @@
...
+    class(cable_netcdf_pio_file_t), intent(inout) :: this
+    character(len=*), intent(in) :: var_name
+    integer(kind=CABLE_NETCDF_INT32_KIND), intent(in) :: values
+    integer, intent(in), optional :: start(:), count(:)
+    integer, allocatable :: start_nonopt(:), count_nonopt(:)
+    type(pio_var_desc_t) :: var_desc
+    call get_start_count_nonoptionals(start_nonopt, count_nonopt, [1], start, count)
+    call check_pio(pio_inq_varid(this%pio_file_desc, var_name, var_desc))
+    call check_pio(pio_put_var(this%pio_file_desc, var_desc, start_nonopt, values))
+  end subroutine
```
Whyborn, line 534: Does this chaining of `optional` dummy arguments work? If `start` is not supplied, is the `start` in `get_start_count_nonoptionals` still not `present`?

src/util/netcdf/pio/cable_netcdf_pio.F90
```diff
@@ -0,0 +1,1199 @@
...
+    character(len=*), intent(in) :: var_name
+    real(kind=CABLE_NETCDF_REAL64_KIND), intent(in) :: values(:, :, :)
+    integer, intent(in), optional :: start(:), count(:)
+    integer, allocatable :: start_nonopt(:), count_nonopt(:)
+    type(pio_var_desc_t) :: var_desc
+    call get_start_count_nonoptionals(start_nonopt, count_nonopt, shape(values), start, count)
+    call check_pio(pio_inq_varid(this%pio_file_desc, var_name, var_desc))
+    call check_pio(pio_put_var(this%pio_file_desc, var_desc, start_nonopt, count_nonopt, values))
+  end subroutine
+
```
Whyborn, line 667: So what happens when someone wants to write a distributed array to a NetCDF variable which does not have an unlimited axis? Say we have something which we just want to have dimensions `lat x lon`, can you use `pio_write_darray` for that?

src/offline/cable_driver_common.F90
```diff
@@ -98,7 +99,8 @@ MODULE cable_driver_common_mod
     satuParam,      &
     snmin,          &
     cable_user,     & ! additional USER switches
-    gw_params
+    gw_params,      &
+    pio_settings
```
Whyborn, line 103: Can we put this into a separate namelist, with the components being "stand-alone" variables which are then bound the `pio_settings` derived type? It's something we want to do anyway, so may as well start now. It can live in `cable.nml` for now, but just as a separate namelist.

Whyborn (commented):
A lot of the internal functions are missing any commentary. I found a lot of the routines not particularly intuitive, so some commentary explaining human-readably what the routines are doing would be a big help I think. Doesn't have to be public facing documentation, more like internal developer comments.

I find the tests not very "human"- a lot of abstractions, that I didn't find very easy to reason about exactly what they're doing. This could probably be solved by some commentary as well.

documentation/docs/user_guide/inputs/cable_nml.md
SeanBryan51: No problem, is this better?
```
The starting MPI rank index (1-based) used to assign the subset of I/O ranks from compute ranks.
```

I just noticed these options should read `any integer greater than 1`, I'll correct this.

src/offline/cable_driver_common.F90
SeanBryan51: Can we introduce this properly in a separate PR? The organisation of namelist files is orthogonal to this work

src/util/netcdf/nf90/cable_netcdf_nf90.F90
SeanBryan51: It looks like the PIO Fortran library currently doesn't support long ints: https://github.com/NCAR/ParallelIO/blob/6db2078e45ff77ff4f64e75c3a46502aba4a05aa/src/flib/pio_types.F90#L135

This should be pretty straight forward to contribute if it is necessary.

I think for now, we can assume only 32-bit int support.

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: I'm gonna say no on this one, `CABLE_NETCDF_UNLIMITED` is a constant and I don't think each implementation should overwrite constants in `cable_netcdf_mod` in general.

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: Good spot. I'm happy to reorder the arguments here.

Optional dimensions allows for writing scalars to netCDF files.

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: Yeah the PIO documentation is a bit all over the place unfortunately. There is an overload for defining variables without dimensions here: https://github.com/NCAR/ParallelIO/blob/6db2078e45ff77ff4f64e75c3a46502aba4a05aa/src/flib/pio_nf.F90#L161-L167

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: I'm gonna say no on this one, my issue with declaring `att_val` with `class(*)` is that it would allow `att_val` to take any type to be passed in there, including things you may not want, and the compiler won't know if you have got the arguments wrong, you will only know at run time.

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: Yes

SeanBryan51: As long as the procedure to which the optional arguments are forwarded to are also optional

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: Yes that's possible - in that case you omit the `frame` argument. This is what the unit tests do actually.

tests/test_cable_netcdf.F90
SeanBryan51: Yeah I'm terrible at naming things - definitely open to ideas.

I guess 'create' to me feels like the return value should be the decomposition, maybe `init_decomp`?

build.bash
SeanBryan51: I think it's important to know why a compiler constraint is needed and to not impose them if it isn't necessary as this could hurt portability in general. So I guess I wanted to highlight here that the compiler bump here is needed only to build the tests and not necessarily to build the model.

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: Added in [477228b](https://github.com/CABLE-LSM/CABLE/pull/706/commits/477228bdd008a142a1eea53a1e6ec89c8f876f85) and [87f98fd](https://github.com/CABLE-LSM/CABLE/pull/706/commits/87f98fdc102c090cdaebf1d391e3a7a5db1d4d50)

documentation/docs/user_guide/inputs/cable_nml.md
SeanBryan51: > I just noticed these options should read `any integer greater than 1`, I'll correct this.

Updated in [d1767af](https://github.com/CABLE-LSM/CABLE/pull/706/commits/d1767af7648cf7b5f5ccdfd72245e6e4c91a4b59)

SeanBryan51: Oops, it's late in the day. This should have read greater than or equal to 🙃 so I actually didn't need fix anything there. I'll roll that change back

SeanBryan51:
Thanks for taking a look Lachlan, I forgot to add docs to cable_netcdf_internal.F90. I realised cable_netcdf_internal.F90 was only really necessary for initialising the I/O handler, so I've trimmed down that file to only include the I/O handler initialisation and renamed it to cable_netcdf_init.F90 in [0f30366](https://github.com/CABLE-LSM/CABLE/pull/706/commits/0f30366af66e3669f37060447b5e0c21c374347a).

src/util/netcdf/pio/cable_netcdf_pio.F90
Whyborn: PIO and NetCDF use different values for `UNLIMITED`?

src/util/netcdf/pio/cable_netcdf_pio.F90
Whyborn: What is the actual result of that operation? I didn't think it was possible to define variables without dimensions?

src/offline/cable_driver_common.F90
Whyborn: Fair call

build.bash
Whyborn: I'm personally not in favour of this- I'd much prefer to just up the default version of `gcc` to the higher version. I don't think updating the `gcc` compiler is a difficult barrier to get over. I like keeping the explanation as to why the version was bumped. This opens the potential for a development to pass the unit tests, but fail in production (although quite unlikely)

tests/test_cable_netcdf.F90
Whyborn: Yea `init_decomp` sounds fine to me.

build.bash
SeanBryan51: Good point, I overlooked the possibility that tests will pass for the more recent compiler, but fail for older compilers.

Happy to keep the compiler versions consistent between test and non-test builds.

I realised now that this is a bigger issue for the Fortuno dependency itself, which requires the more recent intel compilers when building from the ground up via `FetchContent`, which is an even bigger compiler jump.

https://github.com/CABLE-LSM/CABLE/blob/7f57306f3cf2715785e6695b34490635961d2203/build.bash#L108-L114

Intel Fortran compilers should have compatible ABIs with each other, so Fortuno can be built with a more recent compiler, but not necessarily have this enforced on the CABLE test build. I've managed to write a spack package recipe for Fortuno, so I'll try install fortuno via spack with the more recent intel compiler, and use find_package to link against it, and try building the CABLE tests with CABLE's default compiler.

SeanBryan51: Just did some tests building the tests for various intel compilers using Fortuno built with Intel (ifx) 2025.2.0:
- intel-compiler/2019.5.281 **FAILED**
- intel-compiler/2020.3.304 **FAILED**
- intel-compiler/2020.3.304 **FAILED**
- intel-compiler/2021.4.0 **FAILED**
- intel-compiler/2021.8.0 **FAILED**
- intel-compiler/2021.10.0 **SUCCESS**

Looks like, for the most part, ifx and ifort modules are incompatible.

SeanBryan51: > Just did some tests building the tests for various intel compilers using Fortuno built with Intel (ifx) 2025.2.0:

Even some older ifx compilers are not compatible:

- intel-compiler-llvm/2022.0.0 **FAILED**
- intel-compiler-llvm/2023.0.0 **FAILED**
- intel-compiler-llvm/2023.2.0 **SUCCESS**

So in summary, we need at least ifort 2021.10.0 or ifx 2023.2.0 to link against Fortuno built with Intel (ifx) 2025.2.0

documentation/docs/user_guide/inputs/cable_nml.md
SeanBryan51: > No problem, is this better?
> 
> ```
> The starting MPI rank index (1-based) used to assign the subset of I/O ranks from compute ranks.
> ```

Added in [4904bc5](https://github.com/CABLE-LSM/CABLE/pull/706/commits/4904bc5337ee4d51f2a1501a8ef8278216fb406d)

tests/test_cable_netcdf.F90
SeanBryan51: Added in [12376c0](https://github.com/CABLE-LSM/CABLE/pull/706/commits/12376c061cd7049104d90e0ca4833ff9fe4d743d)

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: > PIO and NetCDF use different values for `UNLIMITED`?

They could be, it depends on what the PIO developers use, and whether that's consistent with NetCDF Fortran.

Having a separate constant in `cable_netcdf_mod` means we don't need to worry about this.

build.bash
SeanBryan51: > So in summary, we need at least ifort 2021.10.0 or ifx 2023.2.0 to link against Fortuno built with Intel (ifx) 2025.2.0

This is consistent with module versions table in this article: https://community.intel.com/t5/Blogs/Tech-Innovation/Tools/Intel-Fortran-Compiler-Module-mod-Files-Version-Compatibility/post/1600674

I.e. fortuno requires module file version 13

src/util/netcdf/pio/cable_netcdf_pio.F90
Whyborn: Fair call

build.bash
SeanBryan51: Since we currently build Fortuno via FetchContent using the same compiler as CABLE, we'd have to use ifx 2024.X or greater. Or if we build Fortuno separately (e.g. with spack), we'd have to use ifort 2021.10.0 or ifx 2023.2.0 or higher.

I'm happy to do either of these but probably in a separate PR - as the ifx jump would likely break bitwise reproducibility.

For GCC, I'm happy to bump the version as we don't really test CABLE with gfortran.

Whyborn: That's a massive pain. I really don't want to have different compiler versions for a test build and a production build, but can we justify requiring users who want to use Intel compilers to have minimum version 2023?

My personal opinion is yes- given the intel compilers are proprietary, it will be used mostly on centralised HPC systems and these are likely to remain up-to-date with releases of compilers that they're paying for them. People building locally are more likely to use `gcc` and new versions are easily retrievable via `apt`. So I think requiring these relatively recent compiler versions is ok.

An issue here is other compilers- we know Mathias uses NAG, and I assume there hasn't been any testing with that compiler? It wouldn't surprise me if Fortuno straight doesn't compile with NAG, given the issues with GCC and Intel. I don't have a good suggestion yet for that scenario.

SeanBryan51: Agreed that eventually we should move to always use 2023.X or higher.

I don't have access to NAG so I'm unable to test it, however [Fortuno can be built with NAG 7.2](https://github.com/fortuno-repos/fortuno#:~:text=NAG%207.2%20(build%207202))

SeanBryan51: GCC compiler updated: [7b0565c](https://github.com/CABLE-LSM/CABLE/pull/706/commits/7b0565cd713d8de654a49b0b02d08bd153365316)

Whyborn: > I don't have access to NAG so I'm unable to test it, however [Fortuno can be built with NAG 7.2](https://github.com/fortuno-repos/fortuno#:~:text=NAG%207.2%20(build%207202))

Awesome, seems we're in agreement, happy to resolve this.

src/util/netcdf/pio/cable_netcdf_pio.F90
SeanBryan51: Oops I somehow missed this thread.

If you omit the dimensions when defining a variable, it defines a scalar variable. This comes up when we write out the number of soil and vegetation types to the restart file:

```
        int mvtype ;
                mvtype:long_name = "Number of vegetation types" ;
        int mstype ;
                mstype:long_name = "Number of soil types" ;
```

SeanBryan51:
I worked out where the segfault was coming from in the gfortran build. The root cause of this was coming from the new output module and not the parallel I/O infrastructure in this PR.

<details>

<summary>This small program reproduces the error</summary>

```fortran
program main
    implicit none

    type, abstract :: foo_t
    end type

    type, extends(foo_t) :: bar_t
        integer, allocatable :: values(:)
    end type

    type wrapper_t
      class(foo_t), allocatable :: foo
    end type

    type(wrapper_t) :: wrapper_array(2)

    wrapper_array(1)%foo = foo_factory()

contains

    function foo_factory() result(foo)
        class(foo_t), allocatable :: foo
        foo = bar_t([1,2,3,4,5])
    end function

end program
```

The segfault in this program can be avoided if we remove the wrapper array, which was a reasonable work around in my context.

I'll report this to the GCC bugzilla when I have the chance.

</details>

The GCC and Intel builds are now consistent in behaviour.

@abhaasgoyal @Whyborn do you have any further comments on this work?

tests/test_cable_netcdf.F90
Whyborn, line 1: Would you be able to add a bit of commentary on what the tests are actually doing?

tests/fixtures/cable_netcdf_fixtures.F90
```diff
@@ -0,0 +1,114 @@
...
+  end function
+
+  subroutine run_test_pio(this)
+    class(test_case_cable_netcdf_pio_t), intent(in) :: this
+    type(mpi_grp_t) :: mpi_grp
+
+#ifdef SKIP_PIO_TESTS
+    call skip()
+    return
+#endif
```
Whyborn, line 103: Does this report a success if the PIO tests aren't run? Or does it say that they're skipped? Making sure there's something clear to a developer when the PIO tests aren't run, to distinguish between a true success and a success by omission.

Whyborn (commented):
Couple of minor comments

tests/fixtures/cable_netcdf_fixtures.F90
SeanBryan51: <details>

<summary>Here is the output when PIO tests are skipped</summary>

```
$ ./build.bash -DCMAKE_BUILD_TYPE=Release --mpi --tests
-- The Fortran compiler identification is IntelLLVM 2024.0.2
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Check for working Fortran compiler: /apps/intel-tools/wrappers/ifx - skipped
-- Found MPI_Fortran: /apps/intel/compilers_and_libraries_2019.5.281/linux/mpi/intel64/lib/libmpifort.so (found version "3.1")
-- Found MPI: TRUE (found version "3.1") found components: Fortran
-- Found PkgConfig: /bin/pkg-config (found version "1.4.2")
-- Checking for module 'netcdf-fortran'
--   Found netcdf-fortran, version 4.4.6-development
-- Fortuno: build type: Release (manually selected single-config)
-- Found MPI: TRUE (found version "3.1")
-- Configuring done
-- Generating done
-- Build files have been written to: /home/189/sb8430/cable/build
Scanning dependencies of target fortuno
Scanning dependencies of target cable_common
[  0%] Building Fortran object _deps/fortuno-build/src/CMakeFiles/fortuno.dir/fortuno/basetypes.f90.o
[  1%] Building Fortran object _deps/fortuno-build/src/CMakeFiles/fortuno.dir/fortuno/env.f90.o
[  2%] Building Fortran object CMakeFiles/cable_common.dir/src/util/cable_climate_type_mod.F90.o
[  3%] Building Fortran object CMakeFiles/cable_common.dir/src/util/cable_runtime_opts_mod.F90.o
...
[ 99%] Built target cable-tests
[ 99%] Building Fortran object CMakeFiles/cable-mpi.dir/src/offline/cable_offline_driver.F90.o
[100%] Linking Fortran executable cable-mpi
[100%] Built target cable-mpi
-- Install configuration: "Release"
-- Installing: /home/189/sb8430/cable/./bin/cable-mpi
-- Set runtime path of "./bin/cable-mpi" to ""
Internal ctest changing into directory: /home/189/sb8430/cable/build
UpdateCTestConfiguration  from :/home/189/sb8430/cable/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/189/sb8430/cable/build/DartConfiguration.tcl
Test project /home/189/sb8430/cable/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1
    Start 1: cable-tests-serial

1: Test command: /apps/intel-mpi/2019.5.281/intel64/bin/mpiexec "-n" "1" "/home/189/sb8430/cable/build/cable-tests" "~parallel"
1: Working Directory: /home/189/sb8430/cable/build
1: Test timeout computed to be: 10000000
1: === Fortuno - flextensible unit testing framework for Fortran ===
1:
1: Nr. of ranks: 1
1:
1: # Executing test items
1: ...
1:
1: # Test runs
1: Total:      3
1: Succeeded:  3  (100.0%)
1:
1: === Succeeded ===
1/2 Test #1: cable-tests-serial ...............   Passed    0.81 sec
test 2
    Start 2: cable-tests-parallel

2: Test command: /apps/intel-mpi/2019.5.281/intel64/bin/mpiexec "-n" "4" "/home/189/sb8430/cable/build/cable-tests" "parallel"
2: Working Directory: /home/189/sb8430/cable/build
2: Test timeout computed to be: 10000000
2: === Fortuno - flextensible unit testing framework for Fortran ===
2:
2: Nr. of ranks: 4
2:
2: # Executing test items
2: .sss.
2:
2: # Logged event(s)
2: Skipped    [run] parallel/test_cable_netcdf_pio_write_read_darray_int32
2: Skipped    [run] parallel/test_cable_netcdf_pio_write_read_darray_real32
2: Skipped    [run] parallel/test_cable_netcdf_pio_write_read_darray_real64
2:
2: # Suite set-ups
2: Total:      1
2: Succeeded:  1  (100.0%)
2:
2: # Suite tear-downs
2: Total:      1
2: Succeeded:  1  (100.0%)
2:
2: # Test runs
2: Total:      3
2: Skipped:    3  (100.0%)
2: Succeeded:  0  (  0.0%)
2:
2: === Succeeded ===
2/2 Test #2: cable-tests-parallel .............   Passed    0.53 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =   1.42 sec
```

</details>

tests/test_cable_netcdf.F90
SeanBryan51: Docstrings for each test added in [de5c112](https://github.com/CABLE-LSM/CABLE/pull/706/commits/de5c1125f75b4159665071a37a288894a86ff3a1)

Whyborn (approved):
Looks good to me.

tests/test_cable_netcdf.F90
```diff
@@ -0,0 +1,264 @@
...
+      test_suite("parallel", test_list_t([&
+        test_case_pio("test_cable_netcdf_pio_write_read_darray_int32", test_write_read_darray_int32),&
+        test_case_pio("test_cable_netcdf_pio_write_read_darray_real32", test_write_read_darray_real32),&
+        test_case_pio("test_cable_netcdf_pio_write_read_darray_real64", test_write_read_darray_real64)&
+      ]))&
+    ])
+
+  end function cable_netcdf_test_list
+
+  logical function init_decomp(compmap, start, end, block_per_pe) result(result)
```
abhaasgoyal, line 62: Are/should we return any `.TRUE.`/`.FALSE.` value here?

tests/test_cable_netcdf.F90
```diff
@@ -0,0 +1,264 @@
...
+
+    start = this_rank() * block_per_pe + 1
+    end = start + block_per_pe - 1
+
+    allocate(compmap(start:end))
+    compmap(start:end) = [(i, i=start, end, 1)]
+
+  end function init_decomp
+
+  subroutine test_write_read_darray_int32(io_handler_factory, file_name)
```
abhaasgoyal: I think we could make the 3 `test_write` functions more general/is it clearer to have them separate?

tests/test_cable_netcdf.F90
```diff
@@ -0,0 +1,264 @@
...
+    integer, intent(out) :: start, end, block_per_pe
+    integer i
+
+    block_per_pe = LEN / num_ranks()
+
+    call check(mod(LEN, num_ranks()) == 0, msg="test_cable_netcdf.F90: work not divisible by number of ranks")
+    call check(block_per_pe > 0, msg="test_cable_netcdf.F90: not enough work to distribute among pes")
+    call check(mod(block_per_pe, 4) == 0, msg="test_cable_netcdf.F90: block_per_pe must be divisible by 4")
+    result = check_failed()
+    if (result) return
```
abhaasgoyal: Either calling `check_failed()` directly in the if condition or store it in a temporary variable could be more consistent.

abhaasgoyal (approved):
lgtm, just some small comments/questions

tests/test_cable_netcdf.F90
SeanBryan51: Added in [df5ee26](https://github.com/CABLE-LSM/CABLE/pull/706/commits/df5ee2699d0ff0913ed7a6a5e57086259e6bc636)

tests/test_cable_netcdf.F90
SeanBryan51: The return value is used to exit the test early if any pre-checks fail

tests/test_cable_netcdf.F90
SeanBryan51: Added in [7be9387](https://github.com/CABLE-LSM/CABLE/pull/706/commits/7be9387431777d15f83fd28929e574038714b113)

SeanBryan51:
Thanks @abhaasgoyal @Whyborn for reviewing this, I'll rebase and merge this in.

PR #712, opened 2026-03-17
 Introduce MPI-aware structure variables 

SeanBryan51:
Sub-dividing the grid based on the number of MPI ranks available is key to enabling parallelism in the CABLE offline serial driver. This progresses the work to add MPI support to the serial offline driver, and eventually, to replace the legacy MPI implementation (https://github.com/CABLE-LSM/CABLE/issues/358).

The sub-division of the grid is achieved by introducing several new local structure variables to describe the grid information local to the current MPI rank. These are `mp`, `mland`, `land_x`, `land_y`, `land_decomp_start`, `land_decomp_end`, `patch_decomp_start`, `patch_decomp_end` and `landpt`. Global structure variables are also introduced to describe the grid of the global simulation domain, irrespective of the current MPI rank, as this information is necessary for deriving several local structure variables and for generating parallel I/O decompositions.  These are `mp_global`, `mland_global`, `land_x_global`, `land_y_global` and `landpt_global`.

This change also updates constant arrays which are initialised in `open_met_file` to be stored on the global land grid on each MPI rank. This is required for correctly initialising global structure variables such as `landpt_global` which is dependent on the latitude and longitude coordinates of each grid cell in the global simulation domain. These variables are:

- `cable_iovars.F90:latitude(mland_global)`
- `cable_iovars.F90:longitude(mland_global)`
- `cable_iovars.F90:vegtype_metfile(mland_global, nmetpatches)`
- `cable_iovars.F90:vegpatch_metfile(mland_global, nmetpatches)`
- `cable_iovars.F90:soiltype_metfile(mland_global, nmetpatches)`
- `cable_iovars.F90:PrecipScale(mland_global)`
- `cable_input.F90:elevation(mland_global)`
- `cable_input.F90:avPrecip(mland_global)`

Note that, with these changes, `open_met_file` does not support MPI parallelism yet and should only called by a single rank/process. This will be addressed in a separate PR when parallel I/O support is added to the met input reading routines/modules via the interface introduced in #706, which will allow all MPI ranks to read from the file system collectively for local (distributed) and global data.

Note that for the legacy MPI case, the MPI group used to initialise the structure variables on the master driver will always be a single rank group as the master driver will only perform serial I/O.

This change should go in after https://github.com/CABLE-LSM/CABLE/pull/706.

## Type of change

Please delete options that are not relevant.

- [x] New feature
- [x] New or updated documentation

## Checklist

- [x] The new content is accessible and located in the appropriate section
[...]

Issue #420, opened 2024-10-16
Time bounds in CABLE output


SeanBryan51:
Time bounds define the start and end time times of each sampled field in a time series (see [cf-conventions#cell-boundaries](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries) for more details).

For CABLE output to be compatible with ILAMB (https://github.com/CABLE-LSM/CABLE/issues/210), explicit time bounds are required when model outputs are split across multiple years (see https://github.com/rubisco-sfa/ILAMB/issues/81).

**Requirements:**

1. Time bounds for time series data must be contiguous (even when output is split across multiple files per year).

Resources:

- [Adding bounds for a 1-D coordinate variable](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries:~:text=Bounds%20for%201%2DD%20coordinate%20variables)
- [Example bounds for time series data](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-boundaries:~:text=Example%207.5.%20Methods%20applied%20to%20a%20timeseries)
- [Cell methods](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-methods) attribute records the method used to derive data that represents cell values (could be a useful enhancement)

SeanBryan51:
What interval should the time bounds be defined? I guess this is related to how the averaging of output variables is done over a specific time scale before writing out the output.

@abhaasgoyal I remember you did some work to clean up the output routines when implementing the CABLE runtime checks, do you have any input on this?

abhaasgoyal:
@SeanBryan51 I cleared up writing of many output variables, however `time` seems to be unchanged since it was dependent on many other inputs (there is no checks being done currently for the same and the call to write to NetCDF is done directly here unlike using the `cable_write` module).

https://github.com/CABLE-LSM/CABLE/blob/8f62897b6ff6d283565d5df374271425dbce93c7/src/offline/cable_output.F90#L1687

By looking at some sample output NetCDF's, I found the structure of `time` as follows:

```
        double time(time) ;
                time:units = "<seconds/days> since <year>-<month>-<day> <hh>:<mm>:<ss>" ;
                time:coordinate = "" ;
                time:calendar = "<standard/noleap>" ;
```
For this issue, we can add `double time_bnds(time,nv);` to be written for bounds for the corresponding `time`.

SeanBryan51:
Requirement:
[...]

PR #713, opened 2026-04-07
Add parallel I/O output system

SeanBryan51:
**Note:** This change requires intel compiler (ifort) version 2021.6.0 or higher to compile.

This change brings in a new output system based on the parallel I/O infrastructure introduced in #706, and is a direct replacement of the previous output module used for offline CABLE. The main motivation behind this is to add MPI support to the serial offline driver, and eventually, to replace the legacy MPI implementation (#358). This also makes progress towards the proposed output redesign (#715) by introducing the underlying data structures needed for its implementation.

The new output system brings in the following enhancements:

- Optional parallel I/O support via [PIO](https://github.com/NCAR/ParallelIO)
- A new "aggregator" class for calculating time aggregations throughout a simulation
- New data structures for describing output variables which allows for easily adding new diagnostics and/or restarts variables

This change should be brought in after #712.

## Type of change

- [x] Enhancement

## Checklist

## Testing

CABLE benchcab runs tested using `ifort` 2021.10.0.

```
2026-04-28 06:43:59,566 - INFO - benchcab.benchcab.py:380 - Running comparison tasks...
2026-04-28 06:43:59,593 - INFO - benchcab.benchcab.py:381 - tasks: 168 (models: 2, sites: 42, science configurations: 4)
2026-04-28 06:46:47,366 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
```

----
📚 Documentation preview 📚: https://cable--713.org.readthedocs.build/en/713/


SeanBryan51:
Hi @abhaasgoyal @Whyborn, thank you for your time in reviewing these PRs! This one is now ready for review, and no rush of course, it's the heaviest one so far!

As an introduction, it might be useful to read these developer documentation pages:

- https://cable--713.org.readthedocs.build/en/713/api/module/cable_output_mod.html
- https://cable--713.org.readthedocs.build/en/713/api/module/aggregator_mod.html

Benchcab is running currently - I had to make a small change to achieve bitwise compatibility, I will update the status soon.

Thank you again!

src/util/aggregator.F90
```diff
@@ -0,0 +1,1294 @@
...
+    !! An aggregator for 3-dimensional 64-bit real data.
+#ifdef ENFORCE_SINGLE_PRECISION
+    real(kind=real32), dimension(:,:,:), allocatable :: aggregated_data
+#else
+    real(kind=real64), dimension(:,:,:), allocatable :: aggregated_data
+#endif
+    real(kind=real64), dimension(:,:,:), pointer :: source_data => null()
+  end type aggregator_real64_3d_t
+
+  interface new_aggregator
```
Whyborn, line 219: What's the reason for separating the creation of a new aggregator into `new` and `init` phases? Looking at the code, I don't see anything that would stop an aggregator being initialised with `new_aggregator(source_data=..., method=...)`? Within the `aggregator_mod` itself the handling of each bit may be functionally separate, but from the "user" POV it seems logical to group this into a single call.

src/offline/cable_diagnostics.F90
Whyborn, line 1: This is a sort of temporary file right? To replicate the current output behaviour on `main`, while the new API for defining the output is developed?

src/util/output/cable_output.F90
```diff
@@ -0,0 +1,526 @@
...
+      ! effect as setting `reduction_method = "none"`. This is a legacy flag for
+      ! backward compatibility with the CABLE output namelist settings.
+    integer :: var_type = CABLE_OUTPUT_VAR_TYPE_UNDEFINED
+      !* The netCDF variable type using `CABLE_NETCDF_<type>` constants. If not
+      ! specified, the output module will use the native type of the data as the
+      ! netCDF variable type.
+    real :: scale_by = 1.0
+      !* A multiplicative factor to apply to the native diagnostic values when
+      ! writing output.
+    real :: divide_by = 1.0
```
Whyborn, line 221: Why provide this if we already provide a `scale_by`? If they want a division, they can just supply `1 / divide_by` (and it's a tiny bit more efficient to do a division once then apply multiplications, than to do repeated divisions)

src/util/output/cable_output.F90
```diff
@@ -0,0 +1,526 @@
...
+    procedure, public :: size => cable_output_dim_get_size !! Return the dimension size.
+  end type
+
+  type, public :: cable_output_attribute_t
+    !! Type for describing string valued netCDF file attributes.
+    character(64) :: name !! Name of the attribute.
+    character(256) :: value !! Value of the attribute
+  end type
+
+  type, public :: cable_output_variable_t
```
Whyborn, line 164: I feel like it would make our lives easier in the long run to separate the definitions of output variables, output parameters and restart variables, rather than handle them in one big blob and separate via logical flags on the type. I'm guessing this was something you thought about- what made you decide to do it this way?

src/util/output/cable_output.F90
```diff
@@ -0,0 +1,526 @@
...
+
+  integer(kind=int32), parameter, public :: CABLE_OUTPUT_FILL_VALUE_INT32  = -9999999_int32
+  real(kind=real32),   parameter, public :: CABLE_OUTPUT_FILL_VALUE_REAL32 = -1.0e+33_real32
+  real(kind=real64),   parameter, public :: CABLE_OUTPUT_FILL_VALUE_REAL64 = -1.0e+33_real64
+
+  character(64), parameter :: NATIVE_DIM_NAME_PATCH           = "patch_native"
+  character(64), parameter :: NATIVE_DIM_NAME_PATCH_GLOBAL    = "patch_global_native"
+  character(64), parameter :: NATIVE_DIM_NAME_PATCH_GRID_CELL = "patch_grid_cell_native"
+  character(64), parameter :: NATIVE_DIM_NAME_LAND            = "land_native"
+  character(64), parameter :: NATIVE_DIM_NAME_LAND_GLOBAL     = "land_global_native"
```
Whyborn, line 133: What is the point of these parameters? I might be missing something, but I didn't see anywhere that using these parameters was functionally different to just using the associated strings, and the strings are already intuitive enough names so I don't think it adds clarity.

Whyborn (commented):
Done a first pass. My main issue at the moment is with the structure of the `output_variable_t` definition. The current structure is amenable with what already existed within CABLE, but I don't see it being amenable to the planned user output API. The current structure will lead to a combinatory explosion of the number of output variables defined, e.g. a variable with 2 allowed reductions would require 15 `output_variable_t`, for each combination of reduction (3 including no reduction) and aggregation method (5). These would be allocated regardless of whether the variable is written or not.

I think it would make more sense to have a variable definition, which has only one instance per internal variable. It would contain a reference to the target variable, dimensionality and attributes for the variable. Then we use this and the information supplied by the user's output definition to only define the output variables that are actually going to be written.

SeanBryan51:
Hi @abhaasgoyal and @Whyborn, thanks again for taking a look at this! As per our meetings, I removed the `range_native` component from `cable_output_variable_t` ([fcec7c3](https://github.com/CABLE-LSM/CABLE/pull/713/commits/fcec7c397e32137f8ae8f40dab7f04e489f9f9f9)), and I updated the documentation to better highlight the instructions on adding a new variable dimension ([cd26ac0](https://github.com/CABLE-LSM/CABLE/pull/713/commits/cd26ac09a9e976db4843398d111008782980b8cb)).

As I mentioned, there are some changes that should follow after this PR. These were left out either because they broke bitwise compatibility or because they were too outside the scope of the PR. These are:

- [ ] Remove `#ifdef ENFORCE_SINGLE_PRECISION` to allow for writing double precision data
- [ ] Replace normalised time averaging with incremental time averaging
- [ ] Add time bounds (#420)
- [ ] Replace `divide_by` with `scale_by` in `cable_output_variable_t`
- [ ] Accumulate CASA output variables at daily frequency instead of time step frequency
- [ ] Harmonise output and restart file differences by removing `grid_type == "restart"`
- [ ] Built-in documentation generation for the output and restart variable table
- [ ] Move diagnostics definitions (`src/offline/cable_diagnostics.F90` and `src/offline/cable_diagnostics_casa.F90`) to science modules - to coordinate with CABLE library development

Let me know if you any further questions or comments on this! Many thanks

Whyborn:
Which configurations have you used to check bitwise repro? And have you tested with the various current output options available in the `output_inclusion_type`:

* The variable groupings [here](https://github.com/CABLE-LSM/CABLE/blob/aaee8aea1918998a230e1fcfb3cd46a4e37155c5/src/offline/cable_iovars.F90#L223-L238)
* The different frequencies and grids available [here](https://github.com/CABLE-LSM/CABLE/blob/aaee8aea1918998a230e1fcfb3cd46a4e37155c5/src/offline/cable_iovars.F90#L241-L243)

Since this PR commit removes the old system, just want to make sure we don't break existing setups before introducing the new API.

SeanBryan51:
> Which configurations have you used to check bitwise repro? And have you tested with the various current output options available in the `output_inclusion_type`:
>
> * The variable groupings [here](https://github.com/CABLE-LSM/CABLE/blob/aaee8aea1918998a230e1fcfb3cd46a4e37155c5/src/offline/cable_iovars.F90#L223-L238)
> * The different frequencies and grids available [here](https://github.com/CABLE-LSM/CABLE/blob/aaee8aea1918998a230e1fcfb3cd46a4e37155c5/src/offline/cable_iovars.F90#L241-L243)
>
> Since this PR commit removes the old system, just want to make sure we don't break existing setups before introducing the new API.

@Whyborn thanks for reminding me about this! I'm now checking the grid and frequency settings for both flux tower and spatial using benchcab:
```yaml
science_configurations:
- cable:
    output:
      grid: land
      averaging: all
- cable:
    output:
      grid: land
      averaging: user6
- cable:
    output:
      grid: land
      averaging: daily
- cable:
    output:
      grid: land
      averaging: monthly
- cable:
    output:
      grid: mask
      averaging: all
- cable:
    output:
      grid: mask
      averaging: user6
- cable:
    output:
      grid: mask
      averaging: daily
- cable:
    output:
      grid: mask
      averaging: monthly
```

I'll try out a similar method for testing combinations of output variable groups. More to come.

SeanBryan51:
> I'll try out a similar method for testing combinations of output variable groups. More to come.

Following up on this - I've confirmed the following output groupings are consistent with `main`:
```
science_configurations:
  - cable:
      output:
        patch: True
  - cable:
      output:
        met: True
        flux: False
        radiation: False
        carbon: False
        soil: False
        snow: False
        veg: False
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: True
        radiation: False
        carbon: False
        soil: False
        snow: False
        veg: False
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: True
        carbon: False
        soil: False
        snow: False
        veg: False
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: False
        carbon: True
        soil: False
        snow: False
        veg: False
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: False
        carbon: False
        soil: True
        snow: False
        veg: False
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: False
        carbon: False
        soil: False
        snow: True
        veg: False
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: False
        carbon: False
        soil: False
        snow: False
        veg: True
        params: False
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: False
        carbon: False
        soil: False
        snow: False
        veg: False
        params: True
        balances: False
        casa: False
  - cable:
      output:
        met: False
        flux: False
        radiation: False
        carbon: False
        soil: False
        snow: False
        veg: False
        params: False
        balances: True
        casa: False
```

I have confirmed outputs enabled by `output%casa` are consistent with main.

Whyborn (approved):
I can't think of any more output scenarios that we need to check- I'm happy to merge it in

PR #724, opened 2026-05-22
CABLE Library for AM3/ACCESS3

Whyborn:
# CABLE

## Description

This builds the library for AM3 and future ACCESS3 applications. It requires a rejig of the original library, taking some files out of the default library and including them only in ESM1.6. The intention was that the library would only be the science code, but this is not possible with the current setup.

A summary of things that the science directory depends on, that extends outside the `science` directory that I've identified:

* The biogeophysics derived types are not defined within the science directory, as opposed to the CASA derived types which are defined within the `science` directory. Each of these have different definitions for each application (offline, ESM1.6, AM3). Logically, there should only be one definition, but I don't think this will happen any time soon. In the mean time, including the type modules in `src/coupled/AM3/control/cable/CM3` in the library should be an acceptable solution.
* The trickier problems are the namelists and surface types handling. These are related issues, as the `cable_surface_types_mod` module also reads the namelist that sets the CABLE surface types. Do the surface types need to be handled in this way? Is there any foreseeable situation where anyone would not use the standard CABLE PFT indexing? If we remove this, it makes things significantly easier. It also does the extra work of defining a bunch of required JULES variables. I think the best solution is go back to hard-coded CABLE PFT indices, see [this JULES PR](https://github.com/ACCESS-NRI/JULES/pull/345).

## Testing

----
📚 Documentation preview 📚: https://cable--724.org.readthedocs.build/en/724/


Whyborn:
This now, along with [JULES PR #346](https://github.com/ACCESS-NRI/JULES/pull/346), [access-spack-packages PR #460](https://github.com/ACCESS-NRI/access-spack-packages/pull/460) and [UM PR #148](https://github.com/ACCESS-NRI/UM/pull/148) successfully builds AM3 with CABLE as a library and achieves bitwise compatible results.

ccarouge:
@Whyborn Can you please add the benchcab results for bitwise reproducibility?

.github/workflows/ci.yml
```diff
@@ -12,13 +12,16 @@ jobs:
...
+          - .github/build-ci/manifests/library-gcc.spack.yaml.j2
+          - .github/build-ci/manifests/library-intel.spack.yaml.j2
     uses: access-nri/build-ci/.github/workflows/ci.yml@v3
     with:
       # For defaults for args not specified here (eg. spack/packages/config refs), see https://github.com/ACCESS-NRI/build-ci/blob/v3/.github/workflows/README.md
       spack-manifest-path: ${{ matrix.spack-manifest-path }}
       spack-manifest-data-path: .github/build-ci/data/standard.json
       # We need to install the compilers manually here because the upstream image is too large to run on GitHub-hosted runners
       spack-compiler-manifest-path: .github/build-ci/compilers/spack.yaml
+      access-spack-packages-ref: 1d9acf45c39229ff4e90a9aa49922b1169305476
```
ccarouge: What is this line doing? Why is it needed now and not before?

src/offline/cable_iovars.F90
ccarouge, line 1: Why is this file modified for the library work? I would guess we do not compile this for AM3.

src/coupled/AM3/cable/landuse/casa_ESM15landuse.F90
ccarouge, line 1: Why are we adding this to `science/` when it is only to be used in ESM1.5/1.6? Should the `# define ESM15 YES` be `# define ESM15 NO` in the offline repository? Or even, remove that line from the file and deal with it at compilation as usually done for pre-processor flags.

ccarouge: And why is it part of the AM3 library work?

CMakeLists.txt
```diff
@@ -132,17 +126,15 @@ if(CABLE_LIBRARY)
         src/science/soilsnow/cbl_surfbv.F90
         src/science/soilsnow/cbl_thermal.F90
         src/science/soilsnow/cbl_trimb.F90
+	src/science/landuse/landuse3.F90
+        src/science/landuse/landuse_constant.F90
         src/params/cable_phys_constants_mod.F90
```
ccarouge: Could we put this back in alphabetical order? And with the correct indentation.

CMakeLists.txt
```diff
@@ -88,17 +88,11 @@ if(CABLE_LIBRARY)
...
         src/science/casa-cnp/casa_cnp.F90
-        src/science/casa-cnp/casa_dimension.F90
         src/science/casa-cnp/casa_feedback.F90
         src/science/casa-cnp/casa_inout.F90
         src/science/casa-cnp/casa_param.F90
-        src/science/casa-cnp/casa_phenology.F90
-        src/science/casa-cnp/casa_readbiome.F90
         src/science/casa-cnp/casa_rplant.F90
         src/science/casa-cnp/casa_sumcflux.F90
-        src/science/casa-cnp/casa_variable.F90
```
ccarouge, line 99: What happens to all these CASA files for the ESM1.6 library? Were they added but are in fact not needed?

CMakeLists.txt
```diff
@@ -164,7 +160,47 @@ if(CABLE_LIBRARY)
...
+            src/offline/cable_iovars.F90
+            src/science/landuse/casa_ESM15landuse.F90
+        )
+        if (CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
+            add_compile_options(-r8 -i8)
+        elseif(CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+            add_compile_options(-fdefault-real-8 -fdefault-integer-8 -fdefault-double-8)
+        else()
+            message(STATUS "Cannot compile library with unknown compiler")
+        endif()
```
ccarouge, line 217: These are the same options for ESM1.6 and ACCESS3. Are we doubling them in case we want to modify them for ACCESS3? I would prefer to put this block under the `if(CABLE_LIBRARY)` condition directly and have them appear only once. So that it's clearer the only difference between ESM1.6 and ACCESS3 libraries are the sources.

ccarouge (changes requested):
A few requests for more explanation and some relatively minor changes.

Whyborn:
Benchcab output:

```
2026-07-09 15:54:21,176 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed
```

.github/workflows/ci.yml
Whyborn: That's to make the CI use the correct version of the spack packages. I should get the spack packages PR merged in first, tag it, then refer to that tag here.

src/offline/cable_iovars.F90
Whyborn: This was intended to be a bit of future proofing, as Sean's MPI slightly modified the land use change code which was referred to by the AM3 code, even if not yet active. We came up with a better solution in [this merged PR](https://github.com/CABLE-LSM/CABLE/pull/725). I'll remove this change.

src/coupled/AM3/cable/landuse/casa_ESM15landuse.F90
Whyborn: Unfortunately the land use code had moved in a bit from the form that is in AM3. The JULES [src/science_cable/landuse/landuse3.F90](https://github.com/ACCESS-NRI/JULES/blob/AM3-dev/src/science_cable/landuse/landuse3.F90) *should* be the same as [src/science/landuse/landuse3](https://github.com/CABLE-LSM/CABLE/blob/main/src/science/landuse/landuse3.F90), but it isn't, and the coupling code is written based on the JULES version. A few routine signatures have changed, and I was getting type mismatch errors when I tried to quickly fix it.

I should remove the `#ifdef and #def`. I think it's best to get this in first, then reconcile the differences between the way this routine is called in the coupling and then remove this file.

CMakeLists.txt
Whyborn: I think it makes more sense to declare them separately- it makes it clearer, and there may be instances where the flags are not the same and this preempts that.

CMakeLists.txt
Whyborn: Unfortunately these are bits that had been rewritten for AM3- the way the derived types are structured is different to offline, see [this JULES issue](https://github.com/ACCESS-NRI/JULES/issues/320). This is related to the duplicate `ALLOCATABLE/POINTER` type definitions that I wanted to unify some time back. The AM3 versions of these live in the JULES [src/util/casa](https://github.com/ACCESS-NRI/JULES/tree/AM3-dev/src/util/casa).

CMakeLists.txt
Whyborn: Yep, will fix that.

CMakeLists.txt
ccarouge: I don't think I follow what you are saying. So these files are not the same between ESM1.6 and AM3, ok. But it does not explain what happens for these files for ESM1.6. Right now, these files completely disappear of the compilation of ESM1.6 library in this PR, making me doubt that the compilation of the ESM1.6 library would work.. 
At the same time, I'm pretty sure we said this version of CABLE won't be compatible with ESM1.6, so should we not include an ESM1.6 library in the CMakeLists.txt file?

Whyborn: Yea, I think its either:
1. Include those files in the ESM1.6 library build
2. Drop the ESM1.6 library build altogether

With the conversations around BLAZE ongoing, I think it would be wise to keep the ESM1.6 library code there and add the files. I don't think it would be much work to make the ESM1.6 library build again, but keeping the list of files up to date would make future life easier if someone else was to come work on it.

ccarouge: OK, when you have time, can you look at making sure the ESM1.6 library works. Can it also be added to the compilation CI in CABLE's PRs then so we would do 2 library builds?

CMakeLists.txt
ccarouge: OK.

CMakeLists.txt
Whyborn: Can I make that a separate PR? Separate that work from the AM3 work.

.github/workflows/ci.yml
Whyborn: Oh actually I think I can remove that line completely when the spack packages update is merged, as I think it uses the head of the default branch if not specified?

CMakeLists.txt
```diff
@@ -164,7 +166,47 @@ if(CABLE_LIBRARY)
...
+            src/coupled/AM3/control/casa/types/casa_met_type.F90
+            src/coupled/AM3/control/casa/types/casa_phenology_type.F90
+            src/coupled/AM3/control/casa/types/casa_pool_type.F90
+            src/coupled/AM3/control/casa/types/sum_casa_flux_type.F90
+            src/coupled/AM3/control/casa/types/sum_casa_pool_type.F90
+            src/coupled/AM3/control/casa/types/casa_variable.F90
+            src/coupled/AM3/control/casa/types/phen_variable.F90
+            src/coupled/AM3/control/casa/types/casa_dimension.F90
+            src/coupled/AM3/control/casa/types/casa_inout.F90
+            src/coupled/AM3/control/casa/types/casa_readbiome.F90
```
ccarouge: I would simplify these paths:

`src/coupled/AM3/cable/types/` and `src/coupled/AM3/casa/types`.

src/coupled/AM3/cable/landuse/casa_ESM15landuse.F90
ccarouge: Files that are modified from the offline version for the coupled versions should all be stored under `src/coupled/AM3`. Can we put this under `src/coupled/AM3/cable/landuse`? If there are other files that are slightly different 

Can you also clean up the `src/coupled/AM3` directory tree of everything that isn't needed for the library, please? These files are not used and should not be kept around. As part of this PR, please.

CMakeLists.txt
ccarouge: Should we keep the files in the list then so that it's easier to identify these files are problematic for the work on ESM1.6? Or comment them out with enough explanation to know what's needed?

ccarouge (changes requested):
A few more things to do I think. Small things.

src/coupled/AM3/cable/landuse/casa_ESM15landuse.F90
Whyborn: Yep, can do.

CMakeLists.txt
Whyborn: We can't leave them in the list uncommented, as the build with fail, but I think commenting them out with an explanation is a good idea.

Whyborn:
Ok, I think this is ready to go. Aside from the main purpose of making the library for AM3, this PR:

* Removes all the files that are not necessary for the library build, so that they only exist in one place (i.e. JULES).
* Moves the old landuse code to the `coupled/AM3/cable/landuse` directory, until it can be unified with the offline version.
* Adds CI for the ACCESS3 library build with GCC and Intel, as well as a build of UM with the head of the `AM3-dev` branches for both `UM` and `JULES`. This is the bit I'm not sure is the best choice, since neither of those repositories have CI set up either, so people could break CABLE builds without realising. I can add CI to both those respective repositories as part of the respective library PRs.

Future work related to this is the unifying of the types used between the coupled and offline models. This should be possible with a "medium" amount of work.

ccarouge (approved):
LGTM. Thanks for the additional work.

Whyborn:
@ccarouge There was one small addition required after the discussions today- the spec for the UM CI required a `+cable`. I'll merge if the CI still passes.

PR #725, opened 2026-05-27
Initialize land_decomp_start to 1

Whyborn:
# CABLE

## Description

The recent addition of `to_land_index_global` in `src/offline/cable_iovars.F90` is currently not compatible with running in coupled mode. This routine is called in the land use code, which is intended to be included in coupled applications, so it does need to be included in coupled mode. I believe initialising `land_decomp_start` to 1 will make this function as expected in coupled applications, where we never call the decomposition initialisation. Do you think this is correct @SeanBryan51?

## Type of change

- [x] Bug fix

## Checklist

## Testing

----
📚 Documentation preview 📚: https://cable--725.org.readthedocs.build/en/725/

### DIFF
--- a/src/offline/cable_iovars.F90
+++ b/src/offline/cable_iovars.F90
@@ -106,7 +106,7 @@ MODULE cable_IO_vars_module
        max_vegpatches,   & ! The maximum # of patches in any grid cell
        nmetpatches         ! size of patch dimension in met file, if exists
 
-  INTEGER :: land_decomp_start
+  INTEGER :: land_decomp_start = 0
     !! Starting land point index of this MPI rank in global land array
   INTEGER :: land_decomp_end
     !! Ending land point index of this MPI rank in global land array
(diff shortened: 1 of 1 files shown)


SeanBryan51 (approved):
Thanks Lachlan, this should be pretty harmless so I'm happy for this to go in. Just a suggestion: would be clearer to set `land_decomp_start = 1` in the coupled model (before the land use code is called)? This would make it a bit more obvious that these variables should be set elsewhere at runtime.

Whyborn:
Yea, I can suss out the best place to do this coupled mode. I think it would be best to introduce a function like `set_decomp_coupled` or something, that just sets `land_decomp_start=1`?

Whyborn:
@SeanBryan51 Made that change- haven't decided where it should be called in coupled mode yet, but I think that can be decided later.

SeanBryan51 (approved):
Thanks Lachlan, happy for this to go in when you are.

PR #728, opened 2026-06-18
CI: Add datafile for consistency with other MCRs

CodeGat:
## Background

Currently, `CABLE` is the only Model Component Repository that doesn't use `build-ci`s `inputs.spack-manifest-data-path` input. It's not required by MCRs, but for an update to the CI in `access-spack-packages` (see ACCESS-NRI/access-spack-packages#420) which tests manifests defined in model component repositories, it would be much simpler if this repository had one. 

If there are changes to the manifests in the future, more common data can be added to this file, as well!

## The PR

* Added `.github/build-ci/data/standard.json`, a file that defines key-value pairs that are usable in `.github/build-ci/manifests/*.spack.yaml.j2` files. Currently, only one `target: x86_64` is defined.
* Replaced `x86_64` with `{{ target }}` in the manifest, which will be replaced at spack install time
* Added `inputs.spack-manifest-data-path` to the `build-ci` workflow call

----
📚 Documentation preview 📚: https://cable--728.org.readthedocs.build/en/728/


Whyborn:
Can the entries in `standards.json` take multiple entries for different build configurations?

CodeGat:
Either by naming them differently in the same datafile, or via different datafiles themselves, sure!

Whyborn (approved):
Looks good.

PR #729, opened 2026-07-02
Bring up to date with ACCESS3 science

Whyborn:
# CABLE

## Description

These are the science changes that occurred in the CABLE code that sits within JULES for AM3, that we want to move into the library. The changes removing the calculation of `gamzz` and `trad_corr` were discussed in the CABLE science update PR, [here](https://github.com/ACCESS-NRI/JULES/pull/324#discussion_r2719795074) and [here](https://github.com/ACCESS-NRI/JULES/issues/320#issuecomment-3765844984). I had missed the change from using `ssnow%otss` to `snow%tss` in the `soil_trad` calculation, but appears to be a deliberate change from [this JULES commit](https://github.com/ACCESS-NRI/JULES/commit/77e4ef1da95a7244863c06cefad99d9bc7b6e10d).

I am a little confused though- in your comment [here](https://github.com/ACCESS-NRI/JULES/issues/320#issuecomment-3765844984) @har917, you said that the `trad_corr` is still required in the `impl` call, but [this JULES commit](https://github.com/ACCESS-NRI/JULES/commit/77e4ef1da95a7244863c06cefad99d9bc7b6e10d) removes it from both `expl` and `impl` call. Should the `trad_corr` be re-introduced to the `impl` call?

It may also be preferred to leave `cbl_cbm_mod.F90` within JULES- at the moment, I only moved it since it lived in the same directory as all the geophysics type definitions, which must live within the library, but this isn't required for any of the science code. It may make more sense to leave it within JULES.

## Testing

2026-07-03 12:23:52,415 - INFO - benchcab.benchcab.py:391 - 0 failed, 168 passed

----
📚 Documentation preview 📚: https://cable--729.org.readthedocs.build/en/729/

### DIFF
--- a/src/science/soilsnow/cbl_soilsnow_main.F90
+++ b/src/science/soilsnow/cbl_soilsnow_main.F90
@@ -86,17 +86,6 @@ USE cable_phys_constants_mod,  ONLY: density_liq, density_ice
 
     ssnow%wbliq = ssnow%wb - ssnow%wbice
 
-  ! soilsnow_init_spec, special initalizations in um_init NA for ESM1.5
-
-   xx=soil%css * soil%rhosoil
-   IF (ktau <= 1)                                                              &
-     ssnow%gammzz(:,1) = MAX( (1.0 - soil%ssat) * soil%css * soil%rhosoil      &
-            & + (ssnow%wb(:,1) - ssnow%wbice(:,1) ) * Ccswat * Cdensity_liq           &
-            & + ssnow%wbice(:,1) * Ccsice * Cdensity_ice, xx ) * soil%zse(1) +   &
-            & (1. - ssnow%isflag) * Ccgsnow * ssnow%snowd
-
-
-
     DO k = 1, ms ! for stempv
 
        ! Set liquid soil water fraction (fraction of saturation value):
--- a/src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
+++ b/src/coupled/AM3/control/cable/CM3/cbl_cbm_mod.F90
@@ -77,7 +77,6 @@ REAL :: veg_wt(mp)
 REAL :: veg_trad(mp)      
 REAL :: soil_wt(mp)       
 REAL :: soil_trad(mp)     
-REAL :: trad_corr(mp)     
 
 CHARACTER(LEN=8),  PARAMETER :: subr_name = "ExpCbmT_"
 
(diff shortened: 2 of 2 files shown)


JhanSrbinovsky:
I'll wait for comment from @har917 before I approve

src/science/soilsnow/cbl_soilsnow_main.F90
```diff
@@ -86,17 +86,6 @@ SUBROUTINE soil_snow(dels, soil, ssnow, canopy, met, bal, veg)
 
     ssnow%wbliq = ssnow%wb - ssnow%wbice
 
-  ! soilsnow_init_spec, special initalizations in um_init NA for ESM1.5
-
-   xx=soil%css * soil%rhosoil
```
har917, line 91: Where is the equivalent bit of science in the library - i.e. where is `%gamzz` being initialised on the first call in offline and AM3 (likely `um_init_soilsnow`)?

Whyborn: Apparently this is now stored in the restart. We had a long conversation about this in the [JULES PR](https://github.com/ACCESS-NRI/JULES/pull/324), and the gist was that this bit of code can be removed. In offline mode, we call `spec_init_soil_snow` from the driver which overrides this anyway.

har917:
>I am a little confused though- in your comment https://github.com/ACCESS-NRI/JULES/issues/320#issuecomment-3765844984 @har917, you said that the trad_corr is still required in the impl call, but [this JULES commit](https://github.com/ACCESS-NRI/JULES/commit/77e4ef1da95a7244863c06cefad99d9bc7b6e10d) removes it from both expl and impl call. Should the trad_corr be re-introduced to the impl call?

aargh - this brings back "war" memories which I have (thankfully) mostly forgotten.

For definite - `trad_corr` should not be in `cbm_expl`

For `cbm_impl`:  Basically we need to use `%tss` [end of time step] or (`%otss` [start of time step] and `trad_corr` [change over time step] together in partnership).  If I remember the deliberations/testing, properly Martin and I landed on using `%tss` (for consistency with how we deal with the vegetation contribution to `rad%trad` which is also the [end of time step] value). 

But probably worth looking through the issue/PR around temperature oscillations.

Whyborn:
I found the [relevant issue](https://github.com/ACCESS-NRI/JULES/issues/239)- seems the current form is the one that was deemed to be most accurate i.e. not using `trad_corr` in either `expl` or `impl` and using the end of timestep temperature `tss`.

har917 (approved):
Happy with this.

Whyborn:
I'll do a quick benchcab run to make sure nothing unexpected has happened. If it shows no differences, I'll merge.

Whyborn:
Benchcab reports identical results.

PR #730, opened 2026-07-17
Fix library build for ESM1.6

Whyborn:
# CABLE

## Description

This fixes the ESM1.6 library build. The cause of the breakage was indirectly a small change to `src/science/landuse/landuse3.F90` in [this commit](https://github.com/CABLE-LSM/CABLE/commit/9d5457bf9e78b997196e4a6fbc0bfffb8f3c4bc9#diff-a9e63c5c5e70456053f09751927e847fcac3607a400f29390fb60f282f6831a0). This code is not used coupled model, but is referenced in the other science code so required for compilation. The indexer used there is defined in `src/offline/cable_iovars.F90`, which has diverged from `src/coupled/esm/cable_iovars.F90`. The offline version has some changes which are strict additions and one which is an effective change, which is the change of the checks mechanism. This is only called during the offline I/O, and has no effect on the coupled model.

The `cable_IO_vars_module` doesn't seem to be included in the AM3 build (without library), although it is in code that is not part of a module (I didn't even know that was possible... see [this `src/science/landuse/landuse3.F90` code](https://github.com/CABLE-LSM/CABLE/blob/285ff0a11c98b50b3b690a52b57ef33a6b726094/src/science/landuse/landuse3.F90#L730-L734) that also appears in JULES). The offline version of `cable_iovars.F90` also seems to work with AM3.

The gist is, we have multiple versions of `cable_iovars.F90`, but only need 1. The dependence of the science code on stuff in the IO vars module should be removed, but that is a future problem.

## Type of change

- [x] Bug fix

## Checklist

## Testing

----
📚 Documentation preview 📚: https://cable--730.org.readthedocs.build/en/730/

### DIFF
--- a/.github/build-ci/compilers/spack.yaml
+++ b/.github/build-ci/compilers/spack.yaml
@@ -2,6 +2,7 @@ spack:
   specs:
   - gcc@13.2.0 %gcc@8.5.0
   - intel-oneapi-compilers-classic@2021.10.0
+  - intel-oneapi-compilers@2025.2.0
   view: false
   concretizer:
     unify: false
--- a/.github/build-ci/data/standard.json
+++ b/.github/build-ci/data/standard.json
@@ -1,3 +1,5 @@
 {
+    "gcc_compiler_ver": "13.2.0",
+    "oneapi_compiler_ver": "2021.10.0",
     "target": "x86_64"
 }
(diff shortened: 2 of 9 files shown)


Whyborn:
@ccarouge This is now ready to go. The only change required to get the library itself building was to include `src/offline/cable_iovars.F90` in the library build. This is not ideal, but it's referenced in the `src/science/landuse3.F90` code, which is in turn referenced elsewhere in the science code so it should be included.

The rest is updated the CI. The CI now includes:
* GCC and Intel library builds, with the same GCC and Intel compiler versions as the standalone code
* UM7 build with the same specification as the `oneapi` build in the UM7 CI. This included an additional secret to access the private components, which was added by @CodeGat.

ccarouge (approved):
LGTM

PR #732, opened 2026-07-17
Update CI to be consistent with other MCRs

Whyborn:
# CABLE

## Description

This brings the CABLE CI more in line with our other model component repositories. Uses the `data/standards.json` to define targets and compiler versions, and uses the two-step CI workflow.

----
📚 Documentation preview 📚: https://cable--732.org.readthedocs.build/en/732/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ jobs:
       allow-ssh-into-spack-install: false  # If true, PR author must ssh into instance to complete job
       spack-manifest-data-path: .github/build-ci/data/standard.json
       spack-oci-buildcache-url: oci://ghcr.io/CABLE-LSM/build-ci-buildcache
-      allow-ssh-into-spack-install: false
       spack-compiler-manifest-path: .github/build-ci/compilers/spack.yaml
       # builtin-spack-packages-ref: develop
       # access-spack-packages-ref: api-v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,6 @@ jobs:
       # access-spack-packages-ref: api-v2
       # spack-config-ref: main
       # spack-ref: releases/v1.1
+      
+      # CABLE isn't part of the ACCESS-NRI org so can't access the runners
+      run-self-hosted: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,27 +1,40 @@
 name: Build
 on:
   pull_request:
+  push:
     branches:
-      - main
+      - '[0-9]*.[0-9]*' # Default branch naming
 jobs:
+  pre-ci:
+    name: Pre-CI
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -1,27 +1,43 @@
...
     with:
-      # For defaults for args not specified here (eg. spack/packages/config refs), see https://github.com/ACCESS-NRI/build-ci/blob/v3/.github/workflows/README.md
-      spack-manifest-path: ${{ matrix.spack-manifest-path }}
+      spack-manifest-path: ${{ matrix.file }}
+      allow-ssh-into-spack-install: false  # If true, PR author must ssh into instance to complete job
       spack-manifest-data-path: .github/build-ci/data/standard.json
-      # We need to install the compilers manually here because the upstream image is too large to run on GitHub-hosted runners
-      spack-compiler-manifest-path: .github/build-ci/compilers/spack.yaml
-      allow-ssh-into-spack-install: false
-      spack-oci-buildcache-url: oci://ghcr.io/CABLE-LSM/build-ci-buildcache  # https://github.com/orgs/CABLE-LSM/packages/container/package/build-ci-buildcache
```
CodeGat, line 23: Going to need to add these lines back - CABLE-LSM-specific because we don't have self-hosted runners for this org, and need to install compilers separately

CodeGat (changes requested):

.github/workflows/ci.yml
Whyborn: Ahh ok

Whyborn: That means there needs to be an agreement between what's in `data/standard.json` and `compilers/spack.yaml`?

CodeGat:
You'll need to add back the compilers manifest and the `spack-compiler-manifest-path: .github/build-ci/compilers/spack.yaml` to the CI, too - otherwise it won't have any compilers in the image :grin:

Whyborn:
I thought this was something that had changed in Spack v1, in that compilers were treated just like packages? Do the compilers need be the same as specified in the `standard.json` file?

Whyborn:
Actually, can I make the `compilers/spack.yaml` a Jinja template as well and use the values in the `standard.json`?

CodeGat:
> I thought this was something that had changed in Spack v1, in that compilers were treated just like packages? Do the compilers need be the same as specified in the standard.json file?

That's spack v1.2 that has the ability to install specs in dependent groups, not spack v1.1 :cry: I have a proof of concept in https://github.com/ACCESS-NRI/access-test-component/pull/24/changes#diff-b4c7c915a7b8ac12622ea6b301e345926c430f2ca7db90c8cd69376407e731d8 but I don't think we should move to that just yet

> Actually, can I make the compilers/spack.yaml a Jinja template as well and use the values in the standard.json?

Ah, a new use case! I just looked through the code and realised that we don't actually interpolate jinja stuff in the compilers manifest - we should, so I'll create https://github.com/ACCESS-NRI/build-ci/issues/313. For now though, just restore the original file.

Whyborn:
Ok @CodeGat the CI is now successful

CodeGat (approved):
Awesome, thanks mate

